PR #9, opened 2019-10-24
Add circleci config

fragosti:


fragosti:
Going to merge this for now just so we have something but will move over to drone when we implement deployments.

PR #10, opened 2019-10-25
Add nodemon, and convert ormconfig to be in typescript

fragosti:
I wanted to add a `yarn dev` by using `nodemon` to improve dev iteration speed, but the `ormconfig.json` was getting in my way as it was pointing to `js` files and confusing `ts-node`. So I took inspiration from the data pipeline and moved the orm configuration to an `ormconfig.ts` file, which aside from allowing `yarn dev` to work, should also help us productionize this thing. I can see it at the very least being useful to use ENV variables to define the ormconfig, but being able to compose the ormconfig depending on the deployment also seems useful.


README.md
```diff
@@ -33,9 +33,9 @@ To develop ontop of `0x-api`, follow the following instructions:
 
 4. Open the `.env` file and edit the following fields. Defaults are defined in `config.ts`/`config.js`. The bash environment takes precedence over the `.env` file. If you run `source .env`, changes to the `.env` file will have no effect until you unset the colliding variables.
 
-    - `NETWORK_ID` -- the network you'd like your relayer to run on (e.g: `1` -> mainnet, `42` -> Kovan, 3 -> Ropsten, etc...). Defaults to `42`
+    - `NETWORK_ID` -- the network you'd like your API to run on (e.g: `1` -> mainnet, `42` -> Kovan, 3 -> Ropsten, etc...). Defaults to `42`
```
BMillman19: If we are removing references to the word "relayer" here should we should also change a couple references in `http_service.ts` and `errors.ts`

BMillman19:
Also, as mentioned in slack, looks like there is a dependency on a `metadata.json` file which doesn't exist in this branch or `master`. Looks like we might be able to get away with doing away with the `metadata.json` file in favor of using `constants.NULL_ADDRESS` as the default fee recipient

fragosti:
Changed them in `error.ts` but will follow up with the other changes.

PR #11, opened 2019-10-29
Refactor 0x-api to support multiple APIs

fragosti:
This PR changes the structure of the code to not be so SRA specific. It also does some general cleanup.
- Creates a `handlers` and `routers` directory where you can put routes and handlers for different sub-APIs. See [Router](https://expressjs.com/en/api.html#router). The `mesh_gateway` handlers are all added to a Router, which is then added to the `mesh_gateway` path.
- Moves a bunch of the utilities to a `utils` folder.

We don't explicitly support multiple APIs yet, but I would rather wait to do that until we need it. The code is separate enough now that it would be trivial to implement.


src/routers/mesh_gateway_router.ts
```diff
@@ -0,0 +1,48 @@
...
+import { OrderBookService } from '../services/orderbook_service';
+
+export const createMeshGatewayRouter = (orderBook: OrderBookService): express.Router => {
+    const router = express.Router();
+    const handlers = new MeshGatewayHandlers(orderBook);
+    /**
+     * GET AssetPairs endpoint retrieves a list of available asset pairs and the information required to trade them.
+     * http://sra-spec.s3-website-us-east-1.amazonaws.com/#operation/getAssetPairs
+     */
+    router.get('/asset_pairs', asyncHandler(MeshGatewayHandlers.assetPairsAsync.bind(MeshGatewayHandlers)));
```
BMillman19: why is a static method being used here but not for the others?

src/services/http_service.ts
```diff
@@ -1,58 +1,20 @@
...
-         * http://sra-spec.s3-website-us-east-1.amazonaws.com/#operation/postOrder
-         */
-        app.post('/v3/order', asyncHandler(handlers.postOrderAsync.bind(handlers)));
-        /**
-         * GET Order endpoint retrieves the order by order hash.
-         * http://sra-spec.s3-website-us-east-1.amazonaws.com/#operation/getOrder
-         */
-        app.get('/v3/order/:orderHash', asyncHandler(Handlers.getOrderByHashAsync.bind(Handlers)));
-
+        app.use('/mesh_gateway/', createMeshGatewayRouter(orderBook));
```
BMillman19: how would a migration to the future mesh gateway querying system look like?

src/routers/mesh_gateway_router.ts
fragosti: They are static if they do not use `this._orderBook` and instance methods otherwise

src/services/http_service.ts
fragosti: I've no clue :) I don't necessarily see how this pattern wouldn't work though. Also too much uncertainty around "future mesh gateway querying" to make any decisions based on that.

PR #13, opened 2019-10-29
Use TypeORM entities the correct way

fragosti:
In launch-kit we were not using typeORM decorators because they didn't look good when transpiling to js, but that is no longer a problem.

I also removed `reflect-metadata` as it wasn't being used.


src/config.ts
```diff
@@ -1,14 +1,10 @@
 // tslint:disable:custom-no-magic-numbers
 import { BigNumber } from '0x.js';
 import { assert } from '@0x/assert';
-import * as crypto from 'crypto';
```
BMillman19: Looks like this accidentally got committed as a part of this branch

src/entity/SignedOrderEntity.ts
```diff
@@ -1,65 +1,99 @@
...
-            type: 'varchar',
-        },
-        remainingFillableTakerAssetAmount: {
-            type: 'varchar',
-        },
-    },
-});
+import { Column, Entity, PrimaryColumn } from 'typeorm';
+
+@Entity({ name: 'signed_orders' })
```
BMillman19: should this be `signed_order`?

src/entity/SignedOrderEntity.ts
```diff
@@ -1,65 +1,99 @@
...
+export class SignedOrderEntity {
+    @PrimaryColumn({ name: 'hash', type: 'varchar' })
+    public hash?: string;
```
BMillman19: Are these all optional?

src/entity/SignedOrderEntity.ts
```diff
@@ -1,65 +1,99 @@
...
+        this.expirationTimeSeconds = opts.expirationTimeSeconds;
+        this.makerFee = opts.makerFee;
+        this.takerFee = opts.takerFee;
+        this.makerFeeAssetData = opts.makerFeeAssetData;
+        this.takerFeeAssetData = opts.takerFeeAssetData;
+        this.makerAssetAmount = opts.makerAssetAmount;
+        this.takerAssetAmount = opts.takerAssetAmount;
+        this.salt = opts.salt;
+        this.signature = opts.signature;
+        this.remainingFillableTakerAssetAmount = opts.remainingFillableTakerAssetAmount;
```
BMillman19: I wonder if there is any way we don't need to make changes in 3 different places in order to add a new property

src/services/orderbook_service.ts
```diff
@@ -71,16 +71,16 @@ export class OrderBookService {
         quoteAssetData: string,
     ): Promise<OrderbookResponse> {
         const connection = getDBConnection();
-        const bidSignedOrderModels = (await connection.manager.find(SignedOrderModel, {
+        const bidSignedOrderEntitys = (await connection.manager.find(SignedOrderEntity, {
```
BMillman19: sp: `bidSignedOrderEntities`

BMillman19: same below

src/entity/SignedOrderEntity.ts
fragosti: Yes -- I am not sure **why** but I would rather leave them this way for now.

src/entity/SignedOrderEntity.ts
fragosti: I can't think of a really clean way. FWIW this object existed before I just moved it here.

PR #14, opened 2019-10-29
Use pino as a logger

fragosti:
`console.log` is OK, but pino is way nicer.


src/runners/websocket_service_runner.ts
```diff
@@ -25,4 +25,4 @@ import { utils } from '../utils/utils';
     const meshClient = new WSClient(config.MESH_ENDPOINT);
     // tslint:disable-next-line:no-unused-expression
     new WebsocketService(server, meshClient);
-})().catch(utils.log);
+})().catch(logger.info);
```
BMillman19: `logger.error`

PR #15, opened 2019-10-30
Add a docker-compose file for development and move over to postgres

fragosti:
The `docker-compose` file is so we don't have to manually run mesh, and now postgres in development. It does not use the `Dockerfile` use for deploying the API.

The reason for moving over to postgres is so that we can productionize this and point the API to an RDS instance in production.


.env_example
```diff
@@ -0,0 +1,12 @@
+# Required
```
BMillman19: do we want to provide sample values for these?

.env_example
```diff
@@ -0,0 +1,12 @@
...
+NETWORK_ID=
+ETHEREUM_RPC_URL=
+
+# Optional
+MESH_ENDPOINT=
```
BMillman19: Seems like we have a couple different naming conventions going on here, for example `ETHEREUM_RPC_URL` vs `MESH_ENDPOINT`

.env_example
```diff
@@ -0,0 +1,12 @@
...
+FEE_RECIPIENT=
```
BMillman19: `FEE_RECIPIENT_ADDRESS` ?

.env_example
BMillman19: Also might be good to provide some comments around what each defaults to w/o having to go look int `config.ts`

fragosti: I would provide sample values if it weren't for the `ETHEREUM_RPC_URL` value. Are we ok with putting a valid value in there?

The defaults are mentioned in the README.

PR #17, opened 2019-11-05
More cleanup and hardening

fragosti:
The main attempt here is to make sure the HTTP service can still run when mesh is down or we cannot connect to mesh. Unfortunately the mesh RPC client has an unhandled promise rejection when the you give it an invalid URL. By default, unhandled promise rejections are warnings, and do not crash your app, so I have implemented the same here for the custom rejection handlers. By default, an uncaught **exception** does crash your app. The point with catching them here is to have all logs be formatted in the same way.
There are also some other clean-up items in here.

### DIFF
--- a/package.json
+++ b/package.json
@@ -42,7 +42,7 @@
         "@0x/contract-addresses": "^3.3.0-beta.0",
         "@0x/contract-wrappers": "^12.2.0-beta.0",
         "@0x/json-schemas": "^4.1.0-beta.0",
-        "@0x/mesh-rpc-client": "6.0.0-beta-v3",
+        "@0x/mesh-rpc-client": "^6.0.0-beta-v3",
         "@0x/order-utils": "^8.5.0-beta.0",
         "@0x/order-watcher": "^4.0.17",
         "@0x/subproviders": "^5.1.0-beta.0",
--- a/package.json
+++ b/package.json
@@ -42,7 +42,7 @@
         "@0x/contract-addresses": "^3.3.0-beta.0",
         "@0x/contract-wrappers": "^12.2.0-beta.0",
         "@0x/json-schemas": "^4.1.0-beta.0",
-        "@0x/mesh-rpc-client": "^5.0.0-beta-v3",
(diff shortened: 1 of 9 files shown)


src/index.ts
```diff
@@ -14,19 +14,35 @@ import { WebsocketService } from './services/websocket_service';
...
+        logger.error(err);
+    }
     const orderBookService = new OrderBookService(meshClient);
     // tslint:disable-next-line:no-unused-expression
     new HttpService(app, orderBookService);
-})().catch(logger.error.bind(logger));
+})().catch(error => logger.error(error));
+
+process.on('uncaughtException', err => {
+    logger.error(err);
```
BMillman19: does there need to be an `if(err)` check here as well?

src/services/orderbook_service.ts
```diff
@@ -157,20 +157,22 @@ export class OrderBookService {
         const paginatedApiOrders = paginationUtils.paginate(apiOrders, page, perPage);
         return paginatedApiOrders;
     }
-    constructor(meshClient: WSClient) {
+    constructor(meshClient?: WSClient) {
```
BMillman19: is the `OrderBookService` able to function without a `meshClient` ?

package.json
```diff
@@ -42,7 +42,7 @@
         "@0x/contract-addresses": "^3.3.0-beta.0",
         "@0x/contract-wrappers": "^12.2.0-beta.0",
         "@0x/json-schemas": "^4.1.0-beta.0",
-        "@0x/mesh-rpc-client": "^5.0.0-beta-v3",
```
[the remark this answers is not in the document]
fragosti: Good eye.

src/index.ts
fragosti: No the error is guaranteed to be there.

src/services/orderbook_service.ts
fragosti: It can read, but it cannot write. Perhaps I should 500 on that POST then.

fragosti: I ended up implementing this.

PR #19, opened 2019-11-07
More startup improvements

fragosti:
- It can take a while to connect and sync with mesh, and there is no need to block the addition of express routes until then. 
- The forever module is no longer necessary because k8s will take care of restarts.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -20,4 +20,4 @@ COPY . .
 RUN yarn build
 
 EXPOSE 3000
-CMD [ "./node_modules/.bin/forever", "lib/index.js" ]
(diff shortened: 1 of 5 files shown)


Dockerfile
```diff
@@ -20,4 +20,4 @@ COPY . .
 RUN yarn build
 
 EXPOSE 3000
-CMD [ "./node_modules/.bin/forever", "lib/index.js" ]
```
BMillman19: what does `forever` actually give us now that we'll be losing?

fragosti: Forever just makes sure to restart the the process if it exits. However, kubernetes does this at the container level and is the more standard way of doing it.

PR #20, opened 2019-11-08
Remove NETWORK_ID, upgrade mesh, move websocket endpoint path

fragosti:
- Remove `NETWORK_ID` param and replace with `CHAIN_ID`.
- Remove `networkdId` query param altogether.
- Upgrade version of mesh used in development.
- Implement temporary patch for the latest version of the RPC client.
- Have the mesh gateway websocket endpoint listen at `/mesh_gateway` instead of at the root.


src/config.ts
```diff
@@ -28,9 +28,9 @@ export const HTTP_PORT = _.isEmpty(process.env.HTTP_PORT)
     ? 3000
     : assertEnvVarType('HTTP_PORT', process.env.HTTP_PORT, EnvVarType.Port);
 // Default network id to use when not specified
```
BMillman19: `Default chain id to use..`

PR #25, opened 2019-11-26
update mesh version

xianny:
Very tiny PR to update Mesh to latest version.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -13,7 +13,7 @@ services:
         ports:
             - "5432:5432"
     mesh:
-        image: 0xorg/mesh:7.1.1-beta-0xv3
+        image: 0xorg/mesh:7-0xv3
         restart: always
         environment:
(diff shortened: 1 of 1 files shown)


docker-compose.yml
```diff
@@ -13,7 +13,7 @@ services:
         ports:
             - "5432:5432"
     mesh:
-        image: 0xorg/mesh:7.0.2-beta-0xv3
+        image: 0xorg/mesh:7.1.1-beta-0xv3
```
fragosti: Latest is now 7.2.1-beta-0xv3

xianny:
Mesh team is now tagging new minor/patch releases as `7-0xv3` so we should be able to use this tag for a while. 


> jacob 12:58 PM
> Posting publicly so we can chat together about it. We should post major versions of docker images as tags. 0xorg/mesh:7.2.1-beta-0xv3 and 0xorg/mesh:7-0xv3 so downstream projects and users can depdend on the major version and get bug fixes at the same time, if they choose to occasionally pull

PR #28, opened 2019-12-02
Add 7 day generated protocol fees to the staking API response

fragosti:
I also updated the events-pipeline image so we have meta-data now!

Open to improvements in naming here...


src/services/staking_data_service.ts
```diff
@@ -164,6 +178,19 @@ const nextEpochStatsQuery = `
                     AS operator_share
             FROM staking.operator_share_changes
         )
+        , past_7d_fills_fills_by_pool AS (
```
alexkroeger: doesn't seem to make sense to me to have this in both locations. It's not necessarily related to an epoch, so maybe make it a pool attribute? It makes the most sense to me in nextEpochStats tho.

PR #29, opened 2019-12-05
Add request logging

xianny:
This PR adds logging for incoming requests. 

I tried out using `express-pino-logger` but found it too restrictive. The supported serializers coerced `Express.Request` and `Express.Response` into `IncomingMessage` and `ServerResponse` from the native 'http' module, which doesn't easily expose the request body or params. 

So instead I wrote our own little middleware. It's pretty simple but hopefully useful.

One of the features is that all requests have the logger object attached. So you can do:

```
public async ordersAsync(req: express.Request, res: express.Response): Promise<void> {
    req.logger.info(`Retrieving orders!`);
    // handle the request
}
```

Some examples of the logs produced:

Screen Shot 2019-12-04 at 5 03 45 PM

Screen Shot 2019-12-04 at 5 29 49 PM

(With pino-pretty in dev mode; actual logs won't have so much whitespace)

@alexkroeger , I'm not separating the namespace (e.g. `/sra`) from the path e.g. (`/orders`) like we discussed. I did some experimentation and found that they don't resolve reliably. Specifically, express seems to have trouble resolving the `baseUrl` for POST methods. So we would have some logs with `{ namespace: '/sra', path: '/orders }` and some logs with `{namespace: undefined, path: '/sra/orders' }`. I figured it would be better to be consistent and make them *all* `{ path: '/sra/orders' }`.

### DIFF
--- a/src/logger.ts
+++ b/src/logger.ts
@@ -1,3 +1,45 @@
+// tslint:disable-next-line:no-implicit-dependencies
+import * as core from 'express-serve-static-core';
 import * as pino from 'pino';
 
 export const logger = pino();
+
+/**
+ * log middleware
+ */
+export function logMiddleware(): core.RequestHandler {
+    const handler = (req: any, res: any, next: core.NextFunction) => {
+        const startTime = Date.now();
+        function writeLog(): void {
+            const responseTime = Date.now() - startTime;
+            res.removeListener('finish', writeLog);
+            res.removeListener('close', writeLog);
+            const logMsg = {
+                req: {
+                    url: req.originalUrl.split('?')[0],
+                    method: req.method,
+                    headers: {
+                        'user-agent': req.headers['user-agent'],
+                        host: req.headers.host,
+                    },
+                    body: req.body,
+                    params: req.params,
+                    query: req.query,
+                },
+                res: {
+                    statusCode: res.statusCode,
+                    statusMessage: res.statusMessage,
+                    error: res.error,
+                },
+                responseTime,
+                timestamp: Date.now(),
+            };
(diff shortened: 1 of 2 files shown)


src/logger.ts
```diff
@@ -1,3 +1,45 @@
...
+                    method: req.method,
+                    headers: {
+                        'user-agent': req.headers['user-agent'],
+                        host: req.headers.host,
+                    },
+                    body: req.body,
+                    params: req.params,
+                    query: req.query,
+                },
+                res: {
```
alexkroeger: we may want to log some additional data about our responses to be able to track what happens afterward @BMillman19

src/logger.ts
```diff
@@ -1,3 +1,45 @@
...
+import * as core from 'express-serve-static-core';
 import * as pino from 'pino';
 
 export const logger = pino();
+
+/**
+ * log middleware
+ */
+export function logMiddleware(): core.RequestHandler {
+    const handler = (req: any, res: any, next: core.NextFunction) => {
```
xianny: hack: typescript won't compile with `Express.Request/Response`, the type declarations don't match up with the actual API :/

fragosti: In the other middleware file we import directly from express. `error_handling.ts`

src/logger.ts
xianny: chatted with @BMillman19 , we will be interested in responses when we do the /quote endpoint, but not for this one. thanks for flagging!

fragosti:
Sorry had one more comment -- could we move the middleware method to the `middleware` directory?

PR #30, opened 2019-12-06
Meta-txn price, quote & fill endpoints

BMillman19:
This PR implements Milestone 2 of the Meta-txn Tech Spec. It implements the `/price`, `/quote` and `/fill` endpoints of the `meta-txn` route. The `/fill` route only submits txns on behalf of callers with a whitelisted 0x API Key (currently no one), and otherwise returns an unsigned Ethereum txn that calls `executeTransaction`.

Milestone 3 will be implemented in a subsequent PR to make the meta-txn relay more robust. This PR however should not be blocked for inclusion until that is ready, since we can switch on/off the relaying of txns and it will start in the OFF state.

What is blocking the merging of this PR is adding documentation for the `meta-txn` endpoints to the docs (PR: https://github.com/0xProject/website/pull/280).


src/services/meta_transaction_service.ts
```diff
@@ -0,0 +1,384 @@
...
+        const placeholderZeroExTransaction = this._generateZeroExTransaction(placeholderOrders, sellAmount, buyAmount, signatures, takerAddress, gasPrice);
+
+        // By setting the `from` address to the 0x transaction `signerAddress`, the 0x txn signature
+        // is not checked by the smart contracts. This allows us to estimate the gas of the Ethereum txn
+        // before having the 0x txn signature from the taker. This estimate will always be slightly less
+        // than the gas used in actuality, since the signature verification must be performed when it is
+        // actually submitted.
+        // TODO(fabio): Compute the gas difference between w/ and w/o signature verification and add this gas
+        // amount to the estimate returned here.
+        // TODO(fabio): Since we must run the gas estimation with the `takerAddress` as the `from` address AND
```
fabioberger: @jalextowle

src/constants.ts
```diff
@@ -14,6 +14,7 @@ export const DEFAULT_LOGGER_INCLUDE_TIMESTAMP = true;
 export const ONE_SECOND_MS = 1000;
 export const ONE_MINUTE_MS = ONE_SECOND_MS * 60;
 export const TEN_MINUTES_MS = ONE_MINUTE_MS * 10;
+export const ERC20_BRIDGE_ASSET_PREFIX = '0xdc1600f3';
```
jalextowle, line 17: We should probably try to import stuff like this from the monorepo so that it stays up to date. I can fix this.

.env_example
```diff
@@ -1,6 +1,8 @@
 # Required
 CHAIN_ID=1337
 ETHEREUM_RPC_URL=http://ganache:8545
+META_TXN_RELAY_ADDRESS=
+META_TXN_RELAY_PRIVATE_KEY=
```
fabioberger, line 5: This first iteration uses a single Meta-txn relay address. A subsequent PR will make txn relaying more robust as part of Milestone 3.

src/constants.ts
fabioberger: That was not added in this PR.

src/runners/http_service_runner.ts
```diff
@@ -66,6 +70,19 @@ export async function runHttpServiceAsync(
...
 
+    // Meta transaction http service
+    if (dependencies.metaTransactionService) {
+        app.use(META_TRANSACTION_PATH, createMetaTransactionRouter(dependencies.metaTransactionService));
+    } else {
+        logger.error(`API running without meta transactions service`);
+    }
+    const signerService = new SignerService();
+    const handlers = new SignerHandlers(signerService);
+    app.post(
```
dekz, line 81: Did we want this generally available on all HTTP services or just on certain runners?

src/schemas/meta_transaction_quote_request_schema.json
```diff
@@ -0,0 +1,39 @@
...
+        "sellToken": {
+            "type": "string"
+        },
+        "buyToken": {
+            "type": "string"
+        },
+        "takerAddress": {
+            "$ref": "/addressSchema"
+        },
+        "slippage": {
```
dekz, line 13: We've been using `slippagePercentage` as a query param.

src/services/meta_transaction_service.ts
```diff
@@ -0,0 +1,196 @@
...
+    }
+    private static _range(rangeCount: number): number[] {
+        return [...Array(rangeCount).keys()];
+    }
+    constructor() {
+        this._provider = MetaTransactionService._createWeb3Provider(ETHEREUM_RPC_URL);
+        const swapQuoterOpts = {
+            chainId: CHAIN_ID,
+        };
+        this._swapQuoter = SwapQuoter.getSwapQuoterForMeshEndpoint(this._provider, MESH_WEBSOCKET_URI, swapQuoterOpts);
```
dekz, line 44: This should copy what's in `swap_service.ts` and pass in an Orderbook so it's fetched from the db.

```typescript
constructor(orderbook: Orderbook, provider: SupportedProvider) {
        this._provider = provider;
        const swapQuoterOpts: Partial<SwapQuoterOpts> = {
            chainId: CHAIN_ID,
            expiryBufferMs: QUOTE_ORDER_EXPIRATION_BUFFER_MS,
            liquidityProviderRegistryAddress: LIQUIDITY_POOL_REGISTRY_ADDRESS,
        };
        this._swapQuoter = new SwapQuoter(this._provider, orderbook, swapQuoterOpts);
```

src/services/meta_transaction_service.ts
```diff
@@ -0,0 +1,196 @@
...
+        this._contractWrappers = new ContractWrappers(this._provider, { chainId: CHAIN_ID });
+        this._web3Wrapper = new Web3Wrapper(this._provider);
+    }
+    public async calculateMetaTransactionPriceAsync(
+        params: CalculateMetaTransactionQuoteParams,
+    ): Promise<CalculateMetaTransactionPriceResponse> {
```
dekz, line 50: Could this use `swap_service` to reduce the duplication and trim the results?

src/services/meta_transaction_service.ts
```diff
@@ -0,0 +1,196 @@
...
+
+        // generate the zeroExTransaction object
+        const expirationTimeSeconds = new BigNumber(Date.now() + TEN_MINUTES_MS)
+            .div(ONE_SECOND_MS)
+            .integerValue(BigNumber.ROUND_CEIL);
+        const zeroExTransaction: ZeroExTransaction = {
+            data: txData,
+            salt: generatePseudoRandomSalt(),
+            signerAddress: takerAddress,
+            gasPrice,
```
dekz, line 187: TODO we should round this up to the nearest integer of full gwei so they're bucketed a bit closer

src/services/signer_service.ts
```diff
@@ -0,0 +1,183 @@
...
+        signature: string,
+    ): Promise<BigNumber> {
+        // Verify 0x txn won't expire in next 60 seconds
+        // tslint:disable-next-line:custom-no-magic-numbers
+        const sixtySecondsFromNow = new BigNumber(+new Date() + 60);
+        if (zeroExTransaction.expirationTimeSeconds <= sixtySecondsFromNow) {
+            throw new Error('zeroExTransaction expirationTimeSeconds in less than 60 seconds from now');
+        }
+
+        const devUtils = new DevUtilsContract(this._contractWrappers.contractAddresses.devUtils, this._provider);
```
dekz, line 80: We should have one of these as ABI loading is non negligible.

src/services/signer_service.ts
```diff
@@ -0,0 +1,183 @@
...
+        const orders = decodedArray[1];
+
+        // Verify orders don't expire in next 60 seconds
+        orders.forEach(order => {
+            if (order.expirationTimeSeconds <= sixtySecondsFromNow) {
+                throw new Error('Order included in zeroExTransaction expires in less than 60 seconds from now');
+            }
+        });
+
+        const gasPrice = zeroExTransaction.gasPrice;
```
dekz, line 91: We should have some acceptable limit of gasPrice

src/services/signer_service.ts
```diff
@@ -0,0 +1,183 @@
...
+
+        return protocolFee;
+    }
+    public async generateExecuteTransactionEthereumTransactionAsync(
+        zeroExTransaction: ZeroExTransactionWithoutDomain,
+        signature: string,
+        protocolFee: BigNumber,
+    ): Promise<PartialTxParams> {
+        const gasPrice = zeroExTransaction.gasPrice;
+        const gas = await this._contractWrappers.exchange
```
dekz, line 116: TODO our pattern is to `eth_call` and `estimateGas` in parallel and return the result of `eth_call` validations

src/services/signer_service.ts
```diff
@@ -0,0 +1,183 @@
...
+            zeroExTransaction,
+            signature,
+            protocolFee,
+        );
+
+        const signedEthereumTransaction = await this._privateWalletSubprovider.signTransactionAsync(ethereumTxnParams);
+
+        const transactionHash = await this._contractWrappers.exchange
+            .executeTransaction(zeroExTransaction, signature)
+            .sendTransactionAsync({
```
dekz, line 156: `shouldValidate: false` as we've already validated this with an `eth_call`

src/services/signer_service.ts
```diff
@@ -0,0 +1,183 @@
...
+        let nonce = (this._nonceTrackerSubprovider as any)._nonceCache[senderAddress];
+        if (nonce === undefined) {
+            nonce = await this._getTransactionCountAsync(senderAddress);
+        }
+        return nonce;
+    }
+    private async _getTransactionCountAsync(address: string): Promise<string> {
+        const nonceHex = await this._web3Wrapper.sendRawPayloadAsync<string>({
+            method: 'eth_getTransactionCount',
+            params: [address, 'latest'],
```
dekz, line 179: `pending`?

src/runners/http_service_runner.ts
fabioberger: So, it would be great to get clarity from you on this. I made it available on both the `http_service_runner` and `signer_runner`. I was under the impression that `http_service_runner` is used when testing locally, and the individual runners are spun up in production. Is that correct?

src/schemas/meta_transaction_quote_request_schema.json
fabioberger: I copied `SwapQuoteRequestSchema`. Is that incorrect as well?

src/services/meta_transaction_service.ts
fabioberger: Potentially, but I do expect this to diverge from swap_quote once we actually re-coup the gas + pf from the taker's takerAsset.

src/services/signer_service.ts
fabioberger: Do you mean an upper limit?

src/services/signer_service.ts
fabioberger: Added a TODO.

src/services/signer_service.ts
fabioberger: Yep, good catch. `pending` is def. safer here, since this code path will execute upon re-booting.

src/schemas/meta_transaction_quote_request_schema.json
dekz: Yeah looks like it.

src/runners/http_service_runner.ts
dekz: `index.ts` is ran as a default "everything", used for Launch Kit and local to bring everything up. 

`http_service_runner` is staking/sra/swap quote which we run separately and scale up. So if it lands in here it will be in every available HTTP endpoint.

If we're limiting this endpoint to just 1 then it shouldn't be in `http_service_runner`.

src/services/signer_service.ts
dekz: Yes, as in should we validate this in terms of a gas oracle (ethgasstation for example) so its not submitted at 200 gwei when fast is 8 gwei.

src/services/meta_transaction_service.ts
```diff
@@ -0,0 +1,188 @@
...
+    }
+    public async calculateMetaTransactionQuoteAsync(
+        params: CalculateMetaTransactionQuoteParams,
+    ): Promise<GetMetaTransactionQuoteResponse> {
+        const { takerAddress, sellAmount, buyAmount, swapQuote, price } = await this.calculateMetaTransactionPriceAsync(
+            params,
+        );
+
+        const floatGasPrice = swapQuote.gasPrice;
+        const gasPrice = floatGasPrice.integerValue(BigNumber.ROUND_UP);
```
dekz, line 113: This is in wei, so `8500000000` for example is 8.5 gwei. We want the equivalent of `Math.ceil(8500000000 * 1e-9) * 1e9`

src/utils/utils.ts
```diff
@@ -72,4 +72,25 @@ export const utils = {
...
         return chunkedItems;
     },
+    convertAmountToBigNumber(value: string | number | BigNumber): BigNumber {
+        const num = value || 0;
+        const isBigNumber = BigNumber.isBigNumber(num);
+        if (isBigNumber) {
+            return num as BigNumber;
+        }
+
+        if (_.isString(num) && (num.indexOf('0x') === 0 || num.indexOf('-0x') === 0)) {
```
dekz, line 82: when is `-0x` valid?

src/services/signer_service.ts
fabioberger: Added in a check to make sure gasPrice isn't more than 3x ethGasStation fast.

src/services/meta_transaction_service.ts
fabioberger: Good catch!

src/utils/utils.ts
fabioberger: Good question, this was some copy-pasta from `web3-wrapper`, which in turn was copy-pasta'd from `ethereumjs-utils`. I've gone ahead and imported it from the `web3-wrapper` package instead of copying this code once again.

src/app.ts
```diff
@@ -66,12 +73,14 @@ export async function getDefaultAppDependenciesAsync(
...
         swapService,
+        metaTransactionService,
         provider,
         websocketOpts,
     };
 }
 /**
- * starts the app with dependencies injected
+ * starts the app with dependencies injected. This entry-point is used when running a single instance 0x API
+ * deployment and in tests. It is not used in production deployments where scaling is required.
```
fragosti, line 83: Is it used for Kovan FWIW.

src/handlers/meta_transaction_handlers.ts
```diff
@@ -0,0 +1,175 @@
...
+                    {
+                        field: 'token',
+                        code: ValidationErrorCodes.ValueOutOfRange,
+                        reason: e.message,
+                    },
+                ]);
+            }
+            logger.info('Uncaught error', e);
+            throw new InternalServerError(e.message);
+        }
```
fragosti, line 159: Is there a way to DRY up this error handling code? See similar variations elsewhere.

src/handlers/signer_handlers.ts
```diff
@@ -0,0 +1,111 @@
...
+    private readonly _signerService: SignerService;
+    public static rootAsync(_req: express.Request, res: express.Response): void {
+        const message = `This is the root of the Signer API.`;
+        res.status(HttpStatus.OK).send({ message });
+    }
+    constructor(signerService: SignerService) {
+        this._signerService = signerService;
+    }
+    public async submitZeroExTransactionIfWhitelistedAsync(req: express.Request, res: express.Response): Promise<void> {
+        const apiKey = req.header('0x-api-key');
```
fragosti, line 25: This is probably fine here for now, but this sort of stuff is usually handled in a middle-ware for the relevant sub-APIs.

src/runners/signer_runner.ts
```diff
@@ -0,0 +1,47 @@
...
+         * POST Transaction endpoint takes a signed 0x Transaction and sends it to Ethereum
+         * for execution via `executeTransaction`.
+         *
+         * https://0x.org/docs/guides/v3-specification#executing-a-transaction
+         */
+        app.post(
+            `${META_TRANSACTION_PATH}/fill`,
+            asyncHandler(handlers.submitZeroExTransactionIfWhitelistedAsync.bind(handlers)),
+        );
+
```
fragosti, line 32: Should it be called `/submit`? Technically we could use this for any `ZeroExTransaction` in the future right?

src/runners/signer_runner.ts
```diff
@@ -0,0 +1,47 @@
...
+        const handlers = new SignerHandlers(signerService);
+        /**
+         * POST Transaction endpoint takes a signed 0x Transaction and sends it to Ethereum
+         * for execution via `executeTransaction`.
+         *
+         * https://0x.org/docs/guides/v3-specification#executing-a-transaction
+         */
+        app.post(
+            `${META_TRANSACTION_PATH}/fill`,
+            asyncHandler(handlers.submitZeroExTransactionIfWhitelistedAsync.bind(handlers)),
```
fragosti, line 30: Why is this is not in `SignedHandlers`?

src/services/meta_transaction_service.ts
```diff
@@ -0,0 +1,191 @@
...
+                sellTokenAddress,
+                buyAmount,
+                assetSwapperOpts,
+            );
+        } else {
+            throw new Error('sellAmount or buyAmount required');
+        }
+
+        const makerAssetAmount = swapQuote.bestCaseQuoteInfo.makerAssetAmount;
+        const totalTakerAssetAmount = swapQuote.bestCaseQuoteInfo.totalTakerAssetAmount;
```
fragosti, line 78: Some of this is copy-pastad from the SwapService. Does it make sense to clean up?

src/services/signer_service.ts
```diff
@@ -0,0 +1,215 @@
...
+    public async validateZeroExTransactionFillAsync(
+        zeroExTransaction: ZeroExTransactionWithoutDomain,
+        signature: string,
+    ): Promise<BigNumber> {
+        // Verify 0x txn won't expire in next 60 seconds
+        // tslint:disable-next-line:custom-no-magic-numbers
+        const sixtySecondsFromNow = new BigNumber(+new Date() + 60);
+        if (zeroExTransaction.expirationTimeSeconds <= sixtySecondsFromNow) {
+            throw new Error('zeroExTransaction expirationTimeSeconds in less than 60 seconds from now');
+        }
```
fragosti, line 81: Should this be in a constant?

src/utils/service_utils.ts
```diff
@@ -0,0 +1,132 @@
...
+import { Web3Wrapper } from '@0x/web3-wrapper';
+
+import { CHAIN_ID, FEE_RECIPIENT_ADDRESS } from '../config';
+import { DEFAULT_TOKEN_DECIMALS, ONE_SECOND_MS, PERCENTAGE_SIG_DIGITS, ZERO } from '../constants';
+import { logger } from '../logger';
+import { GetSwapQuoteResponseLiquiditySource } from '../types';
+import { orderUtils } from '../utils/order_utils';
+import { findTokenDecimalsIfExists } from '../utils/token_metadata_utils';
+
+export const serviceUtils = {
```
fragosti, line 19: 👍

src/app.ts
fabioberger: Hence: "It is not used in production deployments **where scaling is required**" ;)

src/runners/signer_runner.ts
fabioberger: Hm, no other handler has these route definitions. They typically go in `*_router.ts` files but signing doesn't have one since it's so lightweight.

src/services/signer_service.ts
fabioberger: The error or what exactly?

src/handlers/signer_handlers.ts
fabioberger: Can you point me to this pattern being used elsewhere in the codebase?

src/handlers/meta_transaction_handlers.ts
fabioberger: Yep, moved into a utils.

src/services/meta_transaction_service.ts
fabioberger: That is true, but I don't think it makes sense to "clean up" these similarities since they will diverge once we add in re-couping gas costs logic.

src/handlers/signer_handlers.ts
fragosti: Ah I guess not in this codebase specifically but I've seen it in other code-bases (ie. it's basically an auth-header, where certain routes are authorized etc...). This works fine though.

src/utils/service_utils.ts
```diff
@@ -129,4 +132,40 @@ export const serviceUtils = {
             breakdown,
         );
     },
+
+    handleSwapThrownErrors(e: Error, buyAmount: BigNumber | undefined): void {
+        // If this is already a transformed error then just re-throw
+        if (isAPIError(e)) {
+            throw e;
+        }
```
fragosti, line 140: nice.

PR #36, opened 2019-12-11
improve runtime error handling

### LINKED ISSUE #35: Runtime errors go straight to stderr
## Current behaviour
* Startup errors *are* caught and logged by pino, but errors thrown from route handling just go straight to stderr.
* The thrown errors get sent as HTML string responses with the full stack trace to the client (maybe this is only in dev mode?)

## Desired behaviour
* All errors should be logged by pino in the correct format
* Not send the stack trace to the client, instead send a nice error object e.g.
```
  res: {
      "statusCode": 500,
      "statusMessage": "Internal Server Error"
    }
```

## Context
Errors aren't actually being caught and logged appropriately. we have a bunch of generic catchers like these:

```
process.on('uncaughtException', err => {
    logger.error(err);
    process.exit(1);
});


process.on('unhandledRejection', err => {
    if (err) {
        logger.error(err);
    }
});
```
which are *intended* to produce logs through pino like this:
```
log { level: 50, ... }
```

instead it looks like they aren't being caught and go to stdout directly like this:
```
log: at new QueryFailedError (/usr/src/app/src/error/QueryFailedError.ts:9:9) kubernetes.namespace_name:api kubernetes.labels.app:api-staging-mainnet @timestamp:December 9th 2019, 18:52:23.916 stream:stderr time:December 9th 2019, 18:52:23.916
```

xianny:
Resolves #35

Before:

Screen Shot 2019-12-11 at 11 20 06 AM

After:
Screen Shot 2019-12-11 at 11 19 55 AM

Tested a bit on Kovan and staging, but not exhaustively.

### DIFF
--- a/src/middleware/error_handling.ts
+++ b/src/middleware/error_handling.ts
@@ -15,11 +15,16 @@ import { logger } from '../logger';
  * Wraps an Error with a JSON human readable reason and status code.
  */
 export function generateError(err: Error): ErrorBodyWithHTTPStatusCode {
+    // handle named errors
     if ((err as any).isAPIError) {
         const apiError = err as APIBaseError;
         const statusCode = apiError.statusCode;
+
+        // populate more information for BAD_REQUEST errors
         if (apiError.statusCode === HttpStatus.BAD_REQUEST) {
             const badRequestError = apiError as BadRequestError;
+
+            // populate validation error information
             if (badRequestError.generalErrorCode === GeneralErrorCodes.ValidationError) {
                 const validationError = badRequestError as ValidationError;
                 return {
@@ -31,6 +36,7 @@ export function generateError(err: Error): ErrorBodyWithHTTPStatusCode {
                     },
                 };
             } else {
+                // if not a validation error, populate the error body with standard bad request text
                 return {
                     statusCode,
                     errorBody: {
@@ -40,6 +46,8 @@ export function generateError(err: Error): ErrorBodyWithHTTPStatusCode {
                 };
             }
         } else {
+            // all named errors that are not BAD_REQUEST
+            // preserve the statusCode and populate the error body with standard status text
             return {
                 statusCode,
                 errorBody: {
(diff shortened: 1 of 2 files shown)


src/middleware/error_handling.ts
```diff
@@ -70,11 +79,15 @@ export function errorHandler(
...
-        return next(err);
+
+    const { statusCode, errorBody } = generateError(err);
+    res.status(statusCode).send(errorBody);
+
+    // If the error is an internal error, log it with the stack!
+    // All other error responses are logged as part of request logging
+    if (statusCode === HttpStatus.INTERNAL_SERVER_ERROR) {
+        // hack (xianny): typeorm errors contain the SQL query which breaks the docker char limit and subsequently breaks log parsing
+        logger.error({ ...err, query: undefined});
```
xianny: Right now it seems like this is the only field that breaks log parsing. If something else breaks it, let's look into truncating JSON objs at that time. Feels like it will take too long to be worth it for now, unless anyone has an easy suggestion to implement?

src/middleware/error_handling.ts
```diff
@@ -70,11 +79,15 @@ export function errorHandler(
     if (res.headersSent) {
         return next(err);
     }
-    if ((err as any).isAPIError || (err as any).statusCode) {
```
xianny: Moved all the `if` statements into `generateError` so we can reason about error cases in one function. It SHOULD be the same logic with the added fallback case of `500`.

src/middleware/error_handling.ts
```diff
@@ -46,13 +47,21 @@ export function generateError(err: Error): ErrorBodyWithHTTPStatusCode {
                 },
             };
         }
+    } else if ((err as any).statusCode) {
+        return {
+            statusCode: HttpStatus.BAD_REQUEST,
+            errorBody: {
+                reason: err.message,
+            },
```
fragosti: Don't fully understand this reasoning.

src/middleware/error_handling.ts
fragosti: This is fine. Might be good to link the actual issue which I believe is a `fluentd` issue.

src/middleware/error_handling.ts
xianny: Previously:

```
if (isAPIError OR err.statusCode) {
    // generateError
    if (isAPIError) {
        do x;
    } else { // effectively means (if err.statusCode)
        do y; // generic HttpStatus.BAD_REQUEST
    }
}
```

Now:
```
if (isAPIerror) {
    do x;
} else if (err.statusCode) {
    do y;
} else {
    return generic 500;
}
```

So it's the same as before, but with the added fallback case if generic 500.

xianny: discussed on Slack. We realised the existing code was borked. If an error has a statuscode then it's a named error and the statuscode should be preserved. Otherwise it's a generic error and should be treated as a generic 500.

PR #43, opened 2019-12-14
Return 404 when user requests a non-existent staking pool id

xianny:
Currently returns 500 with reason:
```
{
  "reason": "Could not find pool with pool_id 600"
}
```

Should be a 404; 500 should be reserved for critical errors that need our attention


xianny:
@fragosti , do we want to do this also for:

```
GET /staking/epochs
GET /staking/stats
```

Or should it be considered a critical error if those are missing?

PR #44, opened 2019-12-16
fixed formatting on stats query

alexkroeger:
Fixed the formatting of the overall stats query (had to eliminate extra semicolons).

### DIFF
--- a/src/queries/staking_queries.ts
+++ b/src/queries/staking_queries.ts
@@ -380,10 +380,11 @@ export const allTimePoolRewardsQuery = `
     GROUP BY 1;
 `;
 
-export const allTimeStatsQuery = `;
-    SELECT; SUM(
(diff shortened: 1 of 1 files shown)


johnrjj:
Is there any way we can expand this query ?

Right now a `GET /staking/stats` returns 
```
{
    "allTime": {
        "totalRewardsPaidInEth": 0.1122
    }
}
```
The ui requires something like:

```
{
    "allTime": {
        "totalRewardsPaidInEth": 0.1122,
        "totalZrxStaked": 1.000
    },
    "currentEpoch": {
        "totalRewardsPaidInEth": 0.1122,
        "totalZrxStaked": 1.000
    }
}
```

tl;dr can we add `totalZrxStaked` to `allTime` and add a `currentEpoch` field with the same properties?

PR #45, opened 2019-12-16
Added current epoch and next epoch pool stats to pools/id endpoint, a…

alexkroeger:
Added `PoolWithStats` info to the pools/`id` endpoint.

Added timestamps to the pool-epoch rewards array for graphing.

### DIFF
--- a/src/types.ts
+++ b/src/types.ts
@@ -109,9 +109,9 @@ export interface RawPoolEpochRewards {
     starting_block_timestamp: string;
     starting_block_number: string;
     starting_transaction_index: string;
-    ending_block_number?: string;
-    ending_timestamp?: string;
-    ending_transaction_hash?: string;
+    ending_block_number: string;
+    ending_timestamp: string;
+    ending_transaction_hash: string;
 }
 
 export interface PoolMetadata {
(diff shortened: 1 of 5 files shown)


src/types.ts
```diff
@@ -109,9 +109,9 @@ export interface RawPoolEpochRewards {
     starting_block_timestamp: string;
     starting_block_number: string;
     starting_transaction_index: string;
-    ending_block_number?: string;
-    ending_timestamp?: string;
-    ending_transaction_hash?: string;
+    ending_block_number: string;
+    ending_timestamp: string;
+    ending_transaction_hash: string;
```
fragosti: hmmm how? we use this type for other queries too, where these can be undefined / null.

johnrjj:
I'm getting `QueryFailedError: bind message supplies 1 parameters, but prepared statement "" requires 0` on `GET /staking/pools/1` when running locally

johnrjj:
Pushed a fix

johnrjj:
`currentEpochStats.operatorShare` and `nextEpochStats.operatorShare` are null , is that expected?

src/types.ts
alexkroeger: rewards only accrue for a finished epoch (known end time)

alexkroeger:
> `currentEpochStats.operatorShare` and `nextEpochStats.operatorShare` are null , is that expected?

I'm seeing values in the query. Can you give a pool ID, network that you're using?

PR #47, opened 2019-12-18
Fix/swap quote schema validation

### LINKED ISSUE #51: Add request params validation and throw 400 on invalid requests
About
===

Currently a `GET` request with no query parameters to https://api.0x.org/swap/quote will cause the API to return `500` with an exception:
```
{"reason":"Cannot read property 'startsWith' of undefined"}
```

It would be best to add query parameter validation, and throw `400` with a helpful message back to the requester specifying missing / invalid parameters.

dekz:
Had to do more of a roundabout schema to avoid a generic `is not any of [subschema 0],[subschema 1]`. This can still occur with `/wholeNumberSchema`.

The typings also do not seem to allow this schema, so `as any` is required for now.

Resolves #51

### DIFF
--- a/src/schemas/swap_quote_request_schema.json
+++ b/src/schemas/swap_quote_request_schema.json
@@ -9,6 +9,12 @@
         },
         "takerAddress": {
             "$ref": "/addressSchema"
+        },
+        "gasPrice": {
+            "$ref": "/wholeNumberSchema"
+        },
+        "slippage": {
+            "$ref": "/numberSchema"
         }
     },
     "required": ["sellToken", "buyToken"],
--- a/src/schemas/swap_quote_request_schema.json
+++ b/src/schemas/swap_quote_request_schema.json
@@ -1,0 +1,36 @@
+{
+    "id": "/SwapQuoteRequestSchema",
+    "properties": {
+        "sellToken": {
+            "type": "string"
+        },
+        "buyToken": {
+            "type": "string"
+        },
+        "takerAddress": {
+            "$ref": "/addressSchema"
+        }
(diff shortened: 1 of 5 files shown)


src/utils/schema_utils.ts
```diff
@@ -9,47 +8,44 @@ const schemaValidator = new SchemaValidator();
...
-                validationResult.errors,
-                (schemaValidationError: SchemaValidationError) =>
-                    schemaValidationErrorToValidationErrorItem(schemaValidationError),
+            const validationErrorItems = validationResult.errors.map((schemaValidationError: SchemaValidationError) =>
+                schemaValidationErrorToValidationErrorItem(schemaValidationError),
             );
             throw new ValidationError(validationErrorItems);
         }
     },
+    addSchema(schema: Schema): void {
```
xianny: curious, what does this do?

src/schemas/swap_quote_request_schema.json
```diff
@@ -0,0 +1,42 @@
...
+    "id": "/SwapQuoteRequestSchema",
+    "properties": {
+        "sellToken": {
+            "type": "string"
+        },
+        "buyToken": {
+            "type": "string"
+        },
+        "takerAddress": {
+            "$ref": "/addressSchema"
```
xianny: does the `addSchema` thing allow us to reference `/addressSchema`'s relative path from `json-schemas`?

dekz: Yes exactly, unused now, probably needed in the future. There was an open question on "how this is done".

PR #48, opened 2019-12-18
Implement basic /swap/tokens endpoint

dekz:
Basic functionality for lighter clients to discover a list of all tokens available and their symbol.

This does not check for any available liquidity in either the database or in bridges. 

Rather than have a list of 2 tokens in the whitelist (when `WHITELIST_ALL_TOKENS` is unspecified) use the token metadata to populate this list.

### DIFF
--- a/src/handlers/swap_handlers.ts
+++ b/src/handlers/swap_handlers.ts
@@ -9,9 +9,9 @@ import { InternalServerError, RevertAPIError, ValidationError, ValidationErrorCo
 import { logger } from '../logger';
 import { isAPIError, isRevertError } from '../middleware/error_handling';
 import { SwapService } from '../services/swap_service';
-import { GetSwapQuoteRequestParams } from '../types';
+import { TokenMetadatasForChains } from '../token_metadatas_for_networks';
+import { ChainId, GetSwapQuoteRequestParams } from '../types';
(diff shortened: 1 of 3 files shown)


src/handlers/swap_handlers.ts
```diff
@@ -76,6 +76,14 @@ export class SwapHandlers {
...
         }
     }
+    // tslint:disable-next-line:prefer-function-over-method
+    public async getSwapTokensAsync(_req: express.Request, res: express.Response): Promise<void> {
+        const tokens = TokenMetadatasForChains.map(tm => ({
+            symbol: tm.symbol,
+            address: tm.tokenAddresses[CHAIN_ID],
+        }));
+        res.status(HttpStatus.OK).send(tokens);
+    }
```
fragosti: Just sending an array in a JSON response can mean trouble later on. Can we return this like 
```
{ tokens: [] }
```

fragosti:
@dekz can we use the pagination abstraction for this like we do in SRA?https://github.com/0xProject/standard-relayer-api/blob/master/http/v3.md#pagination

src/handlers/swap_handlers.ts
dekz: ahh you mean in future changes to the endpoint in that it is difficult to make non-breaking additions?

PR #50, opened 2019-12-18
set up app tests

xianny:
**====== EDIT ======**
Revisited this and realised we needed more refactoring to allow dependency injection of testing mocks.

Pretty fiddly refactor, so I deployed to staging and confirmed smoke tests all pass. 

Changes:
- Cleaned up `/runners` and separated the logic into functions and executions. Removed the duplicate logic in `/index`
- Re-used `/runners` to set up the app in `/app.ts`
- `/index.ts` now uses `/app.ts` to start the API instead of duplicating the logic in `/runners`, so we can be sure that all entrypoints are running the same things
- tests use `/app.ts` to inject dependencies and start the API
- removed `/services/*_http_service.ts` because they aren't real services, they just set up routing. Moved the logic to `/routers`.

Questions:
- should we explicitly declare `/index` as the default entry point in package.json? we don't have one right now

**====== END EDIT ======**

Set up a basic test framework:
- set up CI and yarn scripts
- migrate 0x contracts
- extract express app to prepare for dependency injection
- test that GET /sra/orders returns 200 status

After merging this, will open more smaller PRs with more tests.

I tried splitting the tests into separate parallel CI jobs, but spinning up the env for each job actually takes more time (0:58 vs 4:03) When/if tests start taking longer, we can split up the jobs again. 
Screen Shot 2020-01-08 at 9 19 20 AM


test/app_test.ts
```diff
@@ -0,0 +1,67 @@
...
+    CHAIN_ID: 1337,
+    ENVIRONMENT: 'unittest',
+};
+
+let app: express.Application;
+
+let web3Wrapper: Web3Wrapper;
+let provider: Web3ProviderEngine;
+let accounts: string[];
+// let contractAddresses: ContractAddresses;
```
dekz:
```suggestion

```

test/app_test.ts
```diff
@@ -0,0 +1,67 @@
...
+let blockchainLifecycle: BlockchainLifecycle;
+
+describe('app test', () => {
+    before(async () => {
+        // start ganache and run contract migrations
+        const ganacheConfigs = {
+            total_accounts: 10,
```
dekz: I don't think you need this to be set

src/app.ts
```diff
@@ -0,0 +1,70 @@
...
+    }
+    const orderBookService = new OrderBookService(connection, meshClient);
+    // tslint:disable-next-line:no-unused-expression
+    new SRAHttpService(app, orderBookService);
+    try {
+        // Quote Service
+        const orderStore = new OrderStoreDbAdapter(orderBookService);
+        const orderProvider = new OrderBookServiceOrderProvider(orderStore, orderBookService);
+        const orderBook = new Orderbook(orderProvider, orderStore);
+        const provider = providerUtils.createWeb3Provider(config.ETHEREUM_RPC_URL);
```
dekz: I'm not 100% certain if the `shouldUseInProcessGanache` also opens up ports for RPC access. We may need to inject the provider if we want to get the entire app to use the in process ganache.

None of the tests currently hit the RPC URL (as it's just get orders) but if they did I'm not certain how it would respond.

src/app.ts
```diff
@@ -0,0 +1,70 @@
...
+import { SwapService } from './services/swap_service';
+import { WebsocketService } from './services/websocket_service';
+import { OrderStoreDbAdapter } from './utils/order_store_db_adapter';
+import { providerUtils } from './utils/provider_utils';
+
+// tslint:disable:no-unused-expression
+/**
+ * returns the app
+ */
+export async function getAppAsync(config: any = defaultConfig): Promise<express.Application> {
```
fragosti: can we make the config typed but use `any` in the tests if it is even necessary? I don't see this being used anywhere.

src/db_connection.ts
```diff
@@ -2,9 +2,10 @@ import { Connection, createConnection } from 'typeorm';
 
 import { config } from './ormconfig';
 
+const connection = createConnection(config);
 /**
  * Creates the DB connnection to use in an app
  */
 export async function getDBConnectionAsync(): Promise<Connection> {
-    return createConnection(config);
+    return connection;
```
fragosti: Why this change?

src/app.ts
fragosti: might be worth thinking about how to apply this pattern to the other "apps" in the `runners/` directory.

src/services/staking_data_service.ts
```diff
@@ -120,7 +120,9 @@ export class StakingDataService {
 
         const currentEpochPoolStats = stakingUtils.getEpochPoolStatsFromRaw(rawCurrentEpochPoolStats[0]);
         const nextEpochPoolStats = stakingUtils.getEpochPoolStatsFromRaw(rawNextEpochPoolStats[0]);
-        const pool7dProtocolFeesGenerated = stakingUtils.getPoolProtocolFeesGeneratedFromRaw(rawPoolSevenDayProtocolFeesGenerated[0]);
+        const pool7dProtocolFeesGenerated = stakingUtils.getPoolProtocolFeesGeneratedFromRaw(
```
xianny: just a prettier change

src/db_connection.ts
xianny: `createConnection` actually creates a connection pool. It's intended to be called once per app and continue running in the background. Previous code was creating connection pools every time this method is called.

https://github.com/typeorm/typeorm/blob/master/docs/connection.md

src/index.ts
```diff
@@ -1,66 +1,17 @@
...
-    } catch (err) {
-        logger.error(err);
-    }
-    // OrderWatcher Service
-    if (meshClient) {
-        const orderWatcherService = new OrderWatcherService(connection, meshClient);
-        await orderWatcherService.syncOrderbookAsync();
-    } else {
-        logger.warn('API starting without a connection to mesh');
-    }
```
xianny: All of this was duplicated in the runners. Moved there and called as a function instead.

src/services/sra_http_service.ts
```diff
@@ -1,20 +0,0 @@
...
-
-// tslint:disable-next-line:no-unnecessary-class
-export class SRAHttpService {
-    constructor(app: core.Express, orderBook: OrderBookService) {
-        app.use(cors());
-        app.use(bodyParser.json());
-        app.use(SRA_PATH, createSRARouter(orderBook));
-        app.use(errorHandler);
-    }
-}
```
xianny: changed it to a function instead of a class, and moved it to `/routers`. It's not really a service, it's just setting up routes.

src/services/staking_http_service.ts
```diff
@@ -1,19 +0,0 @@
...
-
-// tslint:disable-next-line:no-unnecessary-class
-export class StakingHttpService {
-    constructor(app: core.Express, stakingDataService: StakingDataService) {
-        app.use(cors());
-        app.use(bodyParser.json());
-        app.use(STAKING_PATH, createStakingRouter(stakingDataService));
-        app.use(errorHandler);
-    }
-}
```
xianny: moved to `/routers`, similar to `sra_http_service.ts`

src/services/swap_http_service.ts
```diff
@@ -1,20 +0,0 @@
...
-
-// tslint:disable-next-line:no-unnecessary-class
-export class SwapHttpService {
-    constructor(app: core.Express, swapService: SwapService) {
-        app.use(cors());
-        app.use(bodyParser.json());
-        app.use(SWAP_PATH, createSwapRouter(swapService));
-        app.use(errorHandler);
-    }
-}
```
xianny: moved to `/routers`, same as `sra_http_service.ts`

src/app.ts
xianny: got rid of the `any`s

xianny:
Revisited this and realised we needed more refactoring to allow dependency injection of testing mocks.

Changes:
- Cleaned up `/runners` and separated the logic into functions and executions. Removed the duplicate logic in `/index`
- Re-used `/runners` to set up the app in `/app.ts`
- `/index.ts` now uses `/app.ts` to start the API instead of duplicating the logic in `/runners, so we can be sure that all entrypoints are running the same things
- tests use `/app.ts` to inject dependencies and start the API
- removed `/services/*_http_service.ts` because they aren't real services, they just set up routing. Moved the logic to `/routers`.


Questions:
- should we explicitly declare `/index` as the default entry point in package.json? we don't have one right now

src/db_connection.ts
fragosti: Yea but the method was only being called once.  I just don't want to encourage importing this and calling it directly, but rather calling it once and dependency injecting.

src/runners/http_service_runner.ts
```diff
@@ -28,41 +32,62 @@ process.on('unhandledRejection', err => {
...
+    dependencies: AppDependencies,
+    _config: { HTTP_PORT: string; ETHEREUM_RPC_URL: string },
+    _app?: core.Express,
+): Promise<Server> {
+    const app = _app || express();
     app.use(requestLogger());
-    const server = app.listen(config.HTTP_PORT, () => {
-        logger.info(`API (HTTP) listening on port ${config.HTTP_PORT}!\nConfig: ${JSON.stringify(config, null, 2)}`);
+    const server = app.listen(_config.HTTP_PORT, () => {
+        logger.info(`API (HTTP) listening on port ${_config.HTTP_PORT}!\nConfig: ${JSON.stringify(_config, null, 2)}`);
```
fragosti: Why the underscores?

src/runners/http_service_runner.ts
```diff
@@ -28,41 +32,62 @@ process.on('unhandledRejection', err => {
...
     });
-    const stakingDataService = new StakingDataService(connection);
-    // tslint:disable-next-line:no-unused-expression
-    new StakingHttpService(app, stakingDataService);
-    let meshClient;
+
+    // staking http service
+    const connection = dependencies.connection || (await getDBConnectionAsync());
```
fragosti: Shouldn't this be required then?

src/runners/http_service_runner.ts
```diff
@@ -28,41 +32,62 @@ process.on('unhandledRejection', err => {
...
+    const stakingDataService = dependencies.stakingDataService || new StakingDataService(connection);
+    configureStakingHttpRouter(app, stakingDataService);
+
+    // SRA http service
+    const orderBookService = dependencies.orderBookService || new OrderBookService(connection, dependencies.meshClient);
+    configureSRAHttpRouter(app, orderBookService);
+
```
fragosti: Hmm these too. Shouldn't they be required?

src/runners/http_service_runner.ts
dekz: I agree, I'd rather it fail if not present than create if undefined.

That way there is consistently one place where all dependencies are created/initialized.

src/db_connection.ts
xianny: It's kind of confusing to export a function that calls another function that's intended to be only called once. Exporting it encourages it to be called anywhere else.

This way `createConnection` **can** only be called once, and exporting the function as a getter encourages it to be used anywhere else in the app that a connection is needed.

src/routers/sra_router.ts
```diff
@@ -1,10 +1,23 @@
...
+import * as core from 'express-serve-static-core';
 
+import { SRA_PATH } from '../constants';
 import { SRAHandlers } from '../handlers/sra_handlers';
+import { errorHandler } from '../middleware/error_handling';
 import { OrderBookService } from '../services/orderbook_service';
 
-export const createSRARouter = (orderBook: OrderBookService): express.Router => {
+export const configureSRAHttpRouter = (app: core.Express, orderBook: OrderBookService): void => {
+    app.use(cors());
```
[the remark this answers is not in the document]
xianny: Yeah, this was happening before the refactor too. Was just copying code snippets around. On second thought, all of this seems unnecessary. I'm going to remove all the `configure*` methods.

src/runners/http_service_runner.ts
xianny: Removed all of the fallback code. Now there is a `getDefaultAppDependencies` function which is used both for the runners and for index.ts. Defaults depend on the config so we can treat the config as source of truth.

src/routers/sra_router.ts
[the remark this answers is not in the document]
fragosti: Not 100% what you mean but this architecture is there partially so that you can easily delete code in your own fork of the API (if you only want SRA for example).

src/runners/http_service_runner.ts
```diff
@@ -28,41 +28,56 @@ process.on('unhandledRejection', err => {
...
  * as well as adding orders to mesh.
+ * @param dependencies If no mesh client is supplied, the HTTP service will start without it.
+ *                     It will provide defaults for other params.
  */
-(async () => {
-    const connection = await getDBConnectionAsync();
-    const app = express();
+export async function runHttpServiceAsync(
+    dependencies: AppDependencies,
+    config: { HTTP_PORT: string; ETHEREUM_RPC_URL: string },
```
dekz: is `ETHEREUM_RPC_URL` needed here if the provider is in the AppDependencies?

src/runners/http_service_runner.ts
```diff
@@ -28,41 +28,56 @@ process.on('unhandledRejection', err => {
...
-        const orderStore = new OrderStoreDbAdapter(orderBookService);
-        const orderProvider = new OrderBookServiceOrderProvider(orderStore, orderBookService);
-        const orderBook = new Orderbook(orderProvider, orderStore);
-        const provider = providerUtils.createWeb3Provider(config.ETHEREUM_RPC_URL);
-        const swapService = new SwapService(orderBook, provider);
+
+    app.use(errorHandler);
+
+    // websocket service
+    if (dependencies.meshClient) {
```
dekz: Since we require `meshClient` for SRA `POST /orders` this should always be present. Should we fail if this doesn't exist? Or is this helpful for testing?

src/runners/order_watcher_service_runner.ts
```diff
@@ -1,19 +1,38 @@
...
+import { getDefaultAppDependenciesAsync } from '../app';
+import * as defaultConfig from '../config';
 import { logger } from '../logger';
 import { OrderWatcherService } from '../services/order_watcher_service';
+import { providerUtils } from '../utils/provider_utils';
 
+if (require.main === module) {
+    (async () => {
+        const provider = providerUtils.createWeb3Provider(defaultConfig.ETHEREUM_RPC_URL);
+        const { connection, meshClient } = await getDefaultAppDependenciesAsync(provider, defaultConfig);
```
dekz: thoughts on creating the provider in `getDefaultAppDependenciesAsync` ?

src/runners/http_service_runner.ts
xianny: ah good catch. removing!

src/runners/order_watcher_service_runner.ts
xianny: I had that at first, but ran into trouble with the provider for Ganache :/

src/runners/http_service_runner.ts
xianny: Yup this is helpful for testing. Previous implementation also had this if-clause and I preserved the behaviour because I didn't have a good reason to change it... **should** we fail if this doesn't exist? @fragosti thoughts?

src/runners/http_service_runner.ts
```diff
@@ -28,41 +28,56 @@ process.on('unhandledRejection', err => {
...
+    app.use(STAKING_PATH, createStakingRouter(dependencies.stakingDataService));
+
+    // SRA http service
+    app.use(SRA_PATH, createSRARouter(dependencies.orderBookService));
+
+    // swap/quote http service
+    if (dependencies.swapService) {
+        app.use(SWAP_PATH, createSwapRouter(dependencies.swapService));
+    } else {
+        logger.warn(`API running without swap service`);
```
fragosti: `logger.error` ?

src/runners/http_service_runner.ts
fragosti: Depends what you mean by fail. We need to balance keeping as many systems working as possible while also making it very visible when something fails.

I would keep the try-catches but make sure we `logger.error` and then alert on that?

src/runners/http_service_runner.ts
xianny: It was warn before but I can change it to error :)

src/handlers/root_handler.ts
```diff
@@ -1,4 +1,3 @@
-
```
xianny: prettier

PR #54, opened 2019-12-20
Query ERC20Token decimals if not mapped

dekz:
Query for decimals on the token contract if not mapped. Log if this is unmapped for future updates.

Sampling also has a requirement on decimals, so we can assume that if decimals is not present on the token then it is unlikely to be supported by the API/Bridge orders.

Note our ERC20 Token does not have `decimals` in the `ABI`, tbh I think we should just fix that and assume it always exists, if not it's the callers problem.

### DIFF
--- a/src/services/swap_service.ts
+++ b/src/services/swap_service.ts
@@ -194,7 +194,7 @@ export class SwapService {
                 constant: true,
                 inputs: [],
                 name: 'decimals',
-                outputs: [{ name: '', type: 'uint256' }],
+                outputs: [{ name: '', type: 'uint8' }],
                 payable: false,
                 stateMutability: 'view',
                 type: 'function',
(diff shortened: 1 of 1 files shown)


src/services/swap_service.ts
```diff
@@ -184,6 +185,32 @@ export class SwapService {
...
+    private async _fetchTokenDecimalsIfRequiredAsync(tokenAddress: string): Promise<number> {
+        // HACK(dekz): Our ERC20Wrapper does not have decimals as it is optional
+        // so we must encode this ourselves
+        let decimals = findTokenDecimalsIfExists(tokenAddress, CHAIN_ID);
+        if (!decimals) {
+            const decimalsEncoder = new AbiEncoder.Method({
+                constant: true,
+                inputs: [],
+                name: 'decimals',
+                outputs: [{ name: '', type: 'uint256' }],
```
BMillman19: just a small comment here, looks like this version of the erc20 abi defines the type of decimals as `uint8` https://github.com/ethereum/wiki/wiki/Contract-ERC20-ABI

dekz: This shouldn't matter as everything gets padded to 32 bytes. But updated to be consistent.

PR #59, opened 2020-01-08
stop throwing errors for non-ERC20/ERC721 asset data

### LINKED ISSUE #57: Calls to GET /sra/asset_pairs failing
Calls to `GET /sra/asset_pairs` return `500 server error` when there are non-ERC20/ERC721 orders in the orderbook.

I traced it to [this method](https://github.com/0xProject/0x-api/blob/master/src/utils/order_utils.ts#L185-L190) which throws an error if the `assetData` is not ERC20 or ERC721.

I discovered this when the `GET /sra/asset_pairs` smoke test started failing on Kovan, presumably because an ERC1155 order was added.

Screen Shot 2020-01-07 at 1 33 00 PM

https://assertible.com/dashboard#/services/8382da2e-a8fe-4f6d-9dd2-715561996521/tests/625405ec-1e0c-4f2b-a3cd-55e79deff759?resultView=response

## Solutions

a. only return ERC20/ERC721 orders. Filter the orderbook to stop throwing errors and ignore non-ERC20/ERC721 orders instead

b. start supporting the other token standards and return everything. are there sane defaults for the other token standards? We would need the equivalent of `DEFAULT_ERC20_ASSET` for all the standards we'd want to support: 

```
const DEFAULT_ERC20_ASSET = {
    minAmount: new BigNumber(0),
    maxAmount: MAX_TOKEN_SUPPLY_POSSIBLE,
    precision: DEFAULT_ERC20_TOKEN_PRECISION,
};

const erc20AssetDataToAsset = (assetData: string): Asset => ({
    ...DEFAULT_ERC20_ASSET,
    assetData,
});

const assetDataToAsset = (assetData: string): Asset => {
    const decodedAssetData = assetDataUtils.decodeAssetDataOrThrow(assetData);
    let asset: Asset;
    switch (decodedAssetData.assetProxyId) {
        case AssetProxyId.ERC20:
            asset = erc20AssetDataToAsset(assetData);
            break;
        case AssetProxyId.ERC721:
            asset = erc721AssetDataToAsset(assetData);
            break;
        default:
            throw errorUtils.spawnSwitchErr('assetProxyId', decodedAssetData.assetProxyId);
    }
    return asset;
};
```

xianny:
This PR adds support for other asset data types, and fixes #57

Initially I tried filtering out non-ERC20/ERC721 orders before retrieving asset pairs, then realised that the related functions in `order_utils.ts` were also used more generally by the orderbook when getting orders or checking for existence of asset pairs. Seems like it would be a non-trivial refactor, and it's unclear to me what the implications of that refactor would be, e.g. seems undesirable if `orderbook.getOrders` only returns ERC20/ERC721 orders.

I'm unsure of the appropriate default values for ERC1155, MultiAsset, StaticCall, and ERC20Bridge, so this PR is a bit of a WIP. 

Deployed to Kovan and it successfully returns ERC1155 asset pairs instead of throwing a server error. Compare with failing screenshot in #57:
Screen Shot 2020-01-08 at 1 26 55 PM

### DIFF
--- a/src/utils/order_utils.ts
+++ b/src/utils/order_utils.ts
@@ -41,19 +41,19 @@ const DEFAULT_ERC20_ASSET = {
 
 const DEFAULT_ERC1155_ASSET = {
     minAmount: new BigNumber(0),
-    maxAmount: new BigNumber(1),
+    maxAmount: MAX_TOKEN_SUPPLY_POSSIBLE,
     precision: 0,
 };
 
 const DEFAULT_MULTIASSET = {
     minAmount: new BigNumber(0),
-    maxAmount: new BigNumber(1),
+    maxAmount: MAX_TOKEN_SUPPLY_POSSIBLE,
     precision: 0,
 };
 
 const DEFAULT_STATIC_CALL = {
-    minAmount: new BigNumber(0),
-    maxAmount: new BigNumber(1),
+    minAmount: new BigNumber(1),
+    maxAmount: MAX_TOKEN_SUPPLY_POSSIBLE,
     precision: 0,
 };
 const proxyIdToDefaults: { [id: string]: Partial<Asset> } = {
--- a/src/utils/order_utils.ts
+++ b/src/utils/order_utils.ts
@@ -39,30 +39,42 @@ const DEFAULT_ERC20_ASSET = {
     precision: DEFAULT_ERC20_TOKEN_PRECISION,
 };
 
-const erc721AssetDataToAsset = (assetData: string): Asset => ({
-    ...DEFAULT_ERC721_ASSET,
-    assetData,
-});
+const DEFAULT_ERC1155_ASSET = {
+    minAmount: new BigNumber(0),
+    maxAmount: new BigNumber(1),
+    precision: 0,
+};
 
-const erc20AssetDataToAsset = (assetData: string): Asset => ({
-    ...DEFAULT_ERC20_ASSET,
-    assetData,
-});
+const DEFAULT_MULTIASSET = {
+    minAmount: new BigNumber(0),
+    maxAmount: new BigNumber(1),
+    precision: 0,
+};
+
+const DEFAULT_STATIC_CALL = {
+    minAmount: new BigNumber(0),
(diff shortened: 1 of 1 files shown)


src/utils/order_utils.ts
```diff
@@ -39,30 +39,42 @@ const DEFAULT_ERC20_ASSET = {
     precision: DEFAULT_ERC20_TOKEN_PRECISION,
 };
 
-const erc721AssetDataToAsset = (assetData: string): Asset => ({
-    ...DEFAULT_ERC721_ASSET,
-    assetData,
-});
+const DEFAULT_ERC1155_ASSET = {
+    minAmount: new BigNumber(0),
+    maxAmount: new BigNumber(1),
```
dorothy-zbornak:
```suggestion
    maxAmount: MAX_TOKEN_SUPPLY_POSSIBLE,
```

src/utils/order_utils.ts
```diff
@@ -39,30 +39,42 @@ const DEFAULT_ERC20_ASSET = {
...
+    precision: 0,
+};
+
+const DEFAULT_MULTIASSET = {
+    minAmount: new BigNumber(0),
+    maxAmount: new BigNumber(1),
```
dorothy-zbornak:
```suggestion
    maxAmount: MAX_TOKEN_SUPPLY_POSSIBLE,
```

src/utils/order_utils.ts
```diff
@@ -39,30 +39,42 @@ const DEFAULT_ERC20_ASSET = {
...
+    precision: 0,
+};
 
-const erc20AssetDataToAsset = (assetData: string): Asset => ({
-    ...DEFAULT_ERC20_ASSET,
-    assetData,
-});
+const DEFAULT_STATIC_CALL = {
+    minAmount: new BigNumber(0),
+    maxAmount: new BigNumber(1),
```
dorothy-zbornak:
```suggestion
    maxAmount: MAX_TOKEN_SUPPLY_POSSIBLE,
```

PR #86, opened 2020-01-28
Added number of fills to pool stats

alexkroeger:
Adds to Market maker stats:
- number of fills (to epoch stats, 7-day stats, and all time stats)
- share of stake (to epoch stats)
- share of fees (to epoch stats)
- share of fills (to epoch stats)

Testing:
- [x] Tested queries on hashalytics

### DIFF
--- a/src/queries/staking_queries.ts
+++ b/src/queries/staking_queries.ts
@@ -63,6 +63,7 @@ export const poolSevenDayProtocolFeesGeneratedQuery = `
         SELECT
             pi.pool_id
             , SUM(fe.protocol_fee_paid) / 1e18 AS protocol_fees
+            , COUNT(*) AS number_of_fills
         FROM events.fill_events fe
         LEFT JOIN events.blocks b ON b.block_number = fe.block_number
         LEFT JOIN staking.pool_info pi ON fe.maker_address = ANY(pi.maker_addresses)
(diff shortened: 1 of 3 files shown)


src/queries/staking_queries.ts
```diff
@@ -322,9 +336,8 @@ export const nextEpochPoolsStatsQuery = `
             , cs.zrx_staked
             , ts.total_staked
             , cs.zrx_staked / ts.total_staked AS share_of_stake
-            , fbp.protocol_fees AS protocol_fees_generated_in_eth
-            , tr.total_protocol_fees
-            , fbp.protocol_fees / tr.total_protocol_fees AS share_of_fees
+            , 0.00 AS total_protocol_fees_generated_in_eth
+            , 0 AS number_of_fills
```
kimpers: Are these supposed to be hardcoded to `0`?

src/queries/staking_queries.ts
```diff
@@ -390,9 +405,8 @@ export const nextEpochPoolStatsQuery = `
...
```
kimpers: Here too

src/queries/staking_queries.ts
alexkroeger: This was intentional. These are for the next epoch so putting a non-zero value here is misleading I think.

src/queries/staking_queries.ts
alexkroeger: See above

PR #92, opened 2020-02-03
Return sources breakdown for quote

dave4506:
Blocked by publish of asset-swapper with [PR](https://github.com/0xProject/0x-monorepo/pull/2465)

Adds a `source` entry to `/swap/v0/quote` endpoint with breakdown of quote liquidity sources.

Fixed to precision of four decimals for readability.

### DIFF
--- a/src/services/swap_service.ts
+++ b/src/services/swap_service.ts
@@ -15,7 +15,7 @@ import { TxData, Web3Wrapper } from '@0x/web3-wrapper';
 import { ASSET_SWAPPER_MARKET_ORDERS_OPTS, CHAIN_ID, FEE_RECIPIENT_ADDRESS } from '../config';
 import { DEFAULT_TOKEN_DECIMALS, PERCENTAGE_SIG_DIGITS, QUOTE_ORDER_EXPIRATION_BUFFER_MS } from '../constants';
 import { logger } from '../logger';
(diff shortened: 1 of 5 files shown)


dave4506:
Blocked by `@0x/asset-swapper` `4.1.1` release, but converted to:

```
sources: [
{
name: "Native",
proportion: "0.01963"
},
{
name: "Eth2Dai",
proportion: "0.5292"
},
{
name: "Uniswap",
proportion: "0.4512"
}
],
```

Did not add `amount` because proportions are calculated dependent to buy or sell amount, fear it can be confusing for developer and don't seem the other aggregator APIs do that.

src/services/swap_service.ts
```diff
@@ -136,7 +136,7 @@ export class SwapService {
         const breakdown: GetSwapQuoteResponseLiquiditySource[] = [];
         return Object.entries(sourceBreakdown).reduce((acc: GetSwapQuoteResponseLiquiditySource[], [source, percentage]) => {
             return [...acc, {
-                name: source,
+                name: source === 'Native' ? '0x' : source,
```
fragosti: @dekz and I were talking about updating this in swapper too. Is there an enum we can use that is either exported by asset-swapper or otherwise that will cause this compilation to fail if we change the string?

PR #100, opened 2020-02-07
[Kovan] Fix DevUtils Invalid opcode with a stipend

dekz:
Set the Kovan sampler to a specially deployed version with a stipend on the DevUtils contract. 

Fix is coming up in https://github.com/0xProject/0x-monorepo/pull/2474. Once published (Note the eth2dai change makes this time sensitive), we can remove the hack 

Rest is just formatting.


dekz:
https://github.com/0xProject/0x-api/pull/101

PR #106, opened 2020-02-13
Add POST /orders endpoint

fabioberger:
Currently we only support posting individual orders to 0x API and by extension Mesh. This is inefficient from an ETH RPC order validation perspective and can increase the latency experienced in seeing orders show up for takers. 

This PR adds a non-standard SRA endpoint to post multiple orders to 0x API and by extension Mesh. Since this is a new endpoint, it does not break backward compatibility and can simply be added to 0x API.


fabioberger:
deploy kovan

fabioberger:
deploy production

fragosti:
deploy staging

fabioberger:
deploy production

PR #111, opened 2020-02-19
Use HTTP when adding orders to mesh

xianny:
This should fix the `message too big` error that we are getting from Mesh. Chunks orders whenever we are adding orders to Mesh. This already existed in `meshUtils` so it's a tiny code change. 

## Context

We've been getting this error recently especially on start-up when syncing the orderbook with Mesh:

```
{
"err_string":"websocket: close 1009 (message too big): Frame size of 1984660 bytes exceeds maximum accepted frame size",
"level":"error","msg":"rpcSub returned an error",
"myPeerID":"16Uiu2HAm1Z7Z2WTRtgPWkhvfLgbLELKd8YsMyE8iaKXYZRoJdHd7",
"time":"2020-02-06T21:39:59Z"
}
```

 This error is caused by sending too many orders at once to Mesh. We need to always use chunking when calling `meshClient.addOrdersAsync`. The underlying cause is a [bug in the websocket client](https://github.com/theturtle32/WebSocket-Node/issues/359). 

Discussed with @fabioberger who suggested the best way to deal with it is to add the chunking on the 0x-api side. 

## Tests

Deployed on staging, smoke tests are passing


xianny:
deploy staging

dekz:
We are already chunking on a sync, see here: https://github.com/0xProject/0x-api/blob/master/src/services/order_watcher_service.ts#L26 so I thought we'd already be covered by this. I have a gut feeling that something more peculiar is happening on a disconnect. 

When I last tested this I was dropping the Mesh container and restarting it, this would force a reconnect and it would show errors. The underlying connection can report some debug info if `DEBUG="*"` is enabled which could help diagnose.

fragosti:
What @dekz said. Have you looked into using the HTTP endpoint?

xianny:
I'd like to try deploying this fix and monitoring for a couple of days. Our writeup in Quip [(link)](https://0xproject.quip.com/MA1bAOe0zA49/0xAPI-0x-Mesh-error) suggests that right now this happens occasionally when `POST sra/v3/orders` is used. This PR adds chunking for the handler for that endpoint specifically. 

I think it's also important to figure out if chunking everywhere fixes it. If it doesn't that indicates an unknown bug in the client. 

@dekz I'm with you on funky disconnects. I did some testing with the unit tests in https://github.com/0xProject/0x-mesh and noticed weird behaviour on disconnects with the client there too, but it might have just been my local env.

fragosti:
@xianny that `/orders` endpoint has only been out for like a week.

xianny:
Went down a bit of a rabbithole examining the websocket implementation in Mesh and web3.js and trying to rule out bugs there. 

Most likely cause of this problem (despite currently chunking by 100 orders at a time) is inconsistent order size, e.g. ERC1155 orders. There's no chunking built into the Mesh client, and it also doesn't make use of the fragmentation in the `'websocket'` implementation (https://github.com/theturtle32/WebSocket-Node/issues/359); so the frame is just dropped if it's [over 1MB](https://github.com/theturtle32/WebSocket-Node/blob/0b3d4a5eb253132b2219f6f22a420bfe4680e26a/lib/WebSocketClient.js#L42). For performance reasons we don't want to increase the max frame size.

Long-term, we should try to figure out how to use fragmentation in the Mesh client, or add automatic chunking by size of bytes. 

For now we can switch to using the HTTP endpoint which was implemented in Mesh v9 https://github.com/0xProject/0x-mesh/pull/658; just waiting till https://github.com/0xProject/0x-api/pull/113 is merged

xianny:
deploy staging

xianny:
deploy kovan

xianny:
Ready for review again!

# Changes
Adding orders to mesh should now only be done through `meshUtils.addOrdersToMeshAsync`. This method routes through websocket OR http depending on how many orders are being sent. The batch size threshold has a default value set in `constants.ts`; please review.

The implementation of `meshUtils.addOrdersViaHttp` mirrors the implementation in `WSClient` (Mesh) and `WebsocketProvider` (web3.js). It's all JSON-RPC transformations. 

# Testing
I deployed this on Kovan and successfully posted 10 orders (websocket) and 60 orders (http) via `POST /sra/v3/orders`. The code path for orderbook syncing is the same as for posting orders, so this should resolve the issue of dropping frames when syncing the orderbook.

# Potential issues

package.json
```diff
@@ -80,6 +80,7 @@
         "pg": "^7.12.1",
         "pino": "^5.13.5",
         "typeorm": "0.2.18",
+        "web3-providers": "^2.0.0-alpha.1",
```
fabioberger: This dep is unused.

src/constants.ts
```diff
@@ -15,6 +15,12 @@ export const ONE_SECOND_MS = 1000;
 export const ONE_MINUTE_MS = ONE_SECOND_MS * 60;
 export const TEN_MINUTES_MS = ONE_MINUTE_MS * 10;
 
+// Mesh can't handle a large amount of orders over websocket. If the orders submitted are greater than
+// MESH_WEBSOCKET_ORDER_BATCH_LIMIT, we send via HTTP instead
+export const MESH_WEBSOCKET_ORDER_BATCH_LIMIT = 50;
+export const MESH_WEBSOCKET_DEFAULT_ORDER_BATCH_SIZE = 10;
+export const MESH_HTTP_DEFAULT_ORDER_BATCH_SIZE = 100;
```
fabioberger: @xianny how were these numbers chosen? They all seem quite low.

fabioberger: Also, why not just send all of them over HTTP all the time? It seems like it'll always be more efficient given these settings.

src/services/order_watcher_service.ts
```diff
@@ -23,7 +24,11 @@ export class OrderWatcherService {
         // in the future we can attempt to retry these a few times. Ultimately if we
         // cannot validate the order we cannot keep the order around
         // Validate the local state and notify the order watcher of any missed orders
-        const { accepted, rejected } = await meshUtils.addOrdersToMeshAsync(this._meshClient, signedOrders);
+        const { accepted, rejected } = await meshUtils.addOrdersToMeshAsync(
+            this._meshHttpEndpoint,
+            this._meshClient,
```
fabioberger: It would cleaner if there was a `meshClient` class defined in this codebase that under-the-hood decided which transport to use, instead of passing around both an `HTTPEndpoint` and a `WSClient` everywhere. Otherwise, we don't have a clean separation of concerns.

src/utils/mesh_utils.ts
```diff
@@ -9,30 +9,71 @@ import {
...
+        // too many.
+        const shouldSendViaHttp = orders.length > MESH_WEBSOCKET_ORDER_BATCH_LIMIT;
+        const batchSize = shouldSendViaHttp
+            ? MESH_HTTP_DEFAULT_ORDER_BATCH_SIZE
+            : MESH_WEBSOCKET_DEFAULT_ORDER_BATCH_SIZE;
         const validationResults: ValidationResults = { accepted: [], rejected: [] };
         const chunks = _.chunk(orders, batchSize);
         for (const chunk of chunks) {
-            const results = await meshClient.addOrdersAsync(chunk as any);
+            const results = shouldSendViaHttp
```
fabioberger: This is another artifact of not abstracting away the transport to use within a simple Mesh client interface.

src/utils/mesh_utils.ts
```diff
@@ -9,30 +9,71 @@ import {
...
+
+        // send the request
+        const response = await axios({
+            method: 'post',
+            url: meshHttpEndpoint,
+            data,
+        });
+
+        // validate the response
+        utils.isValidJsonRpcResponseOrThrow(response.data, data as any);
```
fabioberger: Why is `data as any` needed here? If there is a good reason, please add a comment.

src/utils/utils.ts
```diff
@@ -9,4 +11,36 @@ export const utils = {
...
         }, initialMap);
     },
+
+    /**
+     * Executes JSON-RPC response validation
+     * Copied from https://github.com/ethereum/web3.js/blob/79a165a205074cfdc14f59a61c41ba9ef5d25172/packages/web3-providers/src/validators/JsonRpcResponseValidator.js
+     */
+    isValidJsonRpcResponseOrThrow: (response: any, payload = false): boolean => {
+        if (_.isObject(response)) {
+            if ((response as any).error) {
```
fabioberger: If we assign `response as any` to a variable at the top, we don't need to keep casting it everywhere.

src/utils/mesh_utils.ts
```diff
@@ -9,30 +9,71 @@ import {
...
             validationResults.accepted = [...validationResults.accepted, ...results.accepted];
             validationResults.rejected = [...validationResults.rejected, ...results.rejected];
         }
         return validationResults;
     },
+    // this logic mirrors the mesh RPC client and WebsocketProvider implementations
+    addOrdersViaHttp: async (
+        meshHttpEndpoint: string,
+        orders: SignedOrder[],
+        pinned: boolean = true,
```
fabioberger: Should be set to `false`. Anyone can submit orders to 0x API and if we pin them, someone could spam our Mesh node and cause our DB to fill up and stop accepting new orders. Our anti-ddos/spam prevention only acts on orders that are not pinned.

Once we implement 0x API keys, we can pin orders from trusted makers. But until then, we shouldn't pin any of them.

src/services/orderbook_service.ts
```diff
@@ -173,16 +174,21 @@ export class OrderBookService {
...
     }
     public async addOrdersAsync(signedOrders: SignedOrder[]): Promise<void> {
-        if (this._meshClient) {
-            const { rejected } = await this._meshClient.addOrdersAsync(signedOrders as any);
+        if (this._meshClient && this._meshHttpEndpoint) {
+            const { rejected } = await meshUtils.addOrdersToMeshAsync(
+                this._meshHttpEndpoint,
+                this._meshClient,
+                signedOrders,
+            );
```
fragosti: We should probably still add orders if we don't have the `meshHttpEndpoint` ENV var. This will also save us the headache of updating all launch-kit related things. Alternatively we can pin the version of 0x-api used in launch-kit.

package.json
fragosti: @xianny I think I got an error that this wasn't installed too. `rm -rf node_modules` fixed it.

README.md
```diff
@@ -64,6 +64,7 @@ To get a local development version of `0x-api` running:
 | `CHAIN_ID`              | Required. No default.                                           | The chain id you'd like your API to run on (e.g: `1` -> mainnet, `42` -> Kovan, `3` -> Ropsten, `1337` -> Ganache). Defaults to `42` in the API, but required for `docker-compose up`. |
 | `ETHEREUM_RPC_URL`      | Required. No default.                                           | The URL used to issue JSON RPC requests. Use `http://ganache:8545` to use the local ganache instance.                                                                                  |
 | `MESH_WEBSOCKET_URI`    | Required. Default for dev: `ws://localhost:60557`               | The URL pointing to the 0x Mesh node. A default node is spun up in `docker-compose up`                                                                                                 |
+| `MESH_HTTP_URI`         | Required. Default for dev: `http://localhost:60556`             | The URL pointing to the Mesh node's HTTP JSON-RPC endpoint. A default node is spun up in `docker-compose up`                                                                           |
```
fragosti: As noted in another comment, maybe this should be optional.

package.json
xianny: I was attempting to import the JsonRpcResponseValidator from here but that didn't work. Will remove this line.

src/constants.ts
xianny: Yeah, sending them over HTTP all the time would make sense. The batch size is kind of a stab in the dark TBH. I think ultimately we have to chunk by size of actual bytes, not number of orders. Setting it to 300 and sending all over HTTP for now.

src/utils/mesh_utils.ts
xianny: Removing

src/services/orderbook_service.ts
xianny: @fabioberger suggested adding all orders over HTTP, which I think makes sense, but that means requiring this env var.

xianny:
deploy kovan

README.md
```diff
@@ -59,18 +59,19 @@ To get a local development version of `0x-api` running:
...
+| `ETHEREUM_RPC_URL`      | Required. No default.                                                                        | The URL used to issue JSON RPC requests. Use `http://ganache:8545` to use the local ganache instance.                                                                                  |
+| `MESH_WEBSOCKET_URI`    | Required. Default for dev: `ws://localhost:60557`                                            | The URL pointing to the 0x Mesh node. A default node is spun up in `docker-compose up`                                                                                                 |
+| `MESH_HTTP_URI`         | Optional. Used when syncing the orderbook; defaults to websocket connection if not provided. | The URL pointing to the Mesh node's HTTP JSON-RPC endpoint. A default node is spun up in `docker-compose up`                                                                           |
```
fragosti: Nit: I would just say "Optional. No default." and then put the description part in the description column.

src/config.ts
```diff
@@ -40,6 +40,7 @@ export const ETHEREUM_RPC_URL = assertEnvVarType('ETHEREUM_RPC_URL', process.env
 export const MESH_WEBSOCKET_URI = _.isEmpty(process.env.MESH_WEBSOCKET_URI)
     ? 'ws://localhost:60557'
     : assertEnvVarType('MESH_WEBSOCKET_URI', process.env.MESH_WEBSOCKET_URI, EnvVarType.Url);
+export const MESH_HTTP_URI = process.env.MESH_HTTP_URI;
```
fragosti: you should still use the `assertEnvVarType` function right? if it isn't falsey.

src/utils/mesh_client.ts
```diff
@@ -0,0 +1,47 @@
...
+
+import { MESH_ORDERS_BATCH_SIZE } from '../constants';
+
+import { utils } from './utils';
+
+export class MeshClient extends WSClient {
+    public async addOrdersViaHttpAsync(orders: SignedOrder[], pinned: boolean = false): Promise<ValidationResults> {
+        if (_.isEmpty(this.httpURI)) {
+            return super.addOrdersAsync(orders, pinned);
+        } else {
```
fragosti: It should just fail right? The function name here is misleading if there is a case where it would post over websockets. You could also rename the function.

src/utils/mesh_client.ts
```diff
@@ -0,0 +1,47 @@
...
+            const validationResults: ValidationResults = { accepted: [], rejected: [] };
+            const chunks = _.chunk(orders, MESH_ORDERS_BATCH_SIZE);
+            chunks.forEach(async chunk => {
+                // format request payload
+                const data = {
+                    jsonrpc: '2.0',
+                    id: +new Date(),
+                    method: 'mesh_addOrders',
+                    params: [chunk, { pinned }],
```
fragosti: Do we need batching logic for HTTP?

src/utils/mesh_client.ts
xianny: renaming function

PR #120, opened 2020-02-26
Ignore certain addresses in the 0xApi

dekz:
Allows for certain addresses to be ignored when specified.
* makerAddress
* makerAssetData (including MultiAsset)
* takerAssetData (including MultiAsset)

There are 2 ways to use this:
* Ingress level (not ever saved to db, never present in SRA, or ever used): `IGNORED_ADDRESSES`
* `/swap/v0/quote` level (saved to db, returned in SRA, never used in quotes): `SWAP_QUOTE_IGNORED_ADDRESSES`


dekz:
This allows us to disable and immediately have the orders available. Rather than a re-sync to get them from Mesh. I did consider ignoring it at the Mesh layer and was on the fence about it. 

What tipped me in this direction was the temporality of certain ignored addresses. Though in some ignored addresses it would be "permanent".

src/config.ts
```diff
@@ -33,6 +34,20 @@ export const WHITELISTED_TOKENS: string[] | '*' = _.isEmpty(process.env.WHITELIS
     ? TokenMetadatasForChains.map(tm => tm.tokenAddresses[CHAIN_ID])
     : assertEnvVarType('WHITELIST_ALL_TOKENS', process.env.WHITELIST_ALL_TOKENS, EnvVarType.WhitelistAllTokens);
 
+// Ignored addresses. These are ignored at the ingress level and are never stored.
+export const IGNORED_ADDRESSES: string[] = _.isEmpty(process.env.IGNORED_ADDRESSES)
+    ? []
+    : assertEnvVarType('IGNORED_ADDRESSES', process.env.IGNORED_ADDRESSES, EnvVarType.AddressList);
+
+// Ignored addresses only for Swap Quote (still present in database and SRA).
+export const SWAP_QUOTE_IGNORED_ADDRESSES: string[] = _.isEmpty(process.env.SWAP_QUOTE_HIDDEN_ADDRESSES)
```
fragosti: nit: this can be just be `swap` as it affects the prices endpoint too I imagine.

src/config.ts
```diff
@@ -33,6 +34,20 @@ export const WHITELISTED_TOKENS: string[] | '*' = _.isEmpty(process.env.WHITELIS
     ? TokenMetadatasForChains.map(tm => tm.tokenAddresses[CHAIN_ID])
     : assertEnvVarType('WHITELIST_ALL_TOKENS', process.env.WHITELIST_ALL_TOKENS, EnvVarType.WhitelistAllTokens);
 
+// Ignored addresses. These are ignored at the ingress level and are never stored.
+export const IGNORED_ADDRESSES: string[] = _.isEmpty(process.env.IGNORED_ADDRESSES)
```
fragosti: `MESH_IGNORED_ADDRESSES` ?

fragosti:
You also need to document the new ENV vars.

dekz:
> The OrderBookService ~kind of fits the bill.  

This falls short of Websockets, which is just a translation proxy from the direct 0x Mesh feed.

PR #121, opened 2020-02-27
Align slippage parameters

dekz:
`slippagePercentage` in AssetSwapper really is more of an over-buy percentage. So instead of requesting to sell 100 tokens, we request to sell 120 tokens. The history of this parameter predates Sampler and was to fetch MORE 0x orders than necessary, in the event that some were cancelled or filled. There would be "backup" orders to fill. 

Unfortunately this has a few issues:
* It is not guaranteed to fetch MORE orders (the same order could be for 200 tokens as an example)
* It is not guaranteed to fetch DISTINCT orders (the same maker could've cancelled ALL orders)
* It messes with gas estimation (you estimate against the first few orders succeeding, not against the first first orders failing and the last few orders filling)
* On large fills it can fetch a lot of orders
* A user adjusting this `slippagePercentage` would have no effect on `bridgeSlippage`

I propose we align this parameter with bridge slippage and reduce it from 20% to 1%. So the effects are:
* We sample to sell up to 101 tokens
* This can be overwritten and we fetch more and allow for more bridge slippage
* OOG reverts due to orders being cancelled result in other classes of reverts (unfillable)

TODO:
- [ ] Update website documentation
- [ ] Upstream the changes to asset-swapper

### DIFF
--- a/src/config.ts
+++ b/src/config.ts
@@ -4,7 +4,13 @@ import { ERC20BridgeSource, SwapQuoteRequestOpts } from '@0x/asset-swapper';
 import { BigNumber } from '@0x/utils';
 import * as _ from 'lodash';
 
-import { DEFAULT_LOCAL_POSTGRES_URI, DEFAULT_LOGGER_INCLUDE_TIMESTAMP, NULL_ADDRESS, NULL_BYTES } from './constants';
+import {
+    DEFAULT_LOCAL_POSTGRES_URI,
+    DEFAULT_LOGGER_INCLUDE_TIMESTAMP,
+    DEFAULT_QUOTE_SLIPPAGE_PERCENTAGE,
+    NULL_ADDRESS,
+    NULL_BYTES,
+} from './constants';
 import { TokenMetadatasForChains } from './token_metadatas_for_networks';
 import { ChainId } from './types';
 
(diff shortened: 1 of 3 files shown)


dekz:
> Do we have any idea how often fills actually eat into the buffer?

Realistically these should be a form of OOG. We recently added a 20% buffer to gas estimation so we could be hiding this slightly now. Though a set of 0x Orders with a bridge "backup" like Kyber would still quickly run into this.

dekz:
> Then we re-expose it with a better name, and use it internally within the API to fine-tune reverts.

Yes I'll add this as a future TODO as we learn from the tweaking these params. If we see no increase in reverts then I feel it is safe to combine these params totally and only expose a `slippagePercentage`.

fragosti:
> Yes I'll add this as a future TODO as we learn from the tweaking these params. If we see no increase in reverts then I feel it is safe to combine these params totally and only expose a slippagePercentage.

I'm still thinking through this, but some things I'd like to avoid:
- Deviating in parameter meaning between parameter names in asset-swapper vs. 0x-api.
- Breaking changes to the API.

The only path forward I can think of:
- Keep `slippagePercentage` and be very clear about the definition in the docs.
- Change the default of `slippagePercentage` to something lower.
- Expose `bridgeSlippage` and document that well.

dekz:
https://github.com/0xProject/website/pull/248

PR #125, opened 2020-02-27
Try installing asset-swapper from gitpkg

fragosti:


fragosti:
Closing as it was only a test.

PR #128, opened 2020-03-03
Fix typo in accessing property

fragosti:
I'm having trouble with my local docker at the moment but this seems to be the cause. The validation code above is checking for undefined on `result` whereas we are then passing `results`.

Can someone test this out, or we can test it out on staging.


fragosti:
deploy staging

PR #133, opened 2020-03-13
Fallback path

dekz:
* Adds fallback path when 0x Native orders are used in the optimal path.
* Adds `guaranteedPrice` in the response now that we're potentially falling back with a path that has additional slippage. It will fill at this price (or better) or revert.
* Bumps the gas schedule to be more accurate with reality of filling via a bridge.
* Bumps default slippage to 3% during these volatile times.
* `gas` is now always returned as our faux estimate or a real estimate (whatever is largest)

# Comparison
### dev vs prod 
Dev has 3%, 1.5% bridgeSlippage, maxFallbackSlippage
Production has 1% slippage and no fallback

image

You can observe the fallback path and increase in slippage to 3% has a significant effect in these volatile times.

### dev vs staging (with fallback)
Dev has 3%, 1.5% bridgeSlippage, maxFallbackSlippage
Staging has 1.5%, 1.5%

image

You can observe the fallback path but a tight slippage isn't suitable in these volatile times. We don't often find a fallback path within 1.5% slippage range.

### DIFF
--- a/src/services/swap_service.ts
+++ b/src/services/swap_service.ts
@@ -99,11 +99,11 @@ export class SwapService {
             makerAssetAmount,
             totalTakerAssetAmount,
             protocolFeeInWeiAmount: protocolFee,
-            gas,
         } = attributedSwapQuote.bestCaseQuoteInfo;
         const {
-            makerAssetAmount: minMakerAssetAmount,
-            totalTakerAssetAmount: minTotalTakerAssetAmount,
+            makerAssetAmount: worstMakerAssetAmount,
+            totalTakerAssetAmount: worstTotalTakerAssetAmount,
+            gas,
         } = attributedSwapQuote.worstCaseQuoteInfo;
         const { orders, gasPrice, sourceBreakdown } = attributedSwapQuote;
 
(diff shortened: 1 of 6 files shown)


src/services/swap_service.ts
```diff
@@ -99,7 +99,12 @@ export class SwapService {
             makerAssetAmount,
             totalTakerAssetAmount,
             protocolFeeInWeiAmount: protocolFee,
+            gas,
```
dorothy-zbornak: Maybe should pull this from worst case, in case we decide to make them different.

src/services/swap_service.ts
```diff
@@ -99,7 +99,12 @@ export class SwapService {
...
         } = attributedSwapQuote.bestCaseQuoteInfo;
+        const {
+            makerAssetAmount: minMakerAssetAmount,
+            totalTakerAssetAmount: minTotalTakerAssetAmount,
```
dorothy-zbornak: If this was a buy, wouldn't this be a `maxTotalTakerAssetAmount`? Maybe it makes more sense to just prefix these with `worst` instead of `min`.

dekz:
deploy staging

PR #135, opened 2020-03-16
Fix swap service launch command

freeatnet:
I think I found a typo in the documentation, so here's a fix.


fragosti:
@freeatnet thank you!

PR #137, opened 2020-03-18
Fix excluded sources

dekz:
Needs to find the value of `CurveUsdcDaiUsdt` i.e `Curve_USDC_DAI_USDT` as that is what is used internally in asset-swapper and returned via the API. 

`excludedSources=CurveUsdcDaiUsdt` was returning `['CurveUsdcDaiUsdt']` and `excludedSources=Curve_USDC_DAI_USDT` was returning `[]`. `Native/0x` and `Kyber` etc are fine as they don't have different key/values.

### DIFF
--- a/src/handlers/swap_handlers.ts
+++ b/src/handlers/swap_handlers.ts
@@ -146,9 +146,13 @@ const findTokenAddressOrThrowApiError = (address: string, field: string, chainId
 };
 
 const parseStringArrForERC20BridgeSources = (excludedSources: string[]): ERC20BridgeSource[] => {
+    // Need to compare value of the enum instead of the key, as values are used by asset-swapper
+    // CurveUsdcDaiUsdt = 'Curve_USDC_DAI_USDT' is excludedSources=Curve_USDC_DAI_USDT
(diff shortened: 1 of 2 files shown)


src/handlers/swap_handlers.ts
```diff
@@ -146,9 +146,13 @@ const findTokenAddressOrThrowApiError = (address: string, field: string, chainId
 };
 
 const parseStringArrForERC20BridgeSources = (excludedSources: string[]): ERC20BridgeSource[] => {
+    // Need to compare value of the enum instead of the key, as values are used by asset-swapper
+    // CurveUsdcDaiUsdt = 'Curve_USDC_DAI_USDT' is excludedSources=Curve_USDC_DAI_USDT
     return excludedSources
         .map(source => (source === '0x' ? 'Native' : source))
```
fragosti: Didn't we change this in swapper too?

dekz: Asset-swapper does/should always use the value `Curve_USDC_DAI_USDT`

PR #138, opened 2020-03-19
Handle WETH and ETH unwrapping/wrapping  with swap quote

dave4506:
#### Unwrapping and Wrapping
You can now unwrap and wrap between ETH and WETH by requesting a quote:
- `sellToken=WETH` and `buyToken=ETH` for unwrapping
- `sellToken=ETH` and `buyToken=WETH` for wrapping

Providing either `sellAmount` and `buyAmount`  will give the same behavior because the "rate" between ETH<>WETH is always 1:1.

I have opted out of making the changes to support unwrapping and wrapping in asset-swapper because it seems out of scope and requires larger changes to AS that seem rather costly for such a feature.
- AS relies on token addresses and asset data to identify assets; with that we would need to introduce some ability to identity `ETH` between `WETH` to `SwapQuoter`'s interface. This would require a large change to AS.
- Behind the scenes, SwapQuoter will need to do quite a bit of work to skip the SOR logic. 
- With the eventual addition of exchange contract proxy pattern that allow for better UX around wrapping/unwrapping; seems to be the case this design will be a stop gap solution that that of a long term solution.

#### Throwing on `buyToken=ETH` 
0x API `/swap` considers `buyToken=ETH` as quotes where `buyToken=WETH`. Instead of the unexpected error, 0x API will throw a 400 error to reflect that 0x API doesn't support receiving ETH as a buyToken. (Only WETH). In the future with changes to the forwarder contract, buying ETH may be supported.

### DIFF
--- a/src/handlers/swap_handlers.ts
+++ b/src/handlers/swap_handlers.ts
@@ -13,7 +13,12 @@ import { SwapService } from '../services/swap_service';
 import { TokenMetadatasForChains } from '../token_metadatas_for_networks';
 import { CalculateSwapQuoteParams, ChainId, GetSwapQuoteRequestParams, GetSwapQuoteResponse } from '../types';
 import { schemaUtils } from '../utils/schema_utils';
-import { findTokenAddress, getTokenMetadataIfExists, isETHSymbol, isWETHSymbol } from '../utils/token_metadata_utils';
+import {
+    findTokenAddress,
+    getTokenMetadataIfExists,
+    isETHSymbol,
+    isWETHSymbolOrAddress,
+} from '../utils/token_metadata_utils';
 
 export class SwapHandlers {
     private readonly _swapService: SwapService;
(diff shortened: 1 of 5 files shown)


fragosti:
Can we link to the documentation change in this PR before merging? I think that will be a good practice moving forward. Looks good!

dave4506:
Docs can be found here: https://github.com/0xProject/website/pull/271

src/handlers/swap_handlers.ts
```diff
@@ -37,10 +37,24 @@ export class SwapHandlers {
...
         } = parseGetSwapQuoteRequestParams(req);
+
         const isETHSell = isETHSymbol(sellToken);
+        // if sellToken is not WETH and buyToken is ETH, throw
+        if (!isWETHSymbol(sellToken) && isETHSymbol(buyToken)) {
+            throw new ValidationError([
+                {
+                    field: 'buyToken',
+                    code: ValidationErrorCodes.TokenNotSupported,
+                    reason: "Buying ETH is unsupported (set to 'WETH' to received wrapped Ether)",
```
dekz, line 48:
```suggestion
                    reason: "Buying ETH is unsupported (set to 'WETH' to receive wrapped Ether)",
```

src/handlers/swap_handlers.ts
```diff
@@ -37,10 +37,24 @@ export class SwapHandlers {
             excludedSources,
             affiliateAddress,
         } = parseGetSwapQuoteRequestParams(req);
+
         const isETHSell = isETHSymbol(sellToken);
+        // if sellToken is not WETH and buyToken is ETH, throw
+        if (!isWETHSymbol(sellToken) && isETHSymbol(buyToken)) {
```
dekz, line 43: I think `sellToken`  can be an address here, not always a symbol. We should handle `isWETHSymbol(WETH_ADDRESS)` as `sellToken=0xc02&buyToken=ETH` should work as expected

```suggestion
        if (!isWETHSymbolOrAddress(sellToken) && isETHSymbol(buyToken)) {
```

src/services/swap_service.ts
```diff
@@ -233,6 +255,57 @@ export class SwapService {
...
+        if (amount === undefined) {
+            throw new Error('sellAmount or buyAmount required');
+        }
+        const data = (isUnwrap
+            ? this._wethContract.withdraw(amount)
+            : this._wethContract.deposit()
+        ).getABIEncodedTransactionData();
+        const value = isUnwrap ? ZERO : amount;
+        const affiliatedData = this._attributeCallData(data, affiliateAddress);
+        const gasPrice = providedGasPrice || (await this._protocolFeeUtils.getGasPriceEstimationOrThrowAsync());
```
dekz, line 281: This can be a TODO but I don't think its worth an aggressive 8 gwei gas price for wrap/unwrap.

src/services/swap_service.ts
```diff
@@ -233,6 +255,57 @@ export class SwapService {
...
+        const gasEstimate = await this._estimateGasOrThrowRevertErrorAsync({
```
dekz, line 282: We could probably safely fix this to a constant and save a RPC request.

src/services/swap_service.ts
```diff
@@ -233,6 +255,57 @@ export class SwapService {
...
+            to: this._wethContract.address,
+            data: affiliatedData,
+            from,
+            value,
+            gasPrice,
+        });
+        const suggestedGasEstimate = gasEstimate.times(GAS_LIMIT_BUFFER_PERCENTAGE + 1).integerValue();
```
dekz, line 289: I think we could remove the Buffer as it's pretty deterministic. 

We add a buffer because fallback/Kyber/Curve costs can sometimes not be guaranteed

PR #139, opened 2020-03-20
Added delegator/events/<address> endpoint

alexkroeger:
Added endpoint for delegator events.

Events include:
- staked
- removed stake
- moved stake
- deposited ZRX
- withdrew ZRX
- earned rewards

Testing:
- [x] Tested locally

### DIFF
--- a/src/utils/staking_utils.ts
+++ b/src/utils/staking_utils.ts
@@ -4,6 +4,7 @@ import {
     AllTimeDelegatorPoolStats,
     AllTimePoolStats,
     AllTimeStakingStats,
+    DelegatorEvent,
     Epoch,
     EpochPoolStats,
     Pool,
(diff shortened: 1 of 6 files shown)


src/utils/staking_utils.ts
```diff
@@ -191,6 +193,18 @@ export const stakingUtils = {
 
         return poolData;
     },
+    getDelegatorEventsFromRaw: (rawDelegatorEvents: RawDelegatorEvent[]): DelegatorEvent[] => {
+        const delegatorEvents: DelegatorEvent[] = rawDelegatorEvents.map(rawEvent => ({
+            eventType: rawEvent.event_type,
+            address: rawEvent.address,
+            blockNumber: Number(rawEvent.block_number),
```
fragosti: If this is undefined it could be a NaN btw, but if it's always 0 or null then we're good.

alexkroeger: I think it can be relied on to be always null instead of undefined given the structure of the query. However, I changed it to make it be able to take on a null value as opposed to Number(null) = 0

alexkroeger:
deploy staging

alexkroeger:
deploy prod

PR #141, opened 2020-03-24
[Fix] event => event_type

alexkroeger:
Changed the field name in the events query to conform with the expected raw output.


alexkroeger:
deploy staging

alexkroeger:
Tested locally

PR #145, opened 2020-03-25
Hotfix: Filter expired orders

xianny:
SRA is returning orders that are expired on average 29 seconds.

Quick fix to filter orders right before returning them.

### DIFF
--- a/src/services/orderbook_service.ts
+++ b/src/services/orderbook_service.ts
@@ -2,7 +2,7 @@ import { APIOrder, OrderbookResponse, PaginatedCollection } from '@0x/connect';
 import { assetDataUtils } from '@0x/order-utils';
 import { AssetPairsItem, OrdersRequestOpts, SignedOrder } from '@0x/types';
 import * as _ from 'lodash';
-import { Connection, In } from 'typeorm';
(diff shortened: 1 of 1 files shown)


xianny:
deploy staging

xianny:
Tests are green on staging.

src/services/orderbook_service.ts
```diff
@@ -69,10 +69,18 @@ export class OrderBookService {
     ): Promise<OrderbookResponse> {
         const [bidSignedOrderEntities, askSignedOrderEntities] = await Promise.all([
             this._connection.manager.find(SignedOrderEntity, {
-                where: { takerAssetData: baseAssetData, makerAssetData: quoteAssetData },
+                where: {
+                    takerAssetData: baseAssetData,
+                    makerAssetData: quoteAssetData,
+                    expirationTimeSeconds: MoreThan(new Date().getTime()),
```
dekz: I believe this `new Date().getTime()` is milliseconds whereas `expirationTimeSeconds` is in seconds.
```suggestion
                    expirationTimeSeconds: MoreThan(Math.floor(new Date().getTime() / 1000)),
```

PR #147, opened 2020-03-25
Feat/filter expired orders

fragosti:
I'm still looking into this but yea lets patch.

### DIFF
--- a/src/services/orderbook_service.ts
+++ b/src/services/orderbook_service.ts
@@ -1,11 +1,12 @@
 import { APIOrder, OrderbookResponse, PaginatedCollection } from '@0x/connect';
-import { assetDataUtils } from '@0x/order-utils';
 import { AssetPairsItem, OrdersRequestOpts, SignedOrder } from '@0x/types';
 import * as _ from 'lodash';
 import { Connection, In } from 'typeorm';
 
+import { ONE_SECOND_MS } from '../constants';
 import { SignedOrderEntity } from '../entities';
 import { ValidationError } from '../errors';
+import { logger } from '../logger';
 import { MeshClient } from '../utils/mesh_client';
 import { meshUtils } from '../utils/mesh_utils';
 import { orderUtils } from '../utils/order_utils';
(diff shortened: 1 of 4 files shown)


fragosti:
deploy prod

fragosti:
deploy production

src/services/orderbook_service.ts
```diff
@@ -111,49 +111,15 @@ export class OrderBookService {
...
-                apiOrder =>
-                    ordersFilterParams.takerAssetProxyId === undefined ||
-                    assetDataUtils.decodeAssetDataOrThrow(apiOrder.order.takerAssetData).assetProxyId ===
-                        ordersFilterParams.takerAssetProxyId,
-            );
-        const paginatedApiOrders = paginationUtils.paginate(apiOrders, page, perPage);
+        const filteredApiOrders = orderUtils.filterOrders(apiOrders, ordersFilterParams);
+        // Remove expired orders
+        const dateNowSeconds = Date.now() / ONE_SECOND_MS;
+        const freshFilteredApiOrders = filteredApiOrders.filter(apiOrder =>
```
PirosB3: Why doesn't `Remove expired orders` go into `orderUtils.filterOrders`? is that because it's a temporary fix?

src/services/orderbook_service.ts
```diff
@@ -111,49 +111,15 @@ export class OrderBookService {
...
+            apiOrder.order.expirationTimeSeconds.gt(dateNowSeconds),
```
PirosB3: could we by any chance return orders that won't expire in the next N seconds?

fragosti: What is N?

src/services/orderbook_service.ts
fragosti: `orderUtils.filterOrders` works on the filters specified in the SRA, which `expirationTimeSeconds` is not part of.

src/services/orderbook_service.ts
PirosB3: sorry! that's my way of saying "some number that is to be determined". So, maybe `N` here could be 60?

fragosti: haha I know. I'm saying I can't move forward without a value for N.

PirosB3: I would recommend at least 60, but I can defer to @steveklebanoff or @blakehendo

fragosti: Going to make it an ENV var...

PirosB3: can we set this to 10 seconds for now? me and @steveklebanoff discussed this

blakehendo: 10 seconds works great.

PR #151, opened 2020-03-28
Await results from Mesh validation

dekz:
Await using a `for of` instead of `forEach` which doesn't await.


dekz:
deploy production

PR #155, opened 2020-03-31
Normalize addresses to lowercase

xianny:
@PirosB3 pointed out that retrieving orders from `GET /sra/v3/orders` doesn't work if the maker address is not all lowercase.

On further investigation I found that retrieving orders required exact string matches between query params and postgres records.

This PR normalizes addresses from incoming requests to all lowercase, so that API requests should be case-insensitive to addresses. Normalization is implemented across all endpoints for consistency.


xianny:
Moved address normalization into a middleware function.

src/services/orderbook_service.ts
```diff
@@ -71,10 +71,10 @@ export class OrderBookService {
     ): Promise<OrderbookResponse> {
         const [bidSignedOrderEntities, askSignedOrderEntities] = await Promise.all([
             this._connection.manager.find(SignedOrderEntity, {
-                where: { takerAssetData: baseAssetData, makerAssetData: quoteAssetData },
+                where: { takerAssetData: baseAssetData.toLowerCase(), makerAssetData: quoteAssetData.toLowerCase() },
```
fragosti, line 74: Do we still need these calls?

test/app_test.ts
```diff
@@ -54,4 +57,31 @@ describe('app test', () => {
...
+            .expect(HttpStatus.OK)
+            .then(response => {
+                expect(response.body.perPage).to.equal(DEFAULT_PER_PAGE);
+                expect(response.body.page).to.equal(DEFAULT_PAGE);
+                expect(response.body.total).to.equal(1);
+                expect(response.body.records[0].order).to.deep.equal(apiOrderResponse);
+            });
+
+        await dependencies.connection.manager.remove(orderModel);
+    });
```
fragosti, line 86: Noice

src/runners/http_service_runner.ts
```diff
@@ -60,6 +61,9 @@ export async function runHttpServiceAsync(
     server.keepAliveTimeout = config.HTTP_KEEP_ALIVE_TIMEOUT;
     server.headersTimeout = config.HTTP_HEADERS_TIMEOUT;
 
+    // transform all values of `req.query.[xx]Address` to lowercase
+    app.use(addressNormalizer);
```
fragosti, line 65: We need to add this to the monolith runner too right?

xianny:
deploy staging

xianny:
deploy staging

PR #156, opened 2020-04-01
Add zeros for sources

dave4506:
If a liquidity source is not utilized in swap quote, `sources` will return the liquidity provider with `proportion` set to zero. This will allow for easier understanding of the potential breakdown of liquidity when requesting a quote. 

Docs update: https://github.com/0xProject/website/pull/274

### DIFF
--- a/src/services/swap_service.ts
+++ b/src/services/swap_service.ts
@@ -238,7 +238,10 @@ export class SwapService {
     private _convertSourceBreakdownToArray(
         sourceBreakdown: SwapQuoteOrdersBreakdown,
     ): GetSwapQuoteResponseLiquiditySource[] {
-        const defaultSourceBreakdown: SwapQuoteOrdersBreakdown = Object.assign({}, ...Object.values(ERC20BridgeSource).map(s => ({ [s]: ZERO })));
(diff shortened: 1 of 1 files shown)


src/services/swap_service.ts
```diff
@@ -237,8 +238,18 @@ export class SwapService {
...
     ): GetSwapQuoteResponseLiquiditySource[] {
+        const defaultSourceBreakdown: SwapQuoteOrdersBreakdown = Object.keys(ERC20BridgeSource).reduce(
+            (acc: SwapQuoteOrdersBreakdown, k: any): SwapQuoteOrdersBreakdown => {
+                return {
+                    ...acc,
+                    [k]: ZERO,
+                };
+            },
+            {},
+        );
```
dorothy-zbornak, line 249: Optional nit:

So I feel like `reduce()` is an anti-pattern because it's never obvious what it's doing, and is usually pretty long-winded. Might I suggest...?
```ts
const defaultSourceBreakdown = Object.assign(...Object.values(ERC20BridgeSource).map(s => ({ [s]: ZERO )}));
```

PR #158, opened 2020-04-01
Add configurable keepAliveTimeout and headersTimeout to http.Server instances

opaolini:
About
===

This PR adds higher default `keepAliveTimeout` and `headersTimeout` settings for `server` from the `http` package that `express` uses.

It also exposes them to be configurable by passing in env vars:
- `HTTP_KEEP_ALIVE_TIMEOUT`
- `HTTP_HEADERS_TIMEOUT`


opaolini:
deploy staging

opaolini:
Added to `.env_example`! Should we pass in them to `.circleci/config.yml` if they are "optional" (the default values are good for most deployments)

opaolini:
deploy production

opaolini:
deploy production

PR #159, opened 2020-04-06
Hotfix: filter expired orders from orderbook

xianny:
Follow-up to https://github.com/0xProject/0x-api/pull/147

@steveklebanoff pointed out we are still getting expired orders from `/sra/orderbook`. I looked at the code and figured out we were only filtering expired orders from `/sra/orders`. This PR adds filtering for all responses from `OrderBookService`. We should continue investigating why the `OrderWatcherService` isn't updating the orderbook properly, but this patch will get rid of expired orders in API responses for now.


xianny:
deploy staging

PR #160, opened 2020-04-06
HTTP sync chunks to prevent content length too large

dekz:
Mesh HTTP sync now uses bytes length chunking (2.5mb) to keep it under the HTTP content length maximum (~5.24mb). We're well under and could tune this up if needed, though I'm leaving a bit of room purposefully.

With this change we sync ~5719 orders in 4 chunks, each chunk is validated in around 0.8-1.2s. We are **not** currently posting in parallel. 

In the event we hit a 500 or a 429, we retry 3 times, with a 1 second delay. Though this retry-axios library should be fully tested in all scenarios. If we lose connection (network) then the entire Mesh service reconnects and re-syncs. So most network level errors are reacted to at a higher level. I.e if EPIPE occurs it will result in a full new order sync, so isn't necessarily required to retry.

```
{"level":"info","pid":28,"hostname":"api-worker-staging-mainnet-74877645d6-5bl64","msg":"OrderWatcherService starting up!","v":1}
{"level":"info","pid":28,"hostname":"api-worker-staging-mainnet-74877645d6-5bl64","msg":"OrderWatcherService syncing orderbook with Mesh","v":1}
{"level":"info","pid":28,"hostname":"api-worker-staging-mainnet-74877645d6-5bl64","msg":"Mesh HTTP sync 1/4 complete 1276ms","v":1}
{"level":"info","pid":28,"hostname":"api-worker-staging-mainnet-74877645d6-5bl64","msg":"Mesh HTTP sync 2/4 complete 813ms","v":1}
{"level":"info","pid":28,"hostname":"api-worker-staging-mainnet-74877645d6-5bl64","msg":"Mesh HTTP sync 3/4 complete 804ms","v":1}
{"level":"info","pid":28,"hostname":"api-worker-staging-mainnet-74877645d6-5bl64","msg":"Mesh HTTP sync 4/4 complete 211ms","v":1}
{"level":"info","pid":28,"hostname":"api-worker-staging-mainnet-74877645d6-5bl64","msg":"OrderWatcherService sync {\"accepted\":5638,\"rejected\":81,\"sent\":5719}","v":1}
```

In the event of consecutive 400/500's we will not know the status of these orders and must reject them. It is unlikely to result in lost orders as we have previously fetched the orders from mesh and are first rejecting and stale orders then adding all orders from mesh.

### Future work
- [ ] Find the Websocket fragmentation limit and potentially use byte length chunking there as well.


src/utils/mesh_client.ts
```diff
@@ -1,45 +1,74 @@
...
 import * as _ from 'lodash';
+import * as rax from 'retry-axios';
 
 import { MESH_ORDERS_BATCH_SIZE } from '../constants';
+import { logger } from '../logger';
 
 import { utils } from './utils';
 
+// Attach retry-axios to the global instance
+rax.attach();
```
fragosti, line 12: Can we make a `utils/axios` file and import axios from there and include this statement in there as well?

fragosti: Like we do for logger.

dekz:
deploy staging

dekz:
deploy staging

dekz:
deploy production

dekz:
deploy staging

dekz:
deploy production

dekz:
deploy staging

dekz:
deploy production

PR #162, opened 2020-04-07
RFQ-T Firm Quotes

feuGeneA:
These changes depend on the changes in https://github.com/0xProject/0x-monorepo/pull/2541 .

Every commit is crafted to be completely independent and atomic.  I suggest you review them one at a time, in order to give better context, rather than just reviewing all of the changes at once.

The following commits should be removed before merging:
* `TEMPORARY: Update gitpkg monorepo references`, https://github.com/0xProject/0x-api/pull/162/commits/26d8527a1e908ffa5a1e2d246d73540134e187ca
    * these will likely change once the monorepo PR is merged
* `TEMPORARY: use gene's ganache snapshot for tests`, https://github.com/0xProject/0x-api/pull/162/commits/0b09babc31a3de0c2461f462394f6c1d2b8a265d
    * this will go away once the monorepo PR is merged and the ganache snapshot is deployed from it.

Checklist:

-   [x] MORE TESTS
-   [x] Create new "should validate?" boolean for firm quotes which is true or false, defaulting to true.  Tell integrators to explicitly send in false if their taker address is a smart contract.


package.json
```diff
@@ -8,8 +8,8 @@
     "scripts": {
         "clean": "shx rm -rf lib 0x_mesh/db",
         "build": "tsc -p tsconfig.json",
-        "test": "ETHEREUM_RPC_URL=http://localhost:8545 CHAIN_ID=1337 mocha --require source-map-support/register --require make-promises-safe lib/test/**/*_test.js --timeout 100000 --exit",
-        "dev": "nodemon -r dotenv/config src/index.ts | pino-pretty",
+        "test": "ETHEREUM_RPC_URL=http://localhost:8545 CHAIN_ID=1337 RFQT_API_KEY_WHITELIST='[\"koolApiKey1\"]' RFQT_MAKER_ENDPOINTS='[\"http://localhost:4000\"]' mocha --require source-map-support/register --require make-promises-safe lib/test/**/*_test.js --timeout 100000 --exit",
+        "dev": "ETHEREUM_RPC_URL=http://localhost:8545 CHAIN_ID=1337 RFQT_API_KEY_WHITELIST='[\"koolApiKey1\"]' RFQT_MAKER_ENDPOINTS='[\"http://localhost:4000\"]' nodemon -r dotenv/config src/index.ts | pino-pretty",
```
feuGeneA, line 15: Is this used anywhere outside of a ganache dev environment?  Are breaking other use cases by introducing the ganache RPC URL here?

src/handlers/swap_handlers.ts
```diff
@@ -63,6 +64,8 @@ export class SwapHandlers {
                 gasPrice,
                 excludedSources,
                 affiliateAddress,
+                apiKey: takerAddress && takerAddress !== NULL_ADDRESS ? req.header('0x-api-key') : undefined,
```
feuGeneA, line 67: Cut the business logic and just pass through the `apiKey`.  Enforce that requirement elsewhere (if it's not already addressed elsewhere).

src/handlers/swap_handlers.ts
```diff
@@ -166,6 +169,8 @@ const parseGetSwapQuoteRequestParams = (req: express.Request): GetSwapQuoteReque
             ? undefined
             : parseStringArrForERC20BridgeSources(req.query.excludedSources.split(','));
     const affiliateAddress = req.query.affiliateAddress;
+    // tslint:disable-next-line:boolean-naming
+    const intentOnFilling = req.query.intentOnFilling === undefined ? false : true;
```
feuGeneA, line 173: `'false'` would come out as "true" here.  Enforce a specific value, eg "true".

src/config.ts
```diff
@@ -100,6 +100,10 @@ export const LIQUIDITY_POOL_REGISTRY_ADDRESS: string | undefined = _.isEmpty(
           EnvVarType.ETHAddressHex,
       );
 
+export const RFQT_API_KEY_WHITELIST: string[] = JSON.parse(process.env.RFQT_API_KEY_WHITELIST || '[]');
```
feuGeneA, line 102: Consider allowing simple csv in the env var instead of requiring full fledged JSON.

src/types.ts
```diff
@@ -342,6 +342,8 @@ export interface GetSwapQuoteRequestParams {
     gasPrice?: BigNumber;
     excludedSources?: ERC20BridgeSource[];
     affiliateAddress?: string;
+    apiKey?: string;
```
feuGeneA, line 345: Do we need this added to _both_ `GetSwapQuoteRequestParams` _and_ `CalculateSwapQuoteParams`?

package.json
```diff
@@ -6,6 +6,7 @@
     "author": "Francesco Agosti <francesco@0x.org>",
     "license": "Apache-2.0",
     "scripts": {
+        "install": "sed -ie 's!import { Provider as Web3Provider } from \"web3/providers\";!import { Web3EthereumProvider as Web3Provider } from \"web3-providers\";!' node_modules/ganache-core/typings/index.d.ts # see https://github.com/trufflesuite/ganache-core/issues/465#issuecomment-610005598",
```
feuGeneA, line 9: this can be deleted now, right? since we're not using migrations any more (now that we ripped out the contract deploy from the tests), so we don't really need this... right?

package.json
```diff
@@ -18,8 +18,8 @@
         "dev:service:swap_http": "nodemon -r dotenv/config src/runners/http_swap_service_runner.ts | pino-pretty",
         "dev:service:order_watcher": "nodemon -r dotenv/config src/runners/order_watcher_service_runner.ts | pino-pretty",
         "watch": "tsc -w",
-        "prettier": "prettier --write '**/*.{ts,tsx,json,md}' --config .prettierrc",
-        "prettier:ci": "prettier --list-different '**/*.{ts,tsx,json,md}' --config .prettierrc",
+        "prettier": "prettier --write 'src/**/*.{ts,tsx,json,md}' --config .prettierrc",
```
feuGeneA, line 22: consider .prettierignore

package.json
feuGeneA: No, this can't be deleted, because `ganache-core` is pulled in via `@0x/subproviders`.

feuGeneA:
@steveklebanoff , @PirosB3 and I had a call where we went over this code together.  Aside from the inline comments I've already made, we also generated this feedback:

* Should we have whitelists in BOTH the 0x API *AND* in RFQ-T providers themselves?  

* We should emit some log entries (a) just before hitting RFQ-T quote providers, and (b) just after hitting them, so we can tell how long that process is taking in the wild.

* How is axios pooling HTTP requests? Is it re-using sockets or reinstantiating them? Utilizing an HTTP keep-alive would be ideal.

* With the code as written now, an RFQ-T taker MUST hold some amount of the taker asset.

    * This is because `SwapService.calculateSwapQuoteAsync()` says `if (from) { ... this._estimateGasOrThrowRevertErrorAsync(...)`.  Since an RFQ-T order always includes a taker address, it's falling into this bucket, and the fill simulation fails if the taker doesn't have any taker assets.

    * The only immediately available hacky fix would be to change it to say `if (from && apiKey) { ... }`.  Ideally it would be `if (from && isRFQT) { ... }`, but there's no notion of "isRFQT" in the code as written; `apiKey` is all we've got in this context.  That feels less than straightforward, but we've already decided to use the presence of both an API key and a taker address as a proxy for "isRFQT" (see https://0xproject.quip.com/qn1xA43e3ANE/0x-API-RFQ-T-Tech-Spec#dVHACAJr4cu), so maybe it's not so "hacky" after all.

    * We need to have TWO test cases, one that has taker holding assets, one that doesn't.

src/handlers/swap_handlers.ts
feuGeneA: This enforcement is already in place in https://github.com/0xProject/0x-monorepo/pull/2541 in `SwapQuoter._getSwapQuoteAsync()`.

src/config.ts
dekz: CSV is what we typically use.

src/services/swap_service.ts
```diff
@@ -103,9 +110,9 @@ export class SwapService {
         const {
             makerAssetAmount: guaranteedMakerAssetAmount,
             totalTakerAssetAmount: guaranteedTotalTakerAssetAmount,
-            gas,
         } = attributedSwapQuote.worstCaseQuoteInfo;
         const { orders, gasPrice, sourceBreakdown } = attributedSwapQuote;
+        const gas = 10;
```
dekz, line 115:
```suggestion
```

src/services/swap_service.ts
```diff
@@ -69,13 +71,18 @@ export class SwapService {
...
         } = params;
         const assetSwapperOpts = {
             ...ASSET_SWAPPER_MARKET_ORDERS_OPTS,
             slippagePercentage,
             bridgeSlippage: slippagePercentage,
             gasPrice: providedGasPrice,
             excludedSources, // TODO(dave4506): overrides the excluded sources selected by chainId
+            apiKey,
+            intentOnFilling,
+            takerAddress: from,
```
dekz, line 85: Consider not conflating `from` and `takerAddress`. As you discovered, this is sometimes not often the case. I.e Via forwarder `takerAddress` is the Forwarder, from is the end user.

What happens when the Forwarder is used? the `takerAddress` in the order MUST be the Forwarder contract.

src/types.ts
dekz: Consider pushing these options under an `rfqt:` namespace.

src/config.ts
```diff
@@ -116,33 +119,11 @@ const EXCLUDED_SOURCES = (() => {
...
-    {},
-    ...(Object.keys(gasSchedule) as ERC20BridgeSource[]).map(k => ({
-        [k]: new BigNumber(gasSchedule[k] + 1.5e5),
-    })),
-);
-
 export const ASSET_SWAPPER_MARKET_ORDERS_OPTS: Partial<SwapQuoteRequestOpts> = {
     excludedSources: EXCLUDED_SOURCES,
     bridgeSlippage: DEFAULT_QUOTE_SLIPPAGE_PERCENTAGE,
-    maxFallbackSlippage: DEFAULT_FALLBACK_SLIPPAGE_PERCENTAGE,
```
dekz, line 141: Is this related to changes in Asset-swapper?

feuGeneA: yup, that's in https://github.com/0xProject/0x-api/pull/162/commits/5f8381d43f7fa58a8decf803c017b91da13615e5

package.json
feuGeneA: Addressed in https://github.com/0xProject/0x-api/pull/162/commits/d11b8508ead8586d348f94d154460e13d453de1e

src/handlers/swap_handlers.ts
feuGeneA: Addressed in https://github.com/0xProject/0x-api/pull/162/commits/b1263163c60edcbc81cf5c5ac26f6c8b439e09bf

src/handlers/swap_handlers.ts
feuGeneA: Addressed in https://github.com/0xProject/0x-api/pull/162/commits/cbecfc4fd6b0d8df190e8f86cf235fc08aad714c

src/types.ts
feuGeneA: Addressed in https://github.com/0xProject/0x-api/pull/162/commits/902b15d5dc6ffc311162e25c210299ed5b38535b

src/config.ts
feuGeneA: Addressed in https://github.com/0xProject/0x-api/pull/162/commits/47227ad4e38d79b72d4ebe0629638433e6606a51

src/services/swap_service.ts
```diff
@@ -52,9 +52,10 @@ export class SwapService {
         const swapQuoterOpts: Partial<SwapQuoterOpts> = {
             chainId: CHAIN_ID,
             expiryBufferMs: QUOTE_ORDER_EXPIRATION_BUFFER_MS,
-            liquidityProviderRegistryAddress: LIQUIDITY_POOL_REGISTRY_ADDRESS,
```
feuGeneA, line 53: This should not have been removed.

package.json
feuGeneA: This is the output I get without limiting it to `src`:

```
$ yarn prettier:ci
yarn run v1.19.1
$ prettier --list-different './**/*.{ts,tsx,json,md}' --config .prettierrc
[error] Unable to expand glob patterns: ./**/*.{ts,tsx,json,md} !**/node_modules/** !./node_modules/** !**/.{git,svn,hg}/** !./.{git,svn,hg}/**
[error] EACCES: permission denied, scandir '/home/gene/dev/0x-api/postgres'
error Command failed with exit code 2.
info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.
```

That `postgres` folder was created by a docker container, and has root-only permissions:
```
$ ls -ld postgres
drwx------ 19 999 root 4096 Apr  8 16:04 postgres
```

I tried using a .pretterignore file, with values of `postgres`, `/postgres`, and `/postgres/`, and none changed this behavior.

feuGeneA:
deploy staging

feuGeneA:
deploy kovan-staging

feuGeneA:
deploy kovan-staging

src/services/swap_service.ts
feuGeneA: I tried to account for this here: https://github.com/0xProject/0x-api/pull/162/commits/0255b25dcc8a9cefccfccdbcb50898dcd9ee9ef2. Does that address the issue?

src/services/swap_service.ts
feuGeneA: Addressed in https://github.com/0xProject/0x-api/pull/162/commits/29d62150c006472a07c9a3e0f7bb65548b4a42a8

.circleci/config.yml
```diff
@@ -9,6 +9,9 @@ jobs:
             - image: circleci/node:8.12.0-browsers
             - image: circleci/postgres:9.6.5-alpine-ram
             - image: 0xorg/ganache-cli:6.0.0
+              environment:
+                SNAPSHOT_NAME: 0x_ganache_snapshot_gene
```
steveklebanoff, line 13: Do we still need this?

src/services/swap_service.ts
```diff
@@ -10,6 +10,7 @@ import {
     SwapQuoter,
```
steveklebanoff, line 10: @feuGeneA Don't we still need some sort of resolve on the issue of `_estimateGasOrThrowRevertErrorAsync` for when the taker address is the forwarder address?

src/services/swap_service.ts
```diff
@@ -10,6 +10,7 @@ import {
...
     SwapQuoterOpts,
```
steveklebanoff, line 11: I guess the question still remains: can we skip the `_estimateGasOrThrowRevertErrorAsync` if we are doing a RFQT firm quote?  

I imagine we'd like to still return a firm quote to a taker even if they do not have allowances or a sufficient asset balance.  What do you think?

package.json
steveklebanoff: Hrm, I don't like that we are not linting `package.json` anymore.  It does seem like this is a linux-specific issue where the `scandir` still occurs regardless of `.prettierignore`

Could you possibly just `chown` the `postgres` dir after it's created?  Or is it recreated every time you `docker-compose up`?

src/services/swap_service.ts
steveklebanoff: AFAIK RFQT for a dex-aggregator contract will still be broken until we resolve this.

Consider a dex.ag smart contract like 1inch.  The takerAddress would be the 1inch contract.  The `_estimateGasOrThrowRevertErrorAsync` call would fail because the `from` address would be the 1inch contract which does not hold a balance of the taker asset.

Suggestion: for now, disable eth gas estimation for RFQT requests

feuGeneA: As discussed with Steve and Jacob elsewhere, we decided to introduce a new query parameter, `skipValidation`.  An integrator using a smart contract as a taker can use this parameter to suppress validation of the transaction.  All other integrators will get the the default, which is to validate.  Implemented in https://github.com/0xProject/0x-api/pull/162/commits/ea8c8e94752e2b07c7d2f60b21ac2acc8e4f01de

.circleci/config.yml
steveklebanoff, line 13: When will we be able to remove this?

feuGeneA: When the monorepo PR merges, I will publish a new snapshot version to the regular name, and then I will remove this.

package.json
feuGeneA: It's recreated every time I `docker-compose up`.

Does this address your concerns? https://github.com/0xProject/0x-api/pull/162/commits/f325f428a8dd863527d0870ee0e1a08438e2a0c6

src/config.ts
```diff
@@ -100,6 +100,10 @@ export const LIQUIDITY_POOL_REGISTRY_ADDRESS: string | undefined = _.isEmpty(
           EnvVarType.ETHAddressHex,
       );
 
+export const RFQT_API_KEY_WHITELIST: string[] = (process.env.RFQT_API_KEY_WHITELIST || '[]').split(',');
```
steveklebanoff, line 103: If `process.env.RFQT_API_KEY_WHITELIST` is `undefined` then `RFQT_API_KEY_WHITELIST` will be `[ '[]' ]` when it should really just be an empty array.  Seems like a relic from the old JSON approach

```
> const RFQT_API_KEY_WHITELIST: string[] = (process.env.RFQT_API_KEY_WHITELIST || '[]').split(',');
undefined
> RFQT_API_KEY_WHITELIST
[ '[]' ]
```

src/config.ts
```diff
@@ -100,6 +100,10 @@ export const LIQUIDITY_POOL_REGISTRY_ADDRESS: string | undefined = _.isEmpty(
...
+
+export const RFQT_MAKER_ENDPOINTS: string[] = (process.env.RFQT_MAKER_ENDPOINTS || '[]').split(',');
```
steveklebanoff, line 105: Same comment as above

src/handlers/swap_handlers.ts
```diff
@@ -166,6 +177,15 @@ const parseGetSwapQuoteRequestParams = (req: express.Request): GetSwapQuoteReque
             ? undefined
             : parseStringArrForERC20BridgeSources(req.query.excludedSources.split(','));
     const affiliateAddress = req.query.affiliateAddress;
+    const rfqt =
+        req.query.intentOnFilling === undefined
+            ? undefined
+            : {
+                  intentOnFilling: ['true', ''].includes(req.query.intentOnFilling) ? true : false,
```
steveklebanoff, line 184: Can we just do something like the following?

```
const rfqt = req.query.intentOnFilling === 'true'  ? { intentOnFilling: true } : undefined;
```

src/handlers/swap_handlers.ts
```diff
@@ -166,6 +177,15 @@ const parseGetSwapQuoteRequestParams = (req: express.Request): GetSwapQuoteReque
...
+              };
+    // tslint:disable-next-line:boolean-naming
+    const skipValidation =
+        req.query.skipValidation === undefined ? false : ['true', ''].includes(req.query.skipValidation) ? true : false;
```
steveklebanoff, line 188: Why do we want to consider `req.query.skipValidation === ''` as true?

src/schemas/swap_quote_request_schema.json
```diff
@@ -21,6 +21,17 @@
         },
         "excludedSources": {
             "type": "string"
+        },
+        "apiKey": {
+            "type": "string"
+        },
+        "intentOnFilling": {
+            "type": "string",
```
steveklebanoff, line 29: Can we use a `boolean` type here? http://json-schema.org/understanding-json-schema/reference/boolean.html

package.json
steveklebanoff: As long as you tested this as it works as expected, this commit LGTM 👍

src/config.ts
feuGeneA: Thank you for catching this.  Addressed in https://github.com/0xProject/0x-api/pull/162/commits/e1693bba254b24852fbedae88b25bf8ebedc5616

src/config.ts
feuGeneA: Addressed in https://github.com/0xProject/0x-api/pull/162/commits/e1693bba254b24852fbedae88b25bf8ebedc5616

src/handlers/swap_handlers.ts
feuGeneA: I did some digging when I updated the schemas here.  I had initially wanted to make them `type: "boolean"` (whatever exactly that would mean!) but the `jsonschema` package didn't handle that nicely.  Later discovered this question which is a good analog to what we're doing here: https://github.com/OAI/OpenAPI-Specification/issues/1782

I changed this here because I liked the idea of the mere presence of this flag indicating it should be true, eg `/quote?intentOnFilling&...`.

src/handlers/swap_handlers.ts
feuGeneA: Same logic behind my recent change to `intentOnFilling`.  I think it makes logical sense to be able to say just `/quote?skipValidation&...`, and when that's the query string, `req.query` here contains `"skipValidation": ""`.

src/schemas/swap_quote_request_schema.json
feuGeneA: I wanted to.  I tried.  `jsonschema` choked on it.

feuGeneA: At test runtime i get this json schema error

`{\"field\":\"instance.intentOnFilling\",\"code\":1001,\"reason\":\"is not of a type(s) boolean\"}`

This happened no matter whether my client said `quote?intentOnFilling` or `quote?intenOnFilling=true`.

steveklebanoff: Got it.

I think this schema could be cleaner if we:
-  get rid of the `flag` functionality (i.e. setting `?intentOnFilling` indicating that `intentOnFilling` is equal to `true`).  I have not seen this used in query parameters before, and IMHO is not super important to support
- Use an `enum` type to force the consumer of the API to specify the text `true`, `false` for our 'boolean' params (intentOnFilling, skipValidaiton).  (See https://stackoverflow.com/a/16826238 and https://json-schema.org/understanding-json-schema/reference/generic.html#enumerated-values
   - Make sure to keep specifying these params optional though

This would allow `?skipValidation=truue` to raise a schema error instead of silently being set to `false` on our backend.

src/handlers/swap_handlers.ts
steveklebanoff: How we resolve https://github.com/0xProject/0x-api/pull/162#discussion_r409163018 will impact this

src/services/swap_service.ts
```diff
@@ -69,13 +77,25 @@ export class SwapService {
...
             bridgeSlippage: slippagePercentage,
             gasPrice: providedGasPrice,
             excludedSources, // TODO(dave4506): overrides the excluded sources selected by chainId
+            apiKey,
+            rfqt:
+                rfqt === undefined || from === undefined || from === NULL_ADDRESS
+                    ? undefined
+                    : {
+                          ...rfqt,
+                          takerAddress: isETHSell ? getContractAddressesForChainOrThrow(CHAIN_ID).forwarder : from,
```
steveklebanoff, line 97: Would be good to add a comment describing what's happening here

src/services/swap_service.ts
```diff
@@ -120,7 +140,7 @@ export class SwapService {
         const affiliatedData = this._attributeCallData(data, affiliateAddress);
 
         let suggestedGasEstimate = new BigNumber(gas);
-        if (from) {
+        if (!skipValidation && from) {
```
steveklebanoff, line 143: 👍

steveklebanoff:
LGTM pending:
* Resolve of schema discussion https://github.com/0xProject/0x-api/pull/162#discussion_r409163018
* Merge of additional testing https://github.com/0xProject/0x-api/pull/170

src/services/swap_service.ts
feuGeneA: Addressed in https://github.com/0xProject/0x-api/commit/03f7673baed642e07d54b95b30c9c59ba1ab9033

src/schemas/swap_quote_request_schema.json
feuGeneA: How's this: https://github.com/0xProject/0x-api/pull/162/commits/bb2814a870d8c874f6f83b4e92d185a25c434ade

feuGeneA:
deploy kovan-staging

src/handlers/swap_handlers.ts
```diff
@@ -166,6 +177,13 @@ const parseGetSwapQuoteRequestParams = (req: express.Request): GetSwapQuoteReque
             ? undefined
             : parseStringArrForERC20BridgeSources(req.query.excludedSources.split(','));
     const affiliateAddress = req.query.affiliateAddress;
+    const rfqt =
+        req.query.intentOnFilling === undefined
+            ? undefined
+            : { intentOnFilling: req.query.intentOnFilling === 'true' ? true : false };
```
steveklebanoff, line 183: Nit but can't this just be `{ intentOnFilling: req.query.intentOnFilling === 'true' }`

src/schemas/swap_quote_request_schema.json
```diff
@@ -27,11 +27,11 @@
         },
         "intentOnFilling": {
             "type": "string",
-            "allowEmptyValue": true
+            "enum": ["true", "false"]
         },
         "skipValidation": {
             "type": "string",
-            "allowEmptyValue": true
+            "enum": ["true", "false"]
```
steveklebanoff, line 34: ❤️

src/handlers/swap_handlers.ts
feuGeneA: Addressed in https://github.com/0xProject/0x-api/pull/162/commits/735fe3e934b3f27397f09d8f0644a8c8271a6210

PR #168, opened 2020-04-13
[WIP] Integration Test Framework

jalextowle:
## Overview
There are several problems with the API's current testing infrastructure. This PR intends to solve all of them by creating a testing framework that melds many of the earlier solutions together into a cohesive piece of infrastructure that can be maintained by the 0x-api team.


src/db_connection.ts
```diff
@@ -2,10 +2,14 @@ import { Connection, createConnection } from 'typeorm';
...
+let connection: Connection;
+
 /**
  * Creates the DB connnection to use in an app
  */
 export async function getDBConnectionAsync(): Promise<Connection> {
+    if (!connection) {
+        connection = await createConnection(config);
+    }
     return connection;
```
xianny, line 14: nice 👍

test/framework/actions.ts
```diff
@@ -0,0 +1,12 @@
...
+
+const API_HTTP_ADDRESS = 'http://localhost:3000';
+
+/**
+ * Makes a HTTP GET request.
+ * @param input Specifies the route and the base URL that should be used to make
+ *        the HTTP GET request.
+ */
+export async function apiGetRequestAsync(input: { route: string; baseURL?: string }): Promise<request.Response> {
+    return request(input.baseURL || API_HTTP_ADDRESS).get(input.route);
```
xianny, line 11: How about lifting the `request` options interface?

Something like this:
```
function requestAsync(requestOpts: RequestOpts): Promise<request.Response> {
    return request(requestOpts);
}
```

I believe `request` options are defined as `CoreOptions` in `@types/request`. 

Essentially this function would be a thin wrapper around the request library of our choice, and we wouldn't have to write a whole bunch of boilerplate around handling different types of requests. Would also make it easier to switch out for a different request library down the road (`request` is deprecated as of February 2020!)

Another nice feature might be to persist the `request` module in the TestManager and allow the user to set defaults by calling `request.defaults(options)`. This is more of a nice-to-have, feel free to pass for now.

test/framework/assertions.ts
```diff
@@ -0,0 +1,17 @@
...
+ *        should be compared with the actual result.
+ */
+export async function assertFieldEqualsAsync(
+    actual: any,
+    input: {
+        field: string;
+        value: any;
+    },
+): Promise<void> {
+    expect(actual[input.field]).to.be.deep.eq(input.value);
```
xianny, line 16: nice!

test/framework/test_manager.ts
```diff
@@ -0,0 +1,112 @@
...
+ * @returns TestManager Returns a `TestManager` that has been given default actions
+ *          and assertions.
+ */
+export function defaultTestManager(): TestManager {
+    return new TestManager(
+        {
+            apiGetRequestAsync: actions.apiGetRequestAsync,
+        },
+        {
+            assertFieldEqualsAsync: assertions.assertFieldEqualsAsync,
```
xianny, line 21: nit: would probably rename this to `assertEquals` for conciseness, anticipating a whole slew of `assertGreaterThan`, `assertLessThan`, etc.

test/framework/test_manager.ts
```diff
@@ -0,0 +1,112 @@
...
+        },
+    );
+}
+
+export type ActionType = string;
+
+export type AssertionType = string;
+
+export interface ActionInfo {
```
xianny, line 30: Maybe just `Action` and `Assertion`?

```
Action {
    name: ActionName,
    input: ...
}

Assertion {
    name: AssertionName,
    input: ...
}
```

test/app_test.ts
```diff
@@ -1,57 +1,47 @@
...
+            actionType: 'apiGetRequestAsync',
+            input: {
+                route: `${SRA_PATH}/orders`,
+            },
+        },
+        assertions: [
+            {
+                assertionType: 'assertFieldEqualsAsync',
+                input: {
+                    field: 'status',
```
xianny, line 21: nit: instead of `field` I would suggest using `path` since it implies handling nested fields.

test/framework/test_manager.ts
```diff
@@ -0,0 +1,112 @@
...
+        if (!actionAsync) {
+            throw new Error('[test-manager] action is not registered');
+        }
+        return actionAsync(action.input);
+    }
+
+    protected async _executeAssertionAsync(assertion: AssertionInfo, actualResult: any): Promise<void> {
+        const assertionAsync = this._assertionsAsync[assertion.assertionType];
+        if (!assertionAsync) {
+            throw new Error('[test-manager] assertion is not registered');
```
xianny, line 108: let's include the name of the assertion too so it's easier to troubleshoot!

test/utils/deployment.ts
```diff
@@ -0,0 +1,121 @@
...
+    data.filter((datum: string) => datum !== '').map((datum: string) => {
+        log.log(prefix, datum.trim());
+    });
+}
+
+async function sleepAsync(duration: number): Promise<void> {
+    return new Promise<void>(resolve => {
+        setTimeout(resolve, duration);
+    });
+}
```
xianny, line 121: deployment is pretty sweet 🙌  any idea how much time processes take to start up/tear down this way?

One thing I see needing is a way to delete all orders from the postgres cache between each test. We don't necessarily need to wipe all the deps, just clear the saved orders. Would this be an appropriate place for a function like that?

test/app_test.ts
```diff
@@ -1,57 +1,47 @@
...
-                expect(response.body.records).to.deep.equal([]);
-            });
-    });
-});
+const manager = defaultTestManager();
+const suite: TestCase[] = [
+    {
+        description: 'should respond to GET /sra/orders',
+        action: {
+            actionType: 'apiGetRequestAsync',
```
fabioberger, line 12: Could we make this an enum?

test/framework/assertions.ts
fabioberger: Will `expect` log a helpful error message if it fails?

test/utils/deployment.ts
```diff
@@ -0,0 +1,121 @@
...
+let yarnStartProcess: ChildProcessWithoutNullStreams;
+
+/**
+ * The configuration object that provides information on how verbose the logs
+ * should be.
+ * @param shouldPrintApiLogs Whether or not the 0x-api logs should be surfaced.
+ * @param shouldPrintDependencyLogs Whether or not the 0x-api's dependencies
+ *        should surface their logs.
+ * TODO(jalextowle): It would be a good improvement to be able to specify log
+ * files where the logs should actually be written.
```
fabioberger, line 16: 👍

fabioberger: As of now, does it output the logs to the same stdout as the test results?

test/utils/deployment.ts
```diff
@@ -0,0 +1,121 @@
...
+            neatlyPrintChunk('[docker-compose up]', chunk);
+        });
+        up.stderr.on('data', chunk => {
+            neatlyPrintChunk('[docker-compose up | error]', chunk);
+        });
+    }
+    // Wait for the dependencies to boot up.
+    // HACK(jalextowle): This should really be replaced by log-scraping, but it
+    // does appear to work for now.
+    await sleepAsync(10000); // tslint:disable-line:custom-no-magic-numbers
```
fabioberger, line 88: Yikes, does it take that look to boot up? Would log scraping make this faster?

test/utils/deployment.ts
```diff
@@ -0,0 +1,121 @@
...
+        cwd: apiRootDir,
+    });
+    if (shouldPrintLogs) {
+        down.stdout.on('data', chunk => {
+            neatlyPrintChunk('[docker-compose down]', chunk);
+        });
+        down.stderr.on('data', chunk => {
+            neatlyPrintChunk('[docker-compose down | error]', chunk);
+        });
+    }
```
fabioberger, line 107: So, the `Mesh` instance launched by `docker-compose up` is writing to `0x_mesh` directory. Can we delete at after a run? Additionally, the Postgres DB changes are also persisted to `postgres` I believe.

test/framework/assertions.ts
jalextowle: As of now, no. I'll add that as an optional parameter to the `AssertionInfo` interface. I'm going to improve error handling today, which should mitigate this to an extent.

test/utils/deployment.ts
jalextowle: That's the upper limit (mostly because of `docker pull`'s inconsistent timing on different runs. I think log scraping will make this considerably faster most of the time.

test/utils/deployment.ts
jalextowle: Yep. Until the log output can be redirected to a file, I've been seeing this as a feature that developers will use while writing and debugging tests, but not as much afterwards.

test/app_test.ts
jalextowle: We can, but that will really reduce the flexibility of the framework (this might be a good thing, but I'm not convinced of that right now). Something that might be a good middle ground is to create constants for common actions and assertions. I'll add this in the next commit.

test/utils/deployment.ts
jalextowle: Yep, this was an oversight on my part. 

I may add archival logic so that you don't actually use the state, but you get a clean slate each time.

test/utils/deployment.ts
jalextowle: The sleeps provide a hard cap on how long it takes. As per Fabio's comment below, I am now deleting the database and mesh records, which added several seconds. It's looking like around ~15 - 20s for setup and very few (1 - 4s at most) for teardown.

I haven't really worked on this at all since I'm just trying to get the framework out the door so we can start to write some tests, but I will look into scraping the logs this afternoon.

test/framework/test_manager.ts
jalextowle: I like the `name` suggestion. I'll definitely add that. 

As far as the `ActionInfo` => `Action` and `AssertionInfo` => `Assertion, I'm not completely sold yet. I originally used this naming convention, but test cases really just describe the actions and assertions that must be used rather than "being" the actions and assertions. Furthermore, I think that `Action` and `Assertion` better describe the types which I ultimately named `Action` and `Assertion`. What are your thoughts on this?

PR #169, opened 2020-04-14
Alert on orderbook syncing errors

xianny:
Throw a specific error `OrderWatcherSyncError` when there are orderbook syncing errors.

In addition, when syncing the orderbook or responding to SRA requests, look for expired orders and manually filter them out, so they are not saved in Postgres or returned in SRA responses.

"Expired" is defined as 10s past current time.

If any expired orders are found to be older than a threshold, log an error `ExpiredOrderError`. The threshold is 3 mins. 

These are both default values and are configurable in env variables. 

Both errors are defined in `errors.ts` and should have a consistent error message that we can search for and alert on.


xianny:
@opaolini can you review the changes in `errors.ts` please? These are the logging changes we discussed.

src/services/order_watcher_service.ts
```diff
@@ -13,37 +14,46 @@ export class OrderWatcherService {
...
             sent: signedOrders.length,
         });
-        // Remove all of the rejected orders
-        if (rejected.length > 0) {
-            await this._onOrderLifeCycleEventAsync(
-                OrderWatcherLifeCycleEvents.Removed,
-                meshUtils.orderInfosToApiOrders(rejected),
-            );
+
+        // 4. Post accepted orders to Mesh and notify if any expired orders were accepted
```
fragosti, line 40: nit: we aren't posting anything here right? Just alerting.

src/errors.ts
```diff
@@ -112,3 +116,43 @@ export enum ValidationErrorCodes {
     InternalError = 1008,
     TokenNotSupported = 1009,
 }
+
+export abstract class AlertError {
+    public abstract message: string;
+    public shouldAlert: boolean = true;
```
opaolini, line 122: I don't see this public boolean used anywhere, is that a `TODO` ?

src/errors.ts
```diff
@@ -112,3 +116,43 @@ export enum ValidationErrorCodes {
...
+}
+
+export class ExpiredOrderError extends AlertError {
+    public message = `Found expired order! `;
```
opaolini, line 126:
```suggestion
    public message = `Found expired order!`;
```

src/errors.ts
```diff
@@ -112,3 +116,43 @@ export enum ValidationErrorCodes {
...
+        const descSort = expired.sort((a, b) =>
+            b.order.expirationTimeSeconds.minus(a.order.expirationTimeSeconds).toNumber(),
+        );
+        const maxExpiration = descSort[0].order.expirationTimeSeconds;
+        if (maxExpiration > MAX_EXPIRATION_ALERT_THRESHOLD_SECONDS) {
+            const error = new ExpiredOrderError(descSort[0].order, MAX_EXPIRATION_ALERT_THRESHOLD_SECONDS, details);
+            logger.error(error);
+        }
+    }
+}
```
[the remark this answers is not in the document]
xianny: using a `find` instead

src/errors.ts
```diff
@@ -112,3 +116,43 @@ export enum ValidationErrorCodes {
...
+ */
+export function alertOnExpiredOrders(expired: APIOrder[], details?: string): void {
+    if (expired.length > 0) {
+        const descSort = expired.sort((a, b) =>
+            b.order.expirationTimeSeconds.minus(a.order.expirationTimeSeconds).toNumber(),
+        );
+        const maxExpiration = descSort[0].order.expirationTimeSeconds;
+        if (maxExpiration > MAX_EXPIRATION_ALERT_THRESHOLD_SECONDS) {
+            const error = new ExpiredOrderError(descSort[0].order, MAX_EXPIRATION_ALERT_THRESHOLD_SECONDS, details);
+            logger.error(error);
```
opaolini, line 155: If we will use the default object serialiser, we might include unnecessary fields (such as `shouldAlert` , just something to be mindful of :)

xianny: I was thinking the default serialiser will include all fields in the object, is that right?

src/errors.ts
xianny: it's not used anywhere, but I was envisioning it being useful when searching logs.

E.g. we could search for `parsed_log.shouldAlert = true` and use that as a Grafana alert rule.

src/errors.ts
xianny: Ah good catch, thanks. this was leftover from when I was concatenating the details into the same string!

src/errors.ts
opaolini: Yes

src/errors.ts
opaolini: Ah I see! So basically a flag that resembles a higher log level

src/errors.ts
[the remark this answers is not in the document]
xianny: how about `logger.ts`?

xianny:
deploy staging

PR #170, opened 2020-04-15
RFQT: More tests

steveklebanoff:
Expands RFQ-T testing

* Test for not returning an order when maker does _not_ have allowances set
* Test for `skipValidation` being default set to `false`
* Test for `skipValidation` behaving as expected when set to `true`
* Test for a bad API key being sent in

### DIFF
--- a/tsconfig.json
+++ b/tsconfig.json
@@ -9,7 +9,7 @@
         "noUnusedLocals": true,
         "emitDecoratorMetadata": true,
         "experimentalDecorators": true,
-        "sourceMap": false,
+        "sourceMap": true,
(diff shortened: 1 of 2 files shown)


tsconfig.json
```diff
@@ -9,7 +9,7 @@
         "noUnusedLocals": true,
         "emitDecoratorMetadata": true,
         "experimentalDecorators": true,
-        "sourceMap": false,
+        "sourceMap": true,
```
steveklebanoff, line 12: I confirmed w/ @fragosti that this is cool, and is already being done in https://github.com/0xProject/0x-api/pull/168

steveklebanoff: I did this so I can get proper TS-aware stack traces when tests fail

PR #172, opened 2020-04-17
RFQ-T Indicative Quotes

feuGeneA:
Adds a `/swap/v0/price` endpoint, delegating almost entirely to the handler for the `/swap/v0/quote` endpoint.

This depends on https://github.com/0xProject/0x-monorepo/pull/2555 and is pinned to a revision there in the `TEMPORARY` commit here, which will be updated when that PR lands.

-   [ ] Expand test cases.  There's only a single one right now.


feuGeneA:
deploy staging

feuGeneA:
deploy kovan-staging

src/config.ts
```diff
@@ -159,6 +159,7 @@ const EXCLUDED_SOURCES = (() => {
 
 const gasSchedule: { [key in ERC20BridgeSource]: number } = {
     [ERC20BridgeSource.Native]: 1.5e5,
+    [ERC20BridgeSource.Rfqt]: 1.5e5,
```
steveklebanoff, line 162: I totally could be missing something, but is this still relevant? Also see https://github.com/0xProject/0x-monorepo/pull/2555/files#r410517580

I expected Rfqt orders just to be using the ERC20BridgeSource.Native bridge source.

src/handlers/swap_handlers.ts
```diff
@@ -142,34 +190,6 @@ export class SwapHandlers {
             throw new InternalServerError(e.message);
         }
     }
-    // tslint:disable-next-line:prefer-function-over-method
-    public async getSwapTokensAsync(_req: express.Request, res: express.Response): Promise<void> {
```
steveklebanoff, line 146: just to confirm:`getSwapTokensAsync` and `getTokenPricesAsync` just got moved, but didn't change. is that correct?

src/handlers/swap_handlers.ts
```diff
@@ -30,7 +30,55 @@ export class SwapHandlers {
...
+    public async getSwapPriceAsync(req: express.Request, res: express.Response): Promise<void> {
+        const params = parseGetSwapQuoteRequestParams(req);
+        if (params.rfqt === undefined) {
+            params.rfqt = {
+                intentOnFilling: false,
+            };
+        }
+        params.skipValidation = true;
+        const quote = await this._calculateSwapQuoteAsync(params, req.header('0x-api-key'));
+        const { price, value, gasPrice, gas, protocolFee, buyAmount, sellAmount } = quote;
```
steveklebanoff, line 58: I know it's a change to the spec, but I think it's actually a hard requirement that we return `sources` here.  Would you mind adding that?

IMHO I think we need this so wallet UXs can show the order sources to the user before they review their order

src/handlers/swap_handlers.ts
feuGeneA: Yes, that is correct.  I tried to explain that in the commit message for https://github.com/0xProject/0x-api/pull/172/commits/66c9d9dd3aa71b96b2a3b85000f9c0a44bc0bffd, but it was a complicated write-up...

src/config.ts
dekz: Usage of this is for a few reasons:
1) Exclude certain sources (controlled via API params) _[Would you disable Native but not RFQT?]_
2) Track internally in asset-swapper to avoid conflicting sources _[There's no conflicts with 0x orders]_
3) Calculate cost based off the source _[this is the same cost as Native]_

If we don't see ourselves doing the above, or if it is coupled to `Native`, then there may not be a need for it?

With the 0xAPI we'd definitely want to consider merging in `Rfqt` and `Native` to the composition source breakdown and show it as `0x`.

src/handlers/swap_handlers.ts
```diff
@@ -30,7 +30,55 @@ export class SwapHandlers {
...
+                    reason: `Could not find token ${symbolOrAddress}`,
+                },
+            ]);
+        }
+        const unitAmount = new BigNumber(1);
+        const records = await this._swapService.getTokenPricesAsync(baseAsset, unitAmount);
+        res.status(HttpStatus.OK).send({ records });
+    }
+    private async _calculateSwapQuoteAsync(
+        params: GetSwapQuoteRequestParams,
```
dekz, line 79: feel like apiKey should live in `GetSwapQuoteRequestParams`. Don't think it needs to travel separately just because it originates via a Header.

src/services/swap_service.ts
```diff
@@ -97,23 +96,24 @@ export class SwapService {
...
         } = params;
+        let _rfqt;
+        if (apiKey !== undefined && (isETHSell || from !== undefined)) {
+            _rfqt = {
+                ...rfqt,
+                apiKey,
+                // If this is a forwarder transaction, then we want to request quotes with the taker as the
+                // forwarder contract. If it's not, then we want to request quotes with the taker set to the
+                // API's takerAddress query parameter, which in this context is known as `from`.
+                takerAddress: isETHSell ? getContractAddressesForChainOrThrow(CHAIN_ID).forwarder : from || '',
```
dekz, line 107: In the `swap_service` constructor we already fetch the addresses for the chain. We may as well store them and pull `forwarder` off of this.

is `''` an a silent RFQT error vs `undefined` vs `NULL_ADDRESS`?

src/handlers/swap_handlers.ts
dekz: definitely helpful for an implementor to show the breakdown

src/config.ts
feuGeneA: I took this out for now, in https://github.com/0xProject/0x-api/pull/172/commits/1d9ad66b0f6720442eab037a1144c4027f66eca4.  See also https://github.com/0xProject/0x-monorepo/pull/2555#discussion_r410517580

I'm up for using this for something cool, but I'd rather not do it in this PR.

src/services/swap_service.ts
feuGeneA: I addressed the address pre-loading in https://github.com/0xProject/0x-api/pull/172/commits/ee23cab5ff5c77aeae3406716f2823d58b218ec9.

There's nothing special about `''`.  Is there a reason to choose `undefined` or `NULL_ADDRESS`?

src/handlers/swap_handlers.ts
feuGeneA: Addressed in https://github.com/0xProject/0x-api/pull/172/commits/1b9b20c212f7d5c0bf698f77e04a95dc6c93ec35

src/handlers/swap_handlers.ts
feuGeneA: Addressed in https://github.com/0xProject/0x-api/pull/172/commits/cc3b686cc78a09ba48ca818fd73ec3cace73928c

feuGeneA:
deploy staging

feuGeneA:
deploy staging

package.json
```diff
@@ -66,7 +66,7 @@
     },
     "dependencies": {
         "@0x/assert": "^3.0.4",
-        "@0x/asset-swapper": "0xProject/gitpkg-registry#0x-asset-swapper-v4.4.0-a458e81f8",
+        "@0x/asset-swapper": "0xProject/gitpkg-registry#0x-asset-swapper-v4.4.0-a9f445173",
```
steveklebanoff, line 69: NOTE: this needs to be updated now, correct?

steveklebanoff:
LGTM pending the two items mentioned.  Would be good to get @dekz blessing here though.

src/handlers/swap_handlers.ts
```diff
@@ -30,7 +30,51 @@ export class SwapHandlers {
...
+        const params = parseGetSwapQuoteRequestParams(req);
+        if (params.rfqt === undefined) {
+            params.rfqt = {
+                intentOnFilling: false,
+            };
+        }
+        params.skipValidation = true;
+        const quote = await this._calculateSwapQuoteAsync(params);
+        const { price, value, gasPrice, gas, protocolFee, buyAmount, sellAmount, sources, orders } = quote;
+        logger.info(`Serving indicative quote based on the following orders: ${JSON.stringify(orders)}`);
```
dekz, line 57: This would get a little verbose. Can we remove it?

feuGeneA: Addressed in https://github.com/0xProject/0x-api/pull/172/commits/b2f6fe76188ca9bfff737cb40a56826366554f5f

feuGeneA:
deploy staging

src/handlers/swap_handlers.ts
```diff
@@ -31,7 +31,47 @@ export class SwapHandlers {
...
+            name: tm.name,
+            decimals: tm.decimals,
+        }));
+        const filteredTokens = tokens.filter(t => t.address !== NULL_ADDRESS);
+        res.status(HttpStatus.OK).send({ records: filteredTokens });
+    }
+    // tslint:disable-next-line:prefer-function-over-method
+    public async getSwapPriceAsync(req: express.Request, res: express.Response): Promise<void> {
+        const params = parseGetSwapQuoteRequestParams(req, 'price');
+        params.skipValidation = true;
```
steveklebanoff, line 52: nit but it'd be nice to have a comment explaining what's going on here

src/services/swap_service.ts
```diff
@@ -80,23 +81,25 @@ export class SwapService {
...
+        let _rfqt;
+        if (apiKey !== undefined && (isETHSell || from !== undefined)) {
+            _rfqt = {
+                ...rfqt,
+                intentOnFilling: rfqt && rfqt.intentOnFilling ? true : false,
+                apiKey,
+                // If this is a forwarder transaction, then we want to request quotes with the taker as the
+                // forwarder contract. If it's not, then we want to request quotes with the taker set to the
+                // API's takerAddress query parameter, which in this context is known as `from`.
+                takerAddress: isETHSell ? this._forwarderAddress : from || '',
```
steveklebanoff, line 93: I expected this to be `|| undefined`, does it have to be `|| ''`?

src/types.ts
```diff
@@ -433,7 +435,8 @@ export interface CalculateSwapQuoteParams {
     affiliateAddress?: string;
     apiKey?: string;
     rfqt?: {
-        intentOnFilling: boolean;
+        intentOnFilling?: boolean;
```
steveklebanoff, line 438: nit but I'd prefer for both of these to be required

steveklebanoff:
deploy staging

feuGeneA:
deploy staging

PR #174, opened 2020-04-17
Fix expired order filter, better error logging

xianny:
`BigNumber` fields in `SignedOrder` were not being cast and instead were being passed around as `number` primitives, leading to failures when calling `.gt()`. This PR calls `>` instead of `.gt()`. 

Also start logging `error.stack` instead of `error` everywhere. Logging `error` shows up as `Error [[Object object]]` whereas `error.stack` is informative.

successful deploy and OrderWatcherService sync logs here: https://kibana.mesh.0x.org/goto/05c2f0125ec0772142300e7d7cc3f69d

### DIFF
--- a/src/app.ts
+++ b/src/app.ts
@@ -55,7 +55,7 @@ export async function getDefaultAppDependenciesAsync(
     try {
         swapService = createSwapServiceFromOrderBookService(orderBookService, provider);
     } catch (err) {
-        logger.error(err);
+        logger.error(err.stack);
     }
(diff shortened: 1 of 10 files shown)


xianny:
deploy staging

xianny:
deploy production

xianny:
deploy staging

xianny:
deploy staging

xianny:
deploy production

PR #177, opened 2020-04-20
Setup and Teardown Functions

jalextowle:
## Description
Adds setup and teardown functions that can be used while testing the 0x-api.


package.json
```diff
@@ -11,7 +11,7 @@
         "clean:ts": "shx rm -rf lib",
         "clean:docker": "shx rm -rf 0x_mesh/db",
         "build": "tsc -p tsconfig.json",
-        "test": "ETHEREUM_RPC_URL=http://localhost:8545 CHAIN_ID=1337 RFQT_API_KEY_WHITELIST='koolApiKey1,koolApikey2' RFQT_MAKER_ENDPOINTS='https://mock-rfqt1.club' mocha --require source-map-support/register --require make-promises-safe lib/test/**/*_test.js --timeout 100000 --exit",
+        "test": "ETHEREUM_RPC_URL=http://localhost:8545 CHAIN_ID=1337 RFQT_API_KEY_WHITELIST='koolApiKey1,koolApikey2' RFQT_MAKER_ENDPOINTS='https://mock-rfqt1.club' mocha --require source-map-support/register --require make-promises-safe lib/test/**/*_test.js --timeout 200000 --exit",
```
fragosti, line 14: Are we defining the test env here or in that object in the test file?

test/utils/deployment.ts
```diff
@@ -0,0 +1,172 @@
...
+
+    // Spin up the 0x-api dependencies
+    const up = spawn('docker-compose', ['up'], {
+        cwd: apiRootDir,
+        env: {
+            ...process.env,
+            ETHEREUM_RPC_URL: 'http://ganache:8545',
+            ETHEREUM_CHAIN_ID: '1337',
+        },
+    });
```
fragosti, line 89: Since the tests do this for you now, we should modify the README.

test/utils/deployment.ts
```diff
@@ -0,0 +1,172 @@
...
+                ) {
+                    hasSeenLog[2]++;
+                }
+
+                if (hasSeenLog[0] === 1 && hasSeenLog[1] === 1 && hasSeenLog[2] === 2) {
+                    resolve();
+                }
+            }
+        });
+    });
```
fragosti, line 171: I could see this never resolving if one of these dependencies is updated and the log changes. Should we reject with a helpful msg if this takes say, more than 20seconds? You'd know the best value for the seconds.

test/utils/deployment.ts
```diff
@@ -0,0 +1,172 @@
...
+    return new Promise<void>(resolve => {
+        logStream.stdout.on('data', (chunk: Buffer) => {
+            const data = chunk.toString().split('\n');
+            for (const datum of data) {
+                if (/API (HTTP) listening on port 3000!/.test(datum)) {
+                    resolve();
+                }
+            }
+        });
+    });
```
fragosti, line 143: This could be another accidental awaiting forever case. Should we add a timeout?

fragosti:
Just tried running it. Got
```
francescoagosti@0x:Projects/0x-api ‹test/setup-and-teardown*›$ yarn test
yarn run v1.13.0
$ ETHEREUM_RPC_URL=http://localhost:8545 CHAIN_ID=1337 RFQT_API_KEY_WHITELIST='koolApiKey1,koolApikey2' RFQT_MAKER_ENDPOINTS='https://mock-rfqt1.club' mocha --require source-map-support/register --require make-promises-safe lib/test/**/*_test.js --timeout 200000 --exit


  app test
    1) "before all" hook for "should respond to GET /sra/orders"


  0 passing (4m)
  1 failing

  1) app test
       "before all" hook for "should respond to GET /sra/orders":
     Error: Timeout of 200000ms exceeded. For async tests and hooks, ensure "done()" is called; if returning a Promise, ensure it resolves. (/Users/francescoagosti/Projects/0x-api/lib/test/app_test.js)




error Command failed with exit code 1.
info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.
```

package.json
jalextowle: We're defining it in that object in the test file. I'll remove this today

.circleci/config.yml
```diff
@@ -43,6 +43,17 @@ jobs:
                       - node_modules
                   key: v1-dependencies-{{ checksum "yarn.lock" }}
             - run: yarn build
+            - run: ls
             - run: yarn lint
             - run: yarn prettier:ci
+            - setup_remote_docker # Expose the docker daemon so that it can be used in tests.
+            - run: yarn test
+            - run: docker ps
+            - run: ls
```
xianny, line 52: what's `docker ps` and `ls` for? Just for debugging?

test/utils/deployment.ts
```diff
@@ -111,14 +110,16 @@ export async function setupDependenciesAsync(logType: LogType = LogType.Hidden):
...
-            resolve();
-        });
-    });
+    await waitForCloseAsync(
+        spawn('docker-compose', ['down'], {
+            cwd: apiRootDir,
+        }),
+    );
+    await waitForCloseAsync(
+        spawn('docker-compose', ['rm', '-v'], {
```
xianny, line 119: could be timing out if it's prompting for confirmation? use `-f` flag to override

.circleci/config.yml
jalextowle: Yep! I will remove once I figure out the underlying problem

package.json
jalextowle: This actually requires a bigger refactor within the API for this to work. We still need to add some of the custom envvars to the `docker-compose.yml` though. For example, the `ETHEREUM_RPC_URL` needs to be `http://ganache:8545` for `docker-compose up` to work correctly.

jalextowle: I think it would be worth it to get the `CHAIN_ID` from the `provider` if that wouldn't be too much work. It seems somewhat out of scope for this PR though.

xianny:
Awesome work! Test logs/output are looking great too 💯

PR #178, opened 2020-04-21
Extended buyAmount sources

dekz:
Adds additional sources when `buyAmount` is used as an API parameter. Previously Kyber and Curve were not used. 

Now Curve has support for exact sampling of buy amounts. For Kyber we approximate with a 5bps tolerance.


dekz:
deploy staging

dekz:
deploy production

PR #184, opened 2020-04-22
[WIP] Limit axios retrying to only be for mesh client

steveklebanoff:


src/utils/axios_utils.ts
```diff
@@ -1,12 +1,9 @@
...
+import Axios from 'axios';
 import * as rax from 'retry-axios';
 
 import { logger } from '../logger';
 
-// Attach retry-axios to the global instance
-rax.attach();
-
-const DEFAULT_RETRY_CONFIG: rax.RetryConfig = {
+export const DEFAULT_AXIOS_RETRY_CONFIG: rax.RetryConfig = {
```
PirosB3, line 6: If these configs are only used for 1 single occurence (in `MeshClient`), can we simply create an instance of `retryableAxios` inside of `MeshClient` constructor? is there any need to define anything in `axios_utills`?

steveklebanoff: I agree w/ you, but I was trying to follow the suggestion @fragosti put in the original PR: https://github.com/0xProject/0x-api/pull/160#discussion_r404456496

steveklebanoff:
Replaced with https://github.com/0xProject/0x-api/pull/185

PR #185, opened 2020-04-22
Make axios retryable only when desired

steveklebanoff:
Change retry-axios to only apply to an explicit instance, instead of being on the global Axios object.

This fixes the code so RFQT requests are not retried


steveklebanoff:
deploy staging

PR #192, opened 2020-04-23
feat: Add TransactionWatcherService

opaolini:
# About

This PR adds:

* a `TransactionWatcherService` which is responsible for:
  - monitoring transactions sent by the `SignerService`
  - tagging `transactions` as `stuck` based on a defined expected time to be mined
* a `/meta_transaction/v0/status` handler that returns status of transactions from the database
* a simple test using `rinkeby` testnet

Following this PR we will:
- Add support for multiple relayers
- Better unsticking logic -> instead of unsticking all stuck transactions at once, add time to transactions with subsequent nonces that are also stuck.
- Handling of blockchain re-orgs


src/services/signer_service.ts
```diff
@@ -173,12 +205,93 @@ export class SignerService {
...
+        });
+        await this._transactionEntityRepository.save(transactionEntity);
+        const rawTx = `0x${ethTx.serialize().toString('hex')}`;
+        await this._web3Wrapper.sendRawPayloadAsync({
+            method: 'eth_sendRawTransaction',
+            params: [rawTx],
+        });
+        await this._updateTransactionEntityToSubmittedAsync(txHashHex);
+        transactionEntity.status = TransactionStates.Submitted;
+        await this._transactionEntityRepository.save(transactionEntity);
```
fabioberger, line 272: Remove these two lines. We already update it's status with call to `_updateTransactionEntityToSubmittedAsync`.

src/services/signer_service.ts
```diff
@@ -173,12 +205,93 @@ export class SignerService {
...
+        return txHashHex;
+    }
+    private async _updateTransactionEntityToSubmittedAsync(txHash: string): Promise<boolean> {
+        // if the transaction was not updated in the meantime, we change its status to Submitted.
+        try {
+            await this._transactionEntityRepository.manager.transaction(async transactionEntityManager => {
+                const repo = transactionEntityManager.getRepository(TransactionEntity);
+                const newTx = await repo.findOne(txHash);
```
fabioberger, line 280: nit: rename `newTx` to `txn`

src/services/transaction_watcher_service.ts
```diff
@@ -0,0 +1,116 @@
...
+    constructor(dbConnection: Connection, provider: SupportedProvider) {
+        this._transactionRepository = dbConnection.getRepository(TransactionEntity);
+        this._provider = provider;
+        this._web3Wrapper = new Web3Wrapper(this._provider);
+    }
+    public async startAsync(): Promise<void> {
+        while (true) {
+            // finding transactions that need updating for which we have missed
+            // notifications from blocknative.
+            logger.trace('syncing transaciton status');
```
fabioberger, line 24:
```suggestion
            logger.trace('syncing transaction status');
```

src/services/transaction_watcher_service.ts
```diff
@@ -0,0 +1,116 @@
...
+                await this.syncTransactionStatusAsync();
+            } catch (err) {
+                logger.error(`order watcher failed to sync transaction status`, { err: err.stack });
+            }
+            await utils.delay(TX_WATCHER_POLLING_INTERVAL_IN_MS);
+        }
+    }
+    public async syncTransactionStatusAsync(): Promise<void> {
+        await this._syncTransactionStatusAsync();
+    }
```
fabioberger, line 35: This method is superfluous. Nix it.

src/services/transaction_watcher_service.ts
```diff
@@ -0,0 +1,116 @@
...
+        // const now = new Date();
+        // const transactionsToCheck = await this._transactionRepository.find({
+        //     where: [
+        //         {
+        //             status: TransactionStates.Mempool,
+        //             expiresAt: LessThanOrEqual(now),
+        //         },
+        //         {
+        //             status: TransactionStates.Submitted,
+        //             expiresAt: LessThanOrEqual(now),
```
fabioberger, line 64: Does LessThan work? We could also use that 🤷

src/services/transaction_watcher_service.ts
```diff
@@ -0,0 +1,116 @@
...
+                { status: TransactionStates.Stuck },
+            ],
+        });
+        logger.trace(`found ${transactionsToCheck.length} transactions to check status`);
+        for (const tx of transactionsToCheck) {
+            await this._findTransactionStatusAndUpdateAsync(tx);
+        }
+    }
+
+    private async _findTransactionStatusAndUpdateAsync(tx: TransactionEntity): Promise<TransactionEntity> {
```
fabioberger, line 52:
```suggestion
    private async _findTransactionStatusAndUpdateAsync(txEntity: TransactionEntity): Promise<TransactionEntity> {
```

src/services/transaction_watcher_service.ts
```diff
@@ -0,0 +1,116 @@
...
+        //         {
+        //             status: TransactionStates.Submitted,
+        //             expiresAt: LessThanOrEqual(now),
+        //         },
+        //     ],
+        // });
+        const now = new Date();
+        const isExpired = tx.expectedAt <= now;
+        try {
+            const txInBlockchain = await this._web3Wrapper.getTransactionByHashAsync(tx.hash);
```
fabioberger, line 71:
```suggestion
            const tx = await this._web3Wrapper.getTransactionByHashAsync(txEntity.hash);
```

src/services/transaction_watcher_service.ts
```diff
@@ -0,0 +1,116 @@
...
+            if (txInBlockchain !== undefined && txInBlockchain !== null && txInBlockchain.hash !== undefined) {
+                if (txInBlockchain.blockNumber !== null) {
+                    logger.info(
+                        `a transaction with a ${tx.status} status is already on the blockchain, updating status to confirmed`,
```
fabioberger, line 75: This seems outdated now that this isn't just a cleanup job, but run as a regular cron. This case is expected now that we aren't using BlockNative, right? If so, let's remove this log.

src/services/transaction_watcher_service.ts
```diff
@@ -0,0 +1,116 @@
...
+                        { hash: txInBlockchain.hash },
+                    );
+
+                    tx.status = TransactionStates.Confirmed;
+                    return this._saveTransactionAsync(tx);
+                } else if (!isExpired && tx.status !== TransactionStates.Mempool) {
```
fabioberger, line 81: Add a comment above this conditional explaining what it's checking. E.g., 

```
// Checks if the txn is in the mempool but still has it's status set to Unsubmitted or Submitted
```

☝️ That's my understanding of it. Took me a while to get to that though, so this comment should help others grok this quicker.

src/services/transaction_watcher_service.ts
```diff
@@ -0,0 +1,116 @@
...
+        } catch (err) {
+            // TODO(oskar) this throws a type error when calling a tx that was
+            // already dropped. Can we catch this specific error?
+            // TypeError: Cannot read property 'blockNumber' of null
+            // at Object.unmarshalTransaction (/Users/overmorrow/projects/0x/0x-api/node_modules/@0x/web3-wrapper/src/marshaller.ts:85:32)
+            // at Web3Wrapper.<anonymous> (/Users/overmorrow/projects/0x/0x-api/node_modules/@0x/web3-wrapper/src/web3_wrapper.ts:253:40)
+            // at step (/Users/overmorrow/projects/0x/0x-api/node_modules/@0x/web3-wrapper/lib/src/web3_wrapper.js:43:23)
+            // at Object.next (/Users/overmorrow/projects/0x/0x-api/node_modules/@0x/web3-wrapper/lib/src/web3_wrapper.js:24:53)
+            // at fulfilled (/Users/overmorrow/projects/0x/0x-api/node_modules/@0x/web3-wrapper/lib/src/web3_wrapper.js:15:58)
+            // at process._tickCallback (internal/process/next_tick.js:68:7)
```
fabioberger, line 101: This is a bug in Web3Wrapper. We should handle the case where `null` is returned instead of a txn, and we don't currently do that. For now, I would explicitly check for this error, and if it's found, do the logic below. If another error is thrown, we should log it and continue.

src/services/transaction_watcher_service.ts
```diff
@@ -0,0 +1,116 @@
...
+                return this._saveTransactionAsync(tx);
+            }
+        }
+        return tx;
+    }
+    private async _saveTransactionAsync(tx: TransactionEntity): Promise<TransactionEntity> {
+        await this._transactionRepository.manager.transaction(async transactionalEntityManager => {
+            const repo = transactionalEntityManager.getRepository(TransactionEntity);
+            await repo.save(tx);
+        });
```
fabioberger, line 113: This doesn't do much. Given that TransactionWatcher is a singleton instance, there is only one process attempting to update a given txn at any given time. We therefore don't need a txn here.

src/typings.d.ts
```diff
@@ -0,0 +1 @@
+declare module 'ethereumjs-tx';
```
fabioberger, line 1: Rename this file to `global.d.ts`

fabioberger: And add a TODO comment to flesh out the type interfaces we use from this package.

src/utils/utils.ts
```diff
@@ -71,4 +71,8 @@ export const utils = {
         }
         return chunkedItems;
     },
+    delay: async (ms: number): Promise<{}> => {
+        // tslint:disable-next-line
```
fabioberger, line 75: Don't use blanket tslint disables. Always disable specific linter rules. What are you disabling here?

src/utils/utils.ts
```diff
@@ -71,4 +71,8 @@ export const utils = {
         }
         return chunkedItems;
     },
+    delay: async (ms: number): Promise<{}> => {
```
fabioberger, line 74:
```suggestion
    delay: async (ms: number): Promise<void> => {
```

test/transaction_watcher_service_test.ts
```diff
@@ -0,0 +1,90 @@
...
+let signer: DummySigner;
+let transactionEntityRepository: Repository<TransactionEntity>;
+let txWatcher: TransactionWatcherService;
+let connection: Connection;
+
+const LOW_GAS_PRICE = 1337;
+const MID_GAS_PRICE = 4000000000;
+const HIGH_GAS_PRICE = 9000000000;
+const WAIT_DELAY_IN_MS = 5000;
+const SHORT_EXPECTED_MINE_TIME_IN_S = 15;
```
fabioberger, line 29: nit: 

```suggestion
const SHORT_EXPECTED_MINE_TIME_SEC = 15;
```

test/transaction_watcher_service_test.ts
```diff
@@ -0,0 +1,90 @@
...
+}
+
+describe('transaction watcher service', () => {
+    before(async () => {
+        providerEngine = new Web3ProviderEngine();
+        providerEngine.addProvider(new RPCSubprovider(TEST_RINKEBY_RPC_URL));
+        providerUtils.startProviderEngine(providerEngine);
+        connection = await getDBConnectionAsync();
+        transactionEntityRepository = connection.getRepository(TransactionEntity);
+        signer = new DummySigner(
```
fabioberger, line 54: Perhaps `TestSigner` would be more appropriate? `Dummy` makes me think it won't actually sign anything.

test/transaction_watcher_service_test.ts
```diff
@@ -0,0 +1,90 @@
...
+            connection,
+            TEST_RINKEBY_PRIVATE_KEY,
+            TEST_RINKEBY_PUBLIC_ADDRESS,
+            TEST_RINKEBY_RPC_URL,
+        );
+        txWatcher = new TransactionWatcherService(connection, providerEngine);
+        // tslint:disable-next-line
```
fabioberger, line 61: 😱

test/transaction_watcher_service_test.ts
```diff
@@ -0,0 +1,90 @@
...
+        txWatcher.startAsync();
+    });
+    it('monitors the transaction lifecycle correctly', async () => {
+        // send tx with 1 gwei gas price
+        const txHash = await signer.sendTransactionToItselfAsync(new BigNumber(MID_GAS_PRICE));
+        await waitUntilStatusAsync(txHash, TransactionStates.Confirmed, transactionEntityRepository);
+        const storedTx = await transactionEntityRepository.findOne(txHash);
+        expect(storedTx).to.not.be.undefined();
+        expect(storedTx).to.include({ hash: txHash });
+        expect(storedTx).to.include({ status: TransactionStates.Confirmed });
```
fabioberger, line 71: This is duplicitous with `waitUntilStatusAsync` above no?

test/utils/dummy_signer.ts
```diff
@@ -0,0 +1,160 @@
...
+            });
+        } catch (err) {
+            // the TransacitonEntity was updated in the meantime. This will
+            // rollback the database transaction.
+            logger.warn('failed to store transaction with submitted status, rolling back', { err });
+        }
+        return txHashHex;
+    }
+
+    private async _getNonceAsync(senderAddress: string): Promise<string> {
```
fabioberger, line 137: Move this to utils instead of duplicating.

src/types.ts
```diff
@@ -461,3 +461,12 @@ export interface CalculateMetaTransactionQuoteParams {
...
     excludedSources?: ERC20BridgeSource[];
 }
+
+export enum TransactionStates {
+    Unsubmitted = 'unsubmitted',
+    Submitted = 'submitted',
+    Mempool = 'mempool',
+    Stuck = 'stuck',
+    Confirmed = 'confirmed',
+    Dropped = 'dropped',
```
fabioberger, line 471: We're missing one state: `aborted`. When a txn with a specific sender & nonce get's mined, we should check for other txns with those same sender & nonce, and mark them as `aborted`.

src/services/transaction_watcher_service.ts
opaolini: No it doesn't work for any of these methods. I also tried to use "format" and specific strings, but that still didn't work properly, I'll leave a reference to a Github issue.

src/services/transaction_watcher_service.ts
opaolini: There are two processes, one is a `SignerService` and `TransactionWatcher` service, and it could be the case that `Signer` will update the same transaction with `submitted` status, however we could argue that `TransactionWatcher` update should take precedence (since we do check the status in `SignerService`).

src/constants.ts
```diff
@@ -50,3 +50,10 @@ export const META_TRANSACTION_DOCS_URL = 'https://0x.org/docs/api#meta_transacti
 
 // Signer
 export const ETH_GAS_STATION_API_BASE_URL = 'https://ethgasstation.info';
+export const UNSTICKING_TRANSACTION_GAS_MULTIPLIER = 1.5;
+export const ETH_TRANSFER_GAS_LIMIT = 21000;
+export const SIGNER_POLLING_INTERVAL_IN_MS = 5 * 1000;
+
+// TransactionWatcher
+export const EXPECTED_MINED_IN_S = 120 * 1.5;
```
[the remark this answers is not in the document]
opaolini: Added the comment.

However I took a peek at the FAQ [0] and doesn't say that "2 minutes" is a given for all fast transactions. Should we instead set this dynamically from the ETHGASSTATION's API, and then fallback on this constant time?

[0] - https://ethgasstation.info/FAQcalc.php

src/entities/TransactionEntity.ts
```diff
@@ -0,0 +1,60 @@
...
+    @PrimaryColumn({ name: 'ethereum_tx_hash', type: 'varchar' })
+    public hash: string;
+
+    @Column({ name: 'status', type: 'varchar' })
+    public status: string;
+
+    @Column({ name: 'expected_mined_in_sec', type: 'int' })
+    public expectedMinedInSec?: number;
+
+    @Column({ name: 'nonce', type: 'varchar' })
```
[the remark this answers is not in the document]
opaolini: Moved to number, I also used types based on web3js: https://web3js.readthedocs.io/en/v1.2.0/web3-eth.html#
  - nonce: `Number`
  - blockNumber: `Number`
  - gasPrice: `BigNumber`

Does that look right?

src/entities/TransactionEntity.ts
```diff
@@ -0,0 +1,60 @@
...
+    @UpdateDateColumn({ name: 'updated_at' })
+    public updatedAt?: Date;
+
+    @Column({ name: 'expected_at', type: 'timestamptz' })
+    public expectedAt: Date;
+
+    // HACK(oskar) we want all fields to be set, otherwise we should not accept
+    // a transaction entity, however because of this issue:
+    // https://github.com/typeorm/typeorm/issues/1772 we cannot accept undefined
+    // as an argument to the constructor, to not break migrations with serialize
```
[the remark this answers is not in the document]
opaolini: If we throw inside of the constructor `typeorm` will fail on auto-migration (the `serialize` option does this migration).

We could have a static method as suggested in the issue that would throw an error, however I don't know how viable that is.

src/handlers/meta_transaction_handlers.ts
```diff
@@ -158,6 +157,22 @@ export class MetaTransactionHandlers {
             throw new InternalServerError(e.message);
         }
     }
+
+    public async getTransactionStatusAsync(req: express.Request, res: express.Response): Promise<void> {
+        const transactionHash = req.params.txHash;
```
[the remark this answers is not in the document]
opaolini: Added validation using `assert`

src/services/signer_service.ts
```diff
@@ -158,9 +175,24 @@ export class SignerService {
             signature,
             protocolFee,
         );
-
         const signedEthereumTransaction = await this._privateWalletSubprovider.signTransactionAsync(ethereumTxnParams);
```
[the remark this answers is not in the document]
opaolini: Added `(PrivateKeyWalletSubprovider as any)._validateTxParams(txParams);`

src/services/signer_service.ts
```diff
@@ -158,9 +175,24 @@ export class SignerService {
...
-
+        // HACK(oskar) - we use EthereumTx directly to get the hash of the
+        // underlying transaction to store it as unsubmitted before we broadcast
+        // it. We do it after calling `signTransactionAsync` because web3Wrapper
+        // does some checks for txParams.
+        const tx = new EthereumTx(ethereumTxnParams);
+        tx.sign(this._privateKeyBuffer, true);
+        const txHash = tx.hash() as Buffer;
```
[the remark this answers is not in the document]
opaolini: I  type-cast here to know that it is an actual Buffer (I had to look it up), as we don't get any information from this imported package. Is there any drawback to using type casts in this situation?

The better solution would be to include that in `global.d.ts` ?

src/services/transaction_watcher_service.ts
opaolini: I've moved this log to a `logger.trace` as there might be a benefit to including this information when testing locally.

src/services/transaction_watcher_service.ts
opaolini: Yup added the comment.

src/services/transaction_watcher_service.ts
opaolini: I try to catch it with:
```
if (err instanceof TypeError)
```

Is that the correct way?

src/services/transaction_watcher_service.ts
opaolini: Changed, so that `TransactionWatcher` saves without a DB transaction.

src/utils/utils.ts
opaolini: `tslint:disable-next-line:no-inferred-empty-object-type`

src/utils/utils.ts
opaolini: `void` does not work in this case:
  ```
  src/utils/utils.ts:77:9 - error TS2322: Type '{}' is not assignable to type 'void'.

77         return new Promise(resolve => setTimeout(resolve, ms));
           ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
  ```

test/utils/dummy_signer.ts
opaolini: Could you expand on this?

I guess we can also just drop this and use `_getTransactionCountAsync` directly?

src/types.ts
opaolini: I have added a routine to mark transactions as `aborted`. Do we want to preserve transactions that have the `dropped` status? Or change their status to `aborted` as well?

src/entities/TransactionEntity.ts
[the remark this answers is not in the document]
fabioberger: Hm, so how do we ensure these entries cannot be stored with empty-string values?

src/runners/transaction_watcher_service_runner.ts
```diff
@@ -0,0 +1,45 @@
...
+
+process.on('unhandledRejection', err => {
+    if (err) {
+        logger.error(err);
+    }
+});
+
+/**
+ * This service tracks transactions and their state changes sent by the meta
+ * transaction relays and persists them to the database.
```
fabioberger, line 31:
```suggestion
 * transaction relays and updates them in the database.
```

src/services/signer_service.ts
```diff
@@ -158,9 +176,18 @@ export class SignerService {
             signature,
             protocolFee,
         );
-
-        const signedEthereumTransaction = await this._privateWalletSubprovider.signTransactionAsync(ethereumTxnParams);
-
+        (PrivateKeyWalletSubprovider as any)._validateTxParams(ethereumTxnParams);
```
fabioberger, line 179: Add `// HACK(oskar): Using private method to validate txParams since not yet exposed by package`

src/services/transaction_watcher_service.ts
```diff
@@ -0,0 +1,116 @@
...
+                { status: TransactionStates.Stuck },
+            ],
+        });
+        logger.trace(`found ${transactionsToCheck.length} transactions to check status`);
+        for (const tx of transactionsToCheck) {
+            await this._findTransactionStatusAndUpdateAsync(tx);
+        }
+    }
+    private async _findTransactionStatusAndUpdateAsync(txEntity: TransactionEntity): Promise<TransactionEntity> {
+        // TODO(oskar) - LessThan does not work on dates in TypeORM queries,
```
fabioberger, line 46:
```suggestion
        // TODO(oskar) - LessThanOrEqual and LessThan do not work on dates in TypeORM queries,
```

src/services/transaction_watcher_service.ts
```diff
@@ -0,0 +1,116 @@
...
+                    txEntity.blockNumber = txInBlockchain.blockNumber;
+                    await this._transactionRepository.save(txEntity);
+                    await this._abortTransactionsWithTheSameNonceAsync(txEntity);
+                    return txEntity;
+                    // Checks if the txn is in the mempool but still has it's status set to Unsubmitted or Submitted
+                } else if (!isExpired && txEntity.status !== TransactionStates.Mempool) {
+                    logger.trace({
+                        message: `a transaction with a ${
+                            txEntity.status
+                        } status is pending, updating status to mempool`,
```
fabioberger, line 70:
```suggestion
                        } status is pending, updating status to TransactionStates.Mempool`,
```

src/services/transaction_watcher_service.ts
```diff
@@ -0,0 +1,116 @@
...
+                    // NOTE(oskar): we currently cancel all transactions that are in the
+                    // "stuck" state. A better solution might be to unstick
+                    // transactions one by one and observing if they unstick the
+                    // subsequent transactions.
+                    txEntity.status = TransactionStates.Stuck;
+                    return this._transactionRepository.save(txEntity);
+                }
+            }
+        } catch (err) {
+            if (err instanceof TypeError) {
```
fabioberger, line 85: Let's add a `// HACK(oskar): web3Wrapper.getTransactionByHashAsync throws a TypeError if the Ethereum node cannot find the transaction and returns NULL instead of the transaction object. We therefore use this to detect this case until @0x/web3-wrapper is fixed.`

src/services/transaction_watcher_service.ts
```diff
@@ -0,0 +1,116 @@
...
+            }
+        }
+        return txEntity;
+    }
+    // Sets the transaction status to 'aborted' for transactions with the same nonce as the passed in txEntity
+    private async _abortTransactionsWithTheSameNonceAsync(txEntity: TransactionEntity): Promise<TransactionEntity[]> {
+        const transactionsToAbort = await this._transactionRepository.find({
+            where: {
+                nonce: txEntity.nonce,
+                hash: Not(txEntity.hash),
```
fabioberger, line 104: Can we include the meta-txn relay address in this query? It'll make it less likely we forget to include it when adding multiple relay address support.

src/services/transaction_watcher_service.ts
```diff
@@ -0,0 +1,116 @@
...
+                // if the transaction has already been marked as dropped, we can skip it
+                status: Not(TransactionStates.Dropped),
```
fabioberger, line 106: Let's overwrite `Dropped` to `Aborted` because we are 100% sure these txns won't ever get mined. That to me, is the fundamental difference between `Dropped` and `Aborted`.

src/types.ts
fabioberger: Change to `aborted`. Elaborated above.

fabioberger: @opaolini mind adding a comment next to each other these explaining them? Some might not be  as self-explanatory.

src/utils/utils.ts
fabioberger: Huh.

test/config.ts
```diff
@@ -0,0 +1,4 @@
+export const TEST_RINKEBY_PUBLIC_ADDRESS = process.env.TEST_RINKEBY_PUBLIC_ADDRESS || '';
+export const TEST_RINKEBY_PRIVATE_KEY = process.env.TEST_RINKEBY_PRIVATE_KEY || '';
+export const TEST_RINKEBY_RPC_URL = process.env.TEST_RINKEBY_RPC_URL || '';
+export const TEST_RINKEBY_CHAIN_ID = 4;
```
fabioberger, line 4: Do we have enough Rinkeby ETH to have this test running on every CI run?

src/services/signer_service.ts
```diff
@@ -173,12 +200,91 @@ export class SignerService {
...
             );
-
+        await this._updateTransactionEntityToSubmittedAsync(txHash);
         return {
             ethereumTransactionHash,
             signedEthereumTransaction,
         };
     }
+    private async _watchForStuckTransactionsAsync(): Promise<void> {
+        while (true) {
```
fabioberger, line 210: Thinking about this again, there is not any way to spin this down. That might not matter in production, but in testing it could be a headache. We should use `@0x/utils`'s `intervalUtils.setAsyncExcludingInterval` method. It has a similar effect to the existing code by disallowing the logic in the loop from executing multiple times in parallel. It is different in that, if it is still executing, it simply skips and waits `STUCK_TX_POLLING_INTERVAL_MS` time again.

The benefit is we get back a timer which can be cancelled on test tear down.

src/services/transaction_watcher_service.ts
```diff
@@ -0,0 +1,116 @@
...
+    private readonly _provider: SupportedProvider;
+    private readonly _web3Wrapper: Web3Wrapper;
+
+    constructor(dbConnection: Connection, provider: SupportedProvider) {
+        this._transactionRepository = dbConnection.getRepository(TransactionEntity);
+        this._provider = provider;
+        this._web3Wrapper = new Web3Wrapper(this._provider);
+    }
+    public async startAsync(): Promise<void> {
+        while (true) {
```
fabioberger, line 21: We comment about using `intervalUtils.setAsyncExcludingInterval` here instead, so we can tear down this job.

test/config.ts
opaolini: yes

src/entities/TransactionEntity.ts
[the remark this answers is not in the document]
opaolini: Well, the best way is to get rid of `synchronize: true` and have a proper migration setup, but that probably is outside of the scope of this PR.

An alternative is to make the `constructor` private and have a `public static` method `create` or `make` which would throw errors on input validation, so instead of:
```
new TransactionEntity({..})
```

it would be:
```
TransacitonEntity.create({})
```

Thoughts?

src/services/transaction_watcher_service.ts
opaolini: Changed the comment from:
```
                // TODO(oskar) `getTransactionByHashAsync` throws a TypeError
                // when tx returned from `eth_getTransactionByHash` is null.
```
To the `HACK` comment.

src/types.ts
opaolini: 👍

src/entities/TransactionEntity.ts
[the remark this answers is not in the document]
fabioberger: I like that as a short-term solution. Let's do that for now.

src/global.d.ts
```diff
@@ -0,0 +1,2 @@
+// TODO(oskar) - prepare type interfaces that we use from this package.
+declare module 'ethereumjs-tx';
```
fabioberger, line 2: @opaolini can you actually flesh this out in this PR? There are plenty of examples in `@0x/typescript-typings`

src/entities/TransactionEntity.ts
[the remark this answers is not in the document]
opaolini: added `public static make`

src/global.d.ts
opaolini:
```
src/global.d.ts:4:16 - error TS2665: Invalid module name in augmentation. Module 'ethereumjs-tx' resolves to an untyped module at '~/projects/0x/0x-api/node_modules/ethereumjs-tx/es5/index.js', which cannot be augmented.
```

And unfortunately I am not entirely sure how to fix this.

test/utils/test_signer.ts
```diff
@@ -0,0 +1,160 @@
...
+import { ETH_TRANSFER_GAS_LIMIT, EXPECTED_MINED_SEC } from '../../src/constants';
+import { TransactionEntity } from '../../src/entities';
+import { logger } from '../../src/logger';
+import { TransactionStates } from '../../src/types';
+import { TEST_RINKEBY_CHAIN_ID } from '../config';
+
+const DEFAULT_TX_VALUE = 1337;
+const HIGH_GAS_PRICE_FOR_UNSTICKING = 13000000000;
+
+export class TestSigner {
```
fragosti, line 24: Hmmm this is a pretty complex mock, does it need to be this complex? why can't we use the real signer?

fragosti:
Oh forgot to mention... should we update the README with these new services?

src/services/signer_service.ts
```diff
@@ -173,12 +214,86 @@ export class SignerService {
...
+            where: { status: TransactionStates.Stuck },
+        });
+        if (stuckTransactions.length === 0) {
+            return;
+        }
+        const gasStationPrice = await this._getGasPriceFromGasStationOrThrowAsync();
+        const targetGasPrice = gasStationPrice.multipliedBy(UNSTICKING_TRANSACTION_GAS_MULTIPLIER);
+        for (const tx of stuckTransactions) {
+            try {
+                await this._unstickTransactionAsync(tx, targetGasPrice.multipliedBy(targetGasPrice));
```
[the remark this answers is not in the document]
opaolini: Oh definitely not, that was a remnant of doing the multiplication inline by the constant multiplier. Fixed!

src/services/signer_service.ts
```diff
@@ -158,9 +189,19 @@ export class SignerService {
             signature,
             protocolFee,
         );
-
-        const signedEthereumTransaction = await this._privateWalletSubprovider.signTransactionAsync(ethereumTxnParams);
-
+        // HACK(oskar): Using private method to validate txParams since it is
+        // not yet exposed by the package.
+        (PrivateKeyWalletSubprovider as any)._validateTxParams(ethereumTxnParams);
```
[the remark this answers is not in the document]
opaolini: True, the `handler` method does the whitelist checks. Renamed!

src/entities/transformers.ts
```diff
@@ -0,0 +1,12 @@
+import { BigNumber } from '@0x/utils';
+import { ValueTransformer } from 'typeorm';
+
+export const BigNumberTransformer: ValueTransformer = {
```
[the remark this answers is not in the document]
opaolini: Perfect, thank you, most of it.

src/services/transaction_watcher_service.ts
```diff
@@ -0,0 +1,117 @@
...
+                { status: TransactionStates.Unsubmitted },
+                { status: TransactionStates.Submitted },
+                { status: TransactionStates.Mempool },
+                { status: TransactionStates.Stuck },
+            ],
+        });
+        logger.trace(`found ${transactionsToCheck.length} transactions to check status`);
+        for (const tx of transactionsToCheck) {
+            await this._findTransactionStatusAndUpdateAsync(tx);
+        }
```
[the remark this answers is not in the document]
opaolini: There is a small benefit to doing it sequentially as we might consider some transactions as aborted, which could be skipped altogether (after marking them as `aborted`).

However we can introduce chunking in the subsequent PR after we introduce multiple relayers, such that we run this in parallel for each of the relayer public addresses.

test/utils/test_signer.ts
opaolini: This `TestSigner` was used during the development to quickly iterate and keep the test simpler.

After we will have the full integration test for meta transactions we might just delete this mock altogether.

opaolini:
@fragosti true.. added the `TransactionWatcher` service to `README.md`, but I guess we should also have a separate PR to:
- update the README for the new `ENV` 's for the `SignerService`
- add `SignerService` runner and relevant entries in `package.json` and add that information to the README

src/services/signer_service.ts
```diff
@@ -173,12 +214,86 @@ export class SignerService {
...
+            metaTxnRelayerAddress: META_TXN_RELAY_ADDRESS,
+            expectedMinedInSec: EXPECTED_MINED_SEC,
+        });
+        await this._transactionEntityRepository.save(transactionEntity);
+        await this._web3Wrapper.sendRawPayloadAsync({
+            method: 'eth_sendRawTransaction',
+            params: [signedEthereumTransaction],
+        });
+        await this._updateTransactionEntityToSubmittedAsync(txHash);
+        return txHash;
```
dekz, line 265: Why doesn't this mark the stuck tx as `Aborted`?  I see `transaction_watcher_service` attempts this by setting all stuck tx with the same nonce to aborted, but that feels like a catch all. Where setting it here would potentially be more targetted?

src/services/transaction_watcher_service.ts
```diff
@@ -0,0 +1,117 @@
...
+    }
+    private async _findTransactionStatusAndUpdateAsync(txEntity: TransactionEntity): Promise<TransactionEntity> {
+        // TODO(oskar) - LessThanOrEqual and LessThan do not work on dates in
+        // TypeORM queries, ref: https://github.com/typeorm/typeorm/issues/3959
+        const now = new Date();
+        const isExpired = txEntity.expectedAt <= now;
```
dekz, line 52: I would suggest `txEntity.expectedAt <= latestBlock.timestamp`

src/services/signer_service.ts
dekz, line 234: Seems like we'd only want to unstick the transaction if its `tx.gasPrice` is < `targetGasPrice`. Otherwise we'd unstick and replace with the same  (or worse) `gasPrice`?

src/services/transaction_watcher_service.ts
```diff
@@ -0,0 +1,117 @@
...
+            }
+        } catch (err) {
+            if (err instanceof TypeError) {
+                // HACK(oskar): web3Wrapper.getTransactionByHashAsync throws a
+                // TypeError if the Ethereum node cannot find the transaction
+                // and returns NULL instead of the transaction object. We
+                // therefore use this to detect this case until @0x/web3-wrapper
+                // is fixed.
+                if (isExpired) {
+                    txEntity.status = TransactionStates.Dropped;
```
dekz, line 91: How certain are we that case is actually `Dropped` and that we're not just talking to a different node behind a load balancer with a different view of the mempool?

> returns NULL instead of the transaction object

I was under the impression this was always geth's behaviour, and parity would return the tx object for unconfirmed transactions. But it's been so long

src/utils/utils.ts
```diff
@@ -71,4 +71,8 @@ export const utils = {
...
+        // tslint:disable-next-line:no-inferred-empty-object-type
+        return new Promise(resolve => setTimeout(resolve, ms));
+    },
```
dekz, line 77:
```suggestion
    delayAsync: async (ms: number): Promise<void> => {
        // tslint:disable-next-line:no-inferred-empty-object-type
        return new Promise<void>(resolve => setTimeout(resolve, ms));
    },
```

test/config.ts
dekz: Should we run this 24/7 legitimately on Kovan? That way we can make actual trades against liquidity.

Or is Kovan too fast to be a good environment?

dekz:
Also FYI, in the last 500,000 blocks there have been 42 cases where the [time between two mined blocks was > 120 seconds](https://explore.duneanalytics.com/queries/3044/source). 180s expected time to mine could still be < 3 blocks.

src/handlers/meta_transaction_handlers.ts
```diff
@@ -158,6 +158,34 @@ export class MetaTransactionHandlers {
...
+        }
+        const tx = await this._metaTransactionService.findTransactionByHashAsync(transactionHash);
+        if (tx === undefined) {
+            throw new NotFoundError();
+        } else {
+            const resp = {
+                hash: tx.hash,
+                status: tx.status,
+                gasPrice: tx.gasPrice,
+                updatedAt: tx.updatedAt,
```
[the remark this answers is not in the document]
opaolini: 👍

src/services/signer_service.ts
```diff
@@ -68,6 +81,24 @@ export class SignerService {
         this._contractWrappers = new ContractWrappers(this._provider, { chainId: CHAIN_ID });
         this._web3Wrapper = new Web3Wrapper(this._provider);
         this._devUtils = new DevUtilsContract(this._contractWrappers.contractAddresses.devUtils, this._provider);
+        this._transactionEntityRepository = dbConnection.getRepository(TransactionEntity);
+        this._privateKeyBuffer = Buffer.from(META_TXN_RELAY_PRIVATE_KEY, 'hex');
+        this._stuckTransactionWatcherTimer = intervalUtils.setAsyncExcludingInterval(
+            async () => {
+                logger.trace('watching for stuck transactions');
+                await this._checkForStuckTransactionsAsync();
```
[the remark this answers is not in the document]
opaolini: Oh, I see now, then `http_service_runner` is not a place for that, as I've put it in the "defaultAppDependencies" if the users would run `0x-API` as a single monolithic app.

The current idea is to have this running as a single instance as it was not specced with idempotency in mind. The best place currently would probably be within a `signer_service_runner` then.

Currently only the `signer_service` holds the private keys, and it does the actual submission of the tx to the network and its also running as a single instance.

If we add multiple relayer support, we could then scale it as multiple instances of `SignerService` if we think we would need to scale it that way.

If single instance would be enough, then we be a bit more smart which private keys to use for example the one with the least amount of pending transactions.

src/services/signer_service.ts
```diff
@@ -173,12 +214,86 @@ export class SignerService {
...
+        try {
+            await this._transactionEntityRepository.manager.transaction(async transactionEntityManager => {
+                const repo = transactionEntityManager.getRepository(TransactionEntity);
+                const txn = await repo.findOne(txHash);
+                if (txn !== undefined && txn.status === TransactionStates.Unsubmitted) {
+                    txn.status = TransactionStates.Submitted;
+                    await transactionEntityManager.save(txn);
+                }
+            });
+            return true;
```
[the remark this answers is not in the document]
opaolini: Yup, we just log if we rolled back. Removed the return values.

src/services/transaction_watcher_service.ts
opaolini: The transaction will only be considered "dropped" when it is expired AND is not present in the mempool according to the RPC provider.

I guess we can never be sure about nodes with a different views of the mempool, only minimize that risk by:
- grabbing the transaction using the same provider we submit it to (which I agree, does not guarantee that the transaction will be reflected in the mempool).
- using multiple RPC providers (alchemy, infura, our self-hosted node) and consider that dropped when all of them agree on that.

Is there a better way?

>  I was under the impression this was always geth's behaviour, and parity would return the tx object for unconfirmed transactions. But it's been so long

That is the case, but the [marshaller](https://github.com/0xProject/0x-monorepo/blob/development/packages/web3-wrapper/src/web3_wrapper.ts#L253) in the web3 wrapper does not actually check if the return is null, and throws a `TypeError`.

src/services/signer_service.ts
opaolini: Currently, we consider transactions `Aborted` ONLY when a transaction with the same nonce is `Confirmed`, whereas here we only know that the transaction was only submitted.

src/services/transaction_watcher_service.ts
opaolini: 👍 changed. We currently store `expectedAt` as `Date` so I went with converting `latestBlock.timestamp` to a date object.

src/services/signer_service.ts
opaolini: Added a check, that skips the transaction where `tx.gasPrice` <= `targetGasPrice` and logs a `warning`.

src/entities/types.ts
```diff
@@ -0,0 +1,11 @@
+import { BigNumber } from '@0x/utils';
+
+export interface TransactionEntityOpts {
+    hash: string;
+    status: string;
+    expectedMinedInSec: number;
+    nonce: number;
+    gasPrice: BigNumber;
+    blockNumber?: number;
+    metaTxnRelayerAddress: string;
```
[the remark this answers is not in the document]
opaolini: Yup, renamed.

src/services/signer_service.ts
```diff
@@ -173,12 +214,86 @@ export class SignerService {
...
+        } catch (err) {
+            // the TransacitonEntity was updated in the meantime. This will
+            // rollback the database transaction.
+            logger.warn('failed to store transaction with submitted status, rolling back', { err });
+        }
+
+        return false;
+    }
+    private _getTransactionHashAndRawTxString(
```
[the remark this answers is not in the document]
opaolini: Changed function name to reflect that the returned values are signed.

opaolini:
deploy staging

opaolini:
deploy staging

opaolini:
deploy staging

PR #195, opened 2020-04-24
fix: Clip Native Order penalty

dekz:
https://github.com/0xProject/0x-monorepo/pull/2565

Native Orders which are perfectly sized to the target amount should not be penalised more than a larger order.


dekz:
deploy staging

dekz:
deploy production

PR #197, opened 2020-04-28
fix: Log 0x api key

steveklebanoff:
image

We feel comfortable having this API key in our logs, and it's very useful to have in Kibana.

### DIFF
--- a/src/middleware/request_logger.ts
+++ b/src/middleware/request_logger.ts
@@ -25,6 +25,7 @@ export function requestLogger(): core.RequestHandler {
                     url: req.originalUrl.split('?')[0],
                     method: req.method,
                     headers: {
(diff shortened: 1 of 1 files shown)


steveklebanoff:
@PirosB3 The value is undefined, and for some reason that causes the `0x-api-key` to not show up at all in the log:

with specified:
image

without:
image

steveklebanoff:
deploy staging

PirosB3:
sounds great! thanks for the follow ip

steveklebanoff:
image

It works ^

steveklebanoff:
deploy production

steveklebanoff:
deploy production

steveklebanoff:
deploy production

PR #198, opened 2020-04-29
fix: Prune Taker fees in Maker asset

dekz:
Asset-swapper cannot currently determine that the taker (if provided) has the ability to pay the taker fee in the maker asset. This is guaranteed by the balance, but is not guaranteed by the allowance of the taker.

Also updates the Kovan ERC20BridgeSampler address. https://github.com/0xProject/0x-monorepo/pull/2570

### DIFF
--- a/package.json
+++ b/package.json
@@ -34,8 +34,8 @@
         "prettier_target": "{.,test/**,src/**}/*.{ts,tsx,json,md}"
     },
     "resolutions": {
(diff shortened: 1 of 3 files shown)


dekz:
deploy staging

dekz:
deploy production

dekz:
deploy production

PR #201, opened 2020-04-29
feat: RFQ-T follow-ups

feuGeneA:
Lots of tasks that needed doing after the initial implementation of RFQ-T. I recommend reviewing commit by commit.

Depends on https://github.com/0xProject/0x-monorepo/pull/2574

-   [x] Test on staging, exercising the new `RFQT_MAKER_ASSET_OFFERINGS` environment variable scheme.
-   [x] Re-pin asset-swapper to monorepo `development`-branch commit, after https://github.com/0xProject/0x-monorepo/pull/2574 merges. (Currently this is pinned to a commit in that PR's branch).

### DIFF
--- a/src/config.ts
+++ b/src/config.ts
@@ -333,6 +333,10 @@ function assertEnvVarType(name: string, value: any, expectedType: EnvVarType): a
                         `second token address for asset pair ${i} for maker endpoint ${makerEndpoint}`,
                         assetPair[1],
                     );
+                    assert.assert(
+                        assetPair[0] !== assetPair[1],
+                        `asset pair array ${i} for maker endpoint ${makerEndpoint} has identical assets`,
+                    );
                 });
             }
             return offerings;
--- a/src/config.ts
+++ b/src/config.ts
@@ -1,6 +1,6 @@
 // tslint:disable:custom-no-magic-numbers
 import { assert } from '@0x/assert';
-import { ERC20BridgeSource, SwapQuoteRequestOpts } from '@0x/asset-swapper';
+import { ERC20BridgeSource, RfqtMakerAssetOfferings, SwapQuoteRequestOpts } from '@0x/asset-swapper';
 import { BigNumber } from '@0x/utils';
 import * as _ from 'lodash';
 import * as validateUUID from 'uuid-validate';
@@ -30,6 +30,7 @@ enum EnvVarType {
     FeeAssetData,
     NonEmptyString,
     APIKeys,
+    RfqtMakerAssetOfferings,
 }
 
 // Network port to listen on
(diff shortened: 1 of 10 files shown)


src/config.ts
```diff
@@ -149,11 +151,22 @@ export const LIQUIDITY_POOL_REGISTRY_ADDRESS: string | undefined = _.isEmpty(
           EnvVarType.ETHAddressHex,
       );
 
-export const RFQT_API_KEY_WHITELIST: string[] =
-    process.env.RFQT_API_KEY_WHITELIST === undefined ? [] : process.env.RFQT_API_KEY_WHITELIST.split(',');
+export const RFQT_API_KEY_WHITELIST: string[] = _.isEmpty(process.env.RFQT_API_KEY_WHITELIST)
+    ? []
+    : assertEnvVarType('RFQT_API_KEY_WHITELIST', process.env.RFQT_API_KEY_WHITELIST, EnvVarType.StringList);
+
+export const RFQT_MAKER_ASSET_OFFERINGS: RfqtMakerAssetOfferings = _.isEmpty(process.env.RFQT_MAKER_ASSET_OFFERINGS)
```
dekz, line 158: Just an opinion, not a blocker.

But this feels backwards to me, we'd need to redeploy and have constant communication to list and delist certain pairs as RFQT providers come online.

I'd prefer to see the MM advertise what they support and we periodically fetch that inside AS. Like the SRA asset_pairs endpoint.

src/config.ts
```diff
@@ -297,6 +310,33 @@ function assertEnvVarType(name: string, value: any, expectedType: EnvVarType): a
...
+                        `asset pair array ${i} for maker endpoint ${makerEndpoint} does not consist of exactly two elements.`,
+                    );
+                    assert.isETHAddressHex(
+                        `first token address for asset pair ${i} for maker endpoint ${makerEndpoint}`,
+                        assetPair[0],
+                    );
+                    assert.isETHAddressHex(
+                        `second token address for asset pair ${i} for maker endpoint ${makerEndpoint}`,
+                        assetPair[1],
+                    );
```
dekz, line 335: perhaps assert that `a != b` incase I mess up the config down the road :P

src/handlers/swap_handlers.ts
```diff
@@ -51,7 +51,18 @@ export class SwapHandlers {
...
+        logger.info({
+            indicativeQuoteServed: {
+                taker: params.takerAddress,
+                apiKey: params.apiKey,
+                buyToken: params.buyToken,
+                sellToken: params.sellToken,
+                buyAmount: params.buyAmount,
+                sellAmount: params.sellAmount,
+                makers: orders.map(o => o.makerAddress),
+            },
```
dekz, line 64: this logging looks brief so LGTM

src/config.ts
steveklebanoff: We will def want to do something like that in a future revision 👍 

For now, MMers are not updating their pairs too often, so this is 'good enough' for a first approach.  The most important thing for now is that we don't waste precious resources making a query for a pair we know the MMer doesn't have.

steveklebanoff:
@feuGeneA Can you add a checkbox to make sure you update our infra to use this new ENV var style?

feuGeneA:
deploy staging

feuGeneA:
Kibana is parsing the logs just fine:
image

feuGeneA:
I verified (manually, by watching staging 0x API in Kibana) that the RFQ-T stuff is working with the new environment variable changes in place.

PR #202, opened 2020-04-30
feat: Added pool breakdown of operator vs member stake, added option to add…

alexkroeger:
… fees to epoch

Changes:
- Added an option to add fees to the epoch endpoint. Fees make the amount of data queried much larger, but they are useful for estimating rewards for an epoch.
- Added a breakdown of stake for the /pools endpoint between members and operator.

Testing:
- [x] Tested locally on Hashalytics

Dependencies:
https://github.com/0xProject/0x-event-pipeline/pull/24

### DIFF
--- a/src/handlers/staking_handlers.ts
+++ b/src/handlers/staking_handlers.ts
@@ -46,7 +46,7 @@ export class StakingHandlers {
             if (!(_req.query.withFees === 'true' || _req.query.withFees === 'false')) {
                 res.status(HttpStatus.BAD_REQUEST).send(`Invalid value for withFees`);
             }
-            isWithFees = (_req.query.withFees === 'true');
+            isWithFees = _req.query.withFees === 'true';
         } else {
             isWithFees = false;
         }
@@ -63,8 +63,8 @@ export class StakingHandlers {
             };
         } else {
             const [currentEpoch, nextEpoch] = await Promise.all([
-            this._stakingDataService.getCurrentEpochAsync(),
-            this._stakingDataService.getNextEpochAsync(),
+                this._stakingDataService.getCurrentEpochAsync(),
+                this._stakingDataService.getNextEpochAsync(),
             ]);
             response = {
                 currentEpoch,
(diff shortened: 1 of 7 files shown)


src/handlers/staking_handlers.ts
```diff
@@ -39,14 +40,37 @@ export class StakingHandlers {
...
-        const response: StakingEpochsResponse = {
-            currentEpoch,
-            nextEpoch,
-        };
+        // optional query string to include fees
+        let isWithFees: boolean;
+        if (_req.query.withFees) {
+            if (!(_req.query.withFees === 'true' || _req.query.withFees === 'false')) {
+                res.status(HttpStatus.BAD_REQUEST).send(`Invalid value for withFees`);
+            }
```
fragosti, line 48: The reason `_req` had the underscore is that it wasn't being used. Now that it is being used, we need to remove the underscore.

src/handlers/staking_handlers.ts
```diff
@@ -39,14 +40,37 @@ export class StakingHandlers {
...
+            isWithFees = _req.query.withFees === 'true';
+        } else {
+            isWithFees = false;
+        }
```
fragosti, line 52: Elsewhere we extract query params in pure functions (so we can actually type the query object)

src/services/staking_data_service.ts
```diff
@@ -41,15 +43,35 @@ export class StakingDataService {
...
+        );
+        if (!rawEpochWithFees) {
+            throw new Error('Could not find the current epoch.');
+        }
+        const epoch = stakingUtils.getEpochWithFeesFromRaw(rawEpochWithFees);
+        return epoch;
+    }
+
     public async getNextEpochAsync(): Promise<Epoch> {
         const rawEpoch: RawEpoch | undefined = _.head(await this._connection.query(queries.nextEpochQuery));
```
fragosti, line 58: Are we now over-fetching information for this query?

src/handlers/staking_handlers.ts
alexkroeger: 👍

src/handlers/staking_handlers.ts
alexkroeger: 👍

src/services/staking_data_service.ts
alexkroeger: nah, it runs a separate query that doesn't SUM over fills

alexkroeger:
deploy staging

PR #208, opened 2020-05-01
fix: sixtySecondsFromNow timestamp

opaolini:


src/services/signer_service.ts
```diff
@@ -75,7 +75,7 @@ export class SignerService {
     ): Promise<BigNumber> {
         // Verify 0x txn won't expire in next 60 seconds
         // tslint:disable-next-line:custom-no-magic-numbers
-        const sixtySecondsFromNow = new BigNumber(+new Date() + 60);
+        const sixtySecondsFromNow = new BigNumber(new Date().getTime() / ONE_SECOND_MS + 60);
```
kimpers, line 78: I just realized that we probably want to floor it as well otherwise we end up with fractions of seconds
```suggestion
        const sixtySecondsFromNow = new BigNumber(Math.floor(new Date().getTime() / ONE_SECOND_MS) + 60);
```

PR #213, opened 2020-05-05
test: meta transaction tests

jalextowle:
# Description

* Introduce meta transaction tests
* split up tests into separate suites with their own setup/teardown flows
* move CI from CircleCI to Github Actions. After merging we have to manually remove CircleCI from Settings > Webhooks in this repo.

The SRA and Swap test suite files could use some polishing, but since there are PRs out for both those suites, this PR makes minimal changes there. 
# Testing Instructions

CI on Github Actions! ✨ 

# Checklist


test/meta_transaction_test.ts
```diff
@@ -0,0 +1,742 @@
...
+                    .toString('hex')}`;
+            }
+
+            describe('single order submission', () => {
+                let validationResults: ValidationResults;
+                const price = '1';
+                const sellAmount = calculateSellAmount(buyAmount, price);
+
+                // NOTE(jalextowle): This must be a `before` hook because `beforeEach`
+                // hooks execute after all of the `before` hooks (even if they are nested).
```
xianny, line 522: I noticed the earlier test cases use `beforeEach`. Seems like setting up mesh and creating a blockchain checkpoint *after* all the `before` hooks would be the desired behaviour? 

Couple qs, 1) what's different about this case vs earlier cases; and 2) do these blocks execute in scope order right now? e.g. top level scope `before`, then nested `before`?

test/meta_transaction_test.ts
```diff
@@ -0,0 +1,742 @@
...
+
+            afterEach(async () => {
+                await blockchainLifecycle.revertAsync();
+                await teardownMeshAsync(SUITE_NAME);
+            });
+
+            // NOTE(jalextowle): Spin up a new Mesh instance so that it will
+            // be available for future test suites.
+            after(async () => {
+                await setupMeshAsync(SUITE_NAME);
```
xianny, line 231: why do we setup mesh again, when every suite has a `beforeEach` that sets up mesh?

.circleci/config.yml
```diff
@@ -46,4 +23,5 @@ jobs:
             - run: yarn lint
             - run: yarn prettier:ci
             - setup_remote_docker # Expose the docker daemon so that it can be used in tests.
-            - run: yarn test
+            - run: yarn test:rfqt
+            - run: yarn test:rest
```
xianny, line 27: why split these?

jalextowle: These must be split because we can't cleanly tear down `app` right now. This will cause it to try to reconnect to Mesh and cause a whole host of other issues. The easiest way to fix it was to separate the tests so that there is only a single suite with an `app` instance. For all future suites, we should use the new flows, but from speaking to the maker team, this may be somewhat difficult for the RFQT tests in the near term.

test/meta_transaction_test.ts
jalextowle: This is supposed to test calling `/price` then `/quote` then `/submit`. If we use the `beforeEach` pattern, then this would revert all of the progress that had been made during previous tests and wouldn't really be as much of an `end-to-end test`.

test/meta_transaction_test.ts
```diff
@@ -0,0 +1,734 @@
...
+            price: testCase.expectedPrice,
+            zeroExTransaction: {
+                signerAddress: takerAddress,
+                domain: { chainId, verifyingContract: contractAddresses.exchange },
+            },
+            orders: testCase.expectedOrders.map(order => stringifyOrderBigNumbers(order)),
+            buyAmount: testCase.expectedBuyAmount,
+            sellAmount: calculateSellAmount(testCase.expectedBuyAmount, testCase.expectedPrice),
+            // NOTE(jalextowle): 0x is the only source that is currently being tested.
+            sources: [
```
dekz, line 369: suggestion: pull this out and `expect(test.quote.sources) to include` so when we add a new source we don't have to update the sources list with a proportion of 0.

test/meta_transaction_test.ts
```diff
@@ -0,0 +1,734 @@
...
+                    const endTakerWethBalance = await weth.balanceOf(takerAddress).callAsync();
+                    const endTakerZrxBalance = await zrx.balanceOf(takerAddress).callAsync();
+                    expect(endMakerWethBalance).to.be.bignumber.eq(startMakerWethBalance.plus(sellAmount));
+                    expect(endMakerZrxBalance).to.be.bignumber.eq(startMakerZrxBalance.minus(buyAmount));
+                    expect(endTakerWethBalance).to.be.bignumber.eq(startTakerWethBalance.minus(sellAmount));
+                    expect(endTakerZrxBalance).to.be.bignumber.eq(startTakerZrxBalance.plus(buyAmount));
+                });
+            });
+
+            // TODO: There is a problem with this test case. It is currently throwing an `IncompleteFillError`
```
dekz, line 617: We will be reworking these tests in the next meta-txn pr, so happy for these to be skipped for now.

test/rfqt_test.ts
```diff
@@ -0,0 +1,531 @@
...
+                            responseData: rfqtIndicativeQuoteResponse,
+                            responseCode: 200,
+                            requestApiKey: 'koolApiKey1',
+                            requestParams: mockedApiParams,
+                        },
+                    ],
+                    async () => {
+                        const appResponse = await request(app)
+                            .get(
+                                `${SWAP_PATH}/quote?buyToken=ZRX&sellToken=WETH&sellAmount=${sellAmount.toString()}&takerAddress=${takerAddress}&intentOnFilling=false&excludedSources=Uniswap,Eth2Dai,Kyber,LiquidityProvider&skipValidation=true`,
```
dekz, line 155: nit: `excludedSources=${EXCLUDED_SOURCES}` so we don't have to adjust this many times.

test/rfqt_test.ts
```diff
@@ -0,0 +1,531 @@
...
+                    sellToken: contractAddresses.etherToken,
+                    buyToken: contractAddresses.zrxToken,
+                    sellAmount: sellAmount.toString(),
+                    buyAmount: undefined,
+                    takerAddress,
+                };
+                return rfqtMocker.withMockedRfqtFirmQuotes(
+                    [
+                        {
+                            endpoint: 'https://mock-rfqt1.club',
```
dekz, line 104: nit: pull this up. DRY

```
  endpoint: 'https://mock-rfqt1.club',
  responseCode: 200,
  requestApiKey: 'koolApiKey1',
  requestParams: mockedApiParams,
```
```
   ...defaults,
  responseData: ganacheZrxWethOrder1,
```

test/rfqt_test.ts
```diff
@@ -0,0 +1,531 @@
...
+                        delete responseJson.gas;
+                        delete responseJson.gasPrice;
+                        delete responseJson.protocolFee;
+                        delete responseJson.value;
+                        expect(responseJson).to.eql({
+                            buyAmount: '100000000000000000',
+                            price: '1',
+                            sellAmount: '100000000000000000',
+                            sources: [
+                                {
```
dekz, line 399: nit: same comment as above, bit if a PITA if we add a new source and have to update all of these fields.

test/utils/deployment.ts
```diff
@@ -168,14 +228,38 @@ async function waitForApiStartupAsync(logStream: ChildProcessWithoutNullStreams)
         logStream.stdout.on('data', (chunk: Buffer) => {
             const data = chunk.toString().split('\n');
             for (const datum of data) {
-                if (/API (HTTP) listening on port 3000!/.test(datum)) {
+                if (/API \(HTTP\) listening on port 3000!/.test(datum)) {
```
dekz, line 231: suggestion (non-blocking): I'd rather sleep then rely on  log output 😆. Or check for port recv

src/services/websocket_service.ts
```diff
@@ -144,7 +144,8 @@ export class WebsocketService {
         this._requestIdToSubscriptionOpts.clear();
         this._server.close();
         if (this._meshSubscriptionId) {
-            void this._meshClient.unsubscribeAsync(this._meshSubscriptionId);
+            await this._meshClient.unsubscribeAsync(this._meshSubscriptionId);
```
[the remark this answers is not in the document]
opaolini: The underlying `send` in `web3-provider` can throw [0], wrapped it in `try/finally`

[0] - https://github.com/ethereum/web3.js/blob/2.x/packages/web3-providers/src/providers/WebsocketProvider.js#L219

src/utils/token_metadata_utils.ts
```diff
@@ -59,12 +59,14 @@ export function isWETHSymbolOrAddress(tokenAddressOrSymbol: string, chainId: num
  * @param chainId the Network where the address should be hosted on.
  */
 export function findTokenAddress(symbolOrAddress: string, chainId: ChainId): string {
-    if (symbolOrAddress.startsWith('0x') && symbolOrAddress.length === ADDRESS_HEX_LENGTH) {
-        return symbolOrAddress;
+    const normalizedSymbolOrAddress = symbolOrAddress.toLowerCase();
+    if (normalizedSymbolOrAddress.startsWith('0x') && normalizedSymbolOrAddress.length === ADDRESS_HEX_LENGTH) {
```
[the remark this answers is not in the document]
opaolini: It does check that, left the error throwing intact. Changed the function signature to  be `findTokenAddressOrThrow`.

opaolini:
Resolved comments by @dekz

test/rfqt_test.ts
```diff
@@ -488,34 +361,22 @@ describe(SUITE_NAME, () => {
...
-                        responseCode: 200,
-                        requestApiKey: 'koolApiKey1',
-                        requestParams: mockedApiParams,
                     },
                 ],
                 async () => {
                     const appResponse = await request(app)
                         .get(
-                            `${SWAP_PATH}/quote?buyToken=ZRX&sellToken=WETH&sellAmount=${sellAmount.toString()}&takerAddress=${takerAddress}&intentOnFilling=true&excludedSources=Uniswap,Eth2Dai,Kyber,LiquidityProvider&skipValidation=true`,
+                            `${SWAP_PATH}/quote?buyToken=ZRX&sellToken=WETH&sellAmount=${DEFAULT_RFQT_RESPONSE_DATA.toString()}&takerAddress=${takerAddress}&intentOnFilling=true&excludedSources=${DEFAULT_EXCLUDED_SOURCES}&skipValidation=true`,
```
xianny, line 379: `sellAmount=${DEFAULT_RFQT_RESPONSE_DATA.toString()}` should be `sellAmount=${DEFAULT_SELL_AMOUNT}`

opaolini: 🤦 good catch, thanks

PR #218, opened 2020-05-08
fix: Add RFQT configs when instantiating SwapQuoter

fabioberger:


dekz:
deploy staging

dekz:
deploy production

src/services/meta_transaction_service.ts
```diff
@@ -1,4 +1,5 @@
 import { Orderbook, SwapQuoter, SwapQuoterOpts } from '@0x/asset-swapper';
+import { OrderPrunerPermittedFeeTypes } from '@0x/asset-swapper/lib/src/types';
```
steveklebanoff, line 2: nit but I always see us importing directly from `@0x/asset-swapper`, not any sub-dirs.   I think we should modify asset-swapper to export this from `index.ts` and then import it from `@0x/asset-swapper`

PR #224, opened 2020-05-14
refactor: Transaction Signing improvements

opaolini:
# Description

This PR:
- fixes an issue where the signer would not sign and broadcast subsequent transactions if it failed to sign and broadcast one of them.
- adds a `cancelled` (naming?) state - a transaction is considered cancelled when the signer failed to sign and broadcast it in a given time (currently `100 s` which might be an overkill)

# Testing Instructions

# Checklist

### DIFF
--- a/src/services/transaction_watcher_signer_service.ts
+++ b/src/services/transaction_watcher_signer_service.ts
@@ -8,9 +8,9 @@ import { ETHEREUM_RPC_URL, META_TXN_RELAY_EXPECTED_MINED_SEC, META_TXN_RELAY_PRI
 import {
     NUMBER_OF_BLOCKS_UNTIL_CONFIRMED,
     ONE_SECOND_MS,
+    TX_HASH_RESPONSE_WAIT_TIME_MS,
     TX_WATCHER_POLLING_INTERVAL_MS,
     UNSTICKING_TRANSACTION_GAS_MULTIPLIER,
-    TX_HASH_RESPONSE_WAIT_TIME_MS,
 } from '../constants';
 import { TransactionEntity } from '../entities';
 import { logger } from '../logger';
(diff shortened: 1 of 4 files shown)


src/services/transaction_watcher_signer_service.ts
```diff
@@ -192,14 +198,33 @@ export class TransactionWatcherSignerService {
...
         for (const tx of unsignedTransactions) {
-            const signer = await this._getNextSignerAsync();
-            await this._signAndBroadcastMetaTxAsync(tx, signer);
+            if (TransactionWatcherSignerService._isUnsubmittedTxExpired(tx)) {
+                logger.error({
+                    message: `found a transaction in an unsubmitted state waiting longer that ${TX_HASH_RESPONSE_WAIT_TIME_MS}ms`,
+                    refHash: tx.refHash,
+                    from: tx.from,
+                });
+                tx.status = TransactionStates.Cancelled;
```
dekz, line 213: non-blocking: do you think aborted makes sense here? Where aborted is we either kicked it out or it was way expired? Not sure if we need the granularity and they both feel `Aborted/Cancelled`.

src/utils/signer.ts
```diff
@@ -72,6 +73,12 @@ export class Signer {
             signature,
             protocolFee,
         );
+        logger.info({
+            message: `attempting to sign and broadcast a meta transaction`,
+            nonce: ethereumTxnParams.nonce,
```
dekz, line 78: Would be better to convert these from hex values otherwise we'll have to manually do it when looking at the logs

src/services/transaction_watcher_signer_service.ts
opaolini: The distinction would help to quickly identify the transactions that were simply unsigned by the `TransactionWatcherSigner` because of error / low balance / transaction rejection etc.

`Aborted` transactions currently refer specifically to transactions which were superseded by a transaction with a higher gas price and the same nonce.

PR #228, opened 2020-05-18
fix: unify the response data from /swap/v0/price and /meta_transaction/v0/price

### LINKED ISSUE #227: Different response interface for /swap/v0/price and /meta_transaction/v0/price
It would be nice if these 2 endpoints had a similar interface to the extent it makes sense. That way the client can expect certain fields to always be available regardless of using the `swap` or `meta_transaction`.

Current endpoint responses:
GET /swap/v0/price
```json
{
  "price": "0.002489",
  "value": "11728833591897130",
  "gasPrice": "33000000000",
  "gas": "540000",
  "protocolFee": "4950000000000000",
  "buyAmount": "1000000000000000000",
  "sellAmount": "2489000000000000",
  "sources": [...]
}
```
GET `/meta_transaction/v0/price`
```json
{
  "price": "0.002489",
  "buyAmount": "1000000000000000000",
  "sellTokenAddress": "0xc02aaa39b223fe8d0a0e5c4f27ead9083c756cc2",
  "buyTokenAddress": "0xe41d2489571d322189246dafa5ebde1f4699f498"
}
```

kimpers:
# Description
This fixes #227.

When integrating with the `/price` endpoints I noticed that the data returned back is very sparse compared to the quote endpoint. I also noticed that two endpoints return different data, so I tried to unify to the point it makes sense.

For us, it would be very helpful to be able to derive the pair from the response data so we don't have to keep track of that locally in the global state.

# Testing Instructions
**TODO:**
- [x] Test through on staging once the format is settled

# Checklist
-   [X] Update [documentation](https://github.com/0xProject/website/blob/development/mdx/api/index.mdx) as needed. **Website Documentation PR:** 0xProject/website#298
-   [X] Prefix PR title with `[WIP]` if necessary.
-   [X] Add tests to cover changes as needed.

### DIFF
--- a/src/services/meta_transaction_service.ts
+++ b/src/services/meta_transaction_service.ts
@@ -138,8 +138,8 @@ export class MetaTransactionService {
 
         const response: CalculateMetaTransactionPriceResponse = {
             takerAddress,
-            sellAmount,
-            buyAmount,
+            buyAmount: makerAssetAmount,
+            sellAmount: totalTakerAssetAmount,
             price,
             swapQuote,
             sources: serviceUtils.convertSourceBreakdownToArray(swapQuote.sourceBreakdown),
--- a/src/types.ts
+++ b/src/types.ts
@@ -369,6 +369,22 @@ export interface Price {
     price: BigNumber;
 }
 
+interface BasePriceResponse {
+    price: BigNumber;
+    buyAmount: BigNumber;
+    sellAmount: BigNumber;
+    sellTokenAddress: string;
+    buyTokenAddress: string;
+    sources: GetSwapQuoteResponseLiquiditySource[];
+}
+
+export interface GetSwapQuotePriceResponse extends BasePriceResponse {
+    value: BigNumber;
+    gasPrice: BigNumber;
+    gas: BigNumber;
+    protocolFee: BigNumber;
+}
+
 export type GetTokenPricesResponse = Price[];
 
 export interface GetMetaTransactionQuoteResponse {
@@ -381,11 +397,7 @@ export interface GetMetaTransactionQuoteResponse {
     sources: GetSwapQuoteResponseLiquiditySource[];
 }
 
-export interface GetMetaTransactionPriceResponse {
-    price: BigNumber;
-    buyAmount: BigNumber;
-    sellAmount: BigNumber;
-}
+export interface GetMetaTransactionPriceResponse extends BasePriceResponse {}
 
 // takerAddress, sellAmount, buyAmount, swapQuote, price
 export interface CalculateMetaTransactionPriceResponse {
--- a/src/types.ts
+++ b/src/types.ts
@@ -378,7 +378,7 @@ interface BasePriceResponse {
     sources: GetSwapQuoteResponseLiquiditySource[];
 }
 
-export interface GetSwapQuotePriceResponse extends BasePriceResponse {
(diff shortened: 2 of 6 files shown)


src/types.ts
```diff
@@ -369,6 +369,22 @@ export interface Price {
     price: BigNumber;
 }
 
+interface BasePriceResponse {
+    price: BigNumber;
+    buyAmount: BigNumber;
+    sellAmount: BigNumber;
+    sellTokenAddress: string;
```
kimpers, line 376: This uses the token address to be in line with the quote endpoint. However, the prices endpoint returns symbols.

Either symbol or token address would work as long as it's consistent.

fragosti:
@kimpers can you create a PR in the documentation as mentioned in https://github.com/0xProject/0x-api/blob/master/PULL_REQUEST_TEMPLATE.md

src/types.ts
```diff
@@ -369,6 +369,22 @@ export interface Price {
...
+    buyTokenAddress: string;
+    sources: GetSwapQuoteResponseLiquiditySource[];
+}
+
+export interface GetSwapQuotePriceResponse extends BasePriceResponse {
```
feuGeneA, line 381: Can we rename this to just `GetSwapPriceResponse`?  Since `SwapQuote` and `SwapPrice` are two distinct things, it feels a little weird to see `...SwapQuotePrice...`.

src/types.ts
dekz: @kimpers are you referring to https://api.0x.org/swap/v0/prices ?

kimpers: Yeah exactly.

It would be nice to have consistency across the board, if we go with token address here maybe we should add it there as well or vice versa.

src/services/meta_transaction_service.ts
```diff
@@ -138,8 +138,8 @@ export class MetaTransactionService {
 
         const response: CalculateMetaTransactionPriceResponse = {
             takerAddress,
-            sellAmount,
-            buyAmount,
+            buyAmount: makerAssetAmount,
+            sellAmount: totalTakerAssetAmount,
```
kimpers, line 142: Noticed that the metatx endpoint would only return the argument you pass to it, whereas the swap endpoint always returns `buyAmount` and `sellAmount`.

kimpers:
deploy staging

kimpers:
> @kimpers can you create a PR in the documentation as mentioned in https://github.com/0xProject/0x-api/blob/master/PULL_REQUEST_TEMPLATE.md

Implemented in 0xProject/website#298.

Additionally, the build is now failing due to a missing CHANGELOG entry. Should I manually add that to the file or what is the process for this?

fragosti:
@kimpers yes sorry about the changelog. Can you modify the prettier glob pattern in package.json to just ignore the changelog? Humans don't touch it anyway.

kimpers:
> @kimpers yes sorry about the changelog. Can you modify the prettier glob pattern in package.json to just ignore the changelog? Humans don't touch it anyway.

Fixed 👍

fragosti:
@kimpers oh I just did https://github.com/0xProject/0x-api/pull/234

kimpers:
> @kimpers oh I just did #234

I see 👍 , dropped the commit now 😄

src/types.ts
fragosti: I agree it should be consistently address since we cannot guarantee symbol.

PR #230, opened 2020-05-18
feat: added an epochs/n endpoint to get info on an arbitrary epoch

alexkroeger:
# Description

Added an endpoint to supply an arbitrary epoch number and obtain epoch info.

# Testing Instructions

- [x] Tested locally on analytics DB

### DIFF
--- a/src/handlers/staking_handlers.ts
+++ b/src/handlers/staking_handlers.ts
@@ -52,10 +52,10 @@ export class StakingHandlers {
 
         let response: Epoch | EpochWithFees;
         if (isWithFees) {
-            const [epoch] = await Promise.all([this._stakingDataService.getEpochNWithFeesAsync(n)]);
+            const epoch = await this._stakingDataService.getEpochNWithFeesAsync(n);
             response = epoch;
         } else {
-            const [epoch] = await Promise.all([this._stakingDataService.getEpochNAsync(n)]);
+            const epoch = await this._stakingDataService.getEpochNAsync(n);
             response = epoch;
         }
         res.status(HttpStatus.OK).send(response);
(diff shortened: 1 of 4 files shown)


alexkroeger:
deploy staging

src/handlers/staking_handlers.ts
```diff
@@ -41,6 +43,23 @@ export class StakingHandlers {
...
+    public async getStakingEpochNAsync(req: express.Request, res: express.Response): Promise<void> {
+        // optional query string to include fees
+        schemaUtils.validateSchema(req.query, schemas.stakingEpochRequestSchema as any);
+        const isWithFees = req.query.withFees ? req.query.withFees === 'true' : false;
+
+        const n = Number(req.params.n);
+
+        let response: Epoch | EpochWithFees;
+        if (isWithFees) {
+            const [epoch] = await Promise.all([this._stakingDataService.getEpochNWithFeesAsync(n)]);
```
kimpers, line 55: NIT: The `Promise.all` kinda superfluous since you're are only awaiting 1 promise.
```suggestion
            const epoch = await this._stakingDataService.getEpochNWithFeesAsync(n);
```

src/handlers/staking_handlers.ts
```diff
@@ -41,6 +43,23 @@ export class StakingHandlers {
...
+            response = epoch;
+        } else {
+            const [epoch] = await Promise.all([this._stakingDataService.getEpochNAsync(n)]);
```
kimpers, line 58: NIT:
```suggestion
            const epoch = await this._stakingDataService.getEpochNAsync(n);
```

PR #235, opened 2020-05-19
feat: added rewards per ZRX to pools endpoint

alexkroeger:
# Description

Added a new field to the `pools` and `pools/<id>` endpoint called `avgMemberRewardEthPerZrx`. It provides the average member reward over the last 3 epochs divided by the average member stake. 

# Testing Instructions

Testing:
- [x] Tested locally on analytics database


alexkroeger:
good to go, test failing bc of prettier due to changelog

PR #240, opened 2020-05-22
fix: Re-pin asset-swapper to latest monorepo commit

feuGeneA:
To pick up a small fix for logging RFQ-T maker endpoint response error codes. 

This is the monorepo commit being pinned to: https://github.com/0xProject/0x-monorepo/commit/98a99d96aab15623188661568785b5354e66172a

At the time of this writing (9am PT, May 28) that's the latest monorepo commit.

This change has been verified on staging.  However, I did not verify the actual case being corrected (the error response code handling from the maker endpoint), because it was not readily apparent to me how to simulate that.  But what I _did_ test was that existing functionality is holding together okay.  To do that I requested a few RFQ-T quotes, and also ran the Assertible smoke tests.

### DIFF
--- a/package.json
+++ b/package.json
@@ -124,7 +124,7 @@
     },
     "dependencies": {
         "@0x/assert": "^3.0.4",
-        "@0x/asset-swapper": "0xProject/gitpkg-registry#0x-asset-swapper-v4.4.0-511cd90c4",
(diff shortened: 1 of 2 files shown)


feuGeneA:
deploy staging

PR #248, opened 2020-06-08
feat: add transaction price information to meta-txn endpoints for UI

fragosti:
# Description

We want to communicate exactly how much money we're saving users in the meta-txn endpoint too. This is a bit different than the regular endpoint:
- We don't need to tell them about the refund they would potentially be getting with gas tokens.
- The estimates don't need to be as accurate.
- We can use `protocolFee` instead of `minimumProtocolFee`.

# Testing Instructions

# Checklist


dekz:
deploy staging

dekz:
deploy staging

dekz:
deploy staging

PR #260, opened 2020-06-17
fix: validation gas limit

dekz:
# Description

Geth shipped a breaking change which did the following.

1.  Removed the "bug" which allowed the caller to have `value` irrespective of their ETH balance
2. Set the gas limit to MaxUint64 if not supplied

These two combination results in an insufficient funds as not even Vitalik has enough funds for 50 gwei @ MaxUint64 gas limit.

As a result we either need to drop (or lower) gas price to 1 wei, resulting in the user requiring 18 ETH for validation. We want to validate the gas price and protocol fee, so we cannot drop the gas price entirely. So this is out.

So we need to either fix the gas limit to something reasonable in the case we're performing a validation (eth call with ?? gas). Or we perform the Estimate gas first then the `eth_call`, unable to parallelize.

In `/swap/v0/quote` we have opted to remove the parallelization and first perform the estimate then the call. 

In MetaTxn we have opted to fix the gas limit to 10e6 as we do not yet have a signer. The reason we wanted to perform them in parallel is for speed and to reveal the `eth_call` error, not the gas estimation error.

### DIFF
--- a/src/constants.ts
+++ b/src/constants.ts
@@ -17,6 +17,7 @@ export const DEFAULT_LOGGER_INCLUDE_TIMESTAMP = true;
 export const ONE_SECOND_MS = 1000;
 export const ONE_MINUTE_MS = ONE_SECOND_MS * 60;
 export const TEN_MINUTES_MS = ONE_MINUTE_MS * 10;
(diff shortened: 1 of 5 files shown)


dekz:
deploy staging

dekz:
deploy production

PR #262, opened 2020-06-18
feat: support ExchangeProxy for swap quotes (Swap V1)

xianny:
# Description

Introduces `swap/v1` and preserves old functionality in `swap/v0`.

Changes:
* Selling of ERC20 and receiving ETH in v1 is no longer an error but is supported with parameters (isToEth) to the ExchangeProxy.
* SwapQuoteConsumer exports call data for ExchangeProxy
* Support RFQT via EP (specify taker address as EP FlashWallet)
* Support ETH output and input (using EP parameters), removing any Forwarder references
* Add allowance target to response
* disabled `shouldBatchBridgeOrders` (waiting for "on-chain fills without protocol fees" to ship)

**Note on versioning**
This PR adds paths for V0 and V1 functionality within the code. This approach is less than ideal as code complexity can quickly grow unmanageable. We're doing this as a one-off because V1 does not have significant changes compared to V0. We'll need to revisit our approach for future major version bumps.

Additional context: Technically V1 does not have breaking interface changes from V0 and might not warrant a major version bump. However, one of our integrators who rely on hardcoded contract addresses or users not expecting a new allowance target will have a badd experience. We'll be running V0 and V1 concurrently to support them. Discussion is still open as to how we will approach similar situations in the future.

# Testing Instructions

Existing unit tests have been copied and moved `swap_test.ts` -> `swap_v0_test.ts` with no meaningful changes. These should still pass.

`swap_test.ts` has the same tests as before, with one difference. When `sellToken=ETH` it should return the exchange proxy address instead of the forwarder address as the `to` field. 

- [x] [Protocol Fee is 0](https://staging.api.0x.org/swap/v1/quote?buyToken=DAI&sellToken=ETH&sellAmount=300000000000000000000&excludedSources=0x) when only on-chain sources are used. Value is also 0 when non ETH sell.
- [x] [On-chain sources are no longer rolled up](https://staging.api.0x.org/swap/v1/quote?buyToken=DAI&sellToken=ETH&sellAmount=300000000000000000000&excludedSources=0x) into DFB.
- [x] [To address is ExchangeProxy](https://staging.api.0x.org/swap/v1/quote?buyToken=DAI&sellToken=ETH&sellAmount=300000000000000000000&excludedSources=0x).
- [x] [DAI -> ETH](https://staging.api.0x.org/swap/v1/quote?buyToken=ETH&sellToken=DAI&sellAmount=300000000000000000000&excludedSources=0x) is supported.

# Checklist

This PR now includes Balancer changes from https://github.com/0xProject/0x-api/pull/277
DO NOT MERGE until:
- [x] 0xProject/0x-monorepo#2613 has been approved & merged
- [x] @0x/asset-swapper and @0x/contract-addresses are updated to non-gitpkg versions, or at least gitpkg publishes that are on the monorepo development branch
- [x] simbot comparing staging vs prod looks good


src/services/swap_service.ts
```diff
@@ -154,6 +160,20 @@ export class SwapService {
             attributedSwapQuote,
         );
 
+        // set the allowance target based on version. V0 is legacy param to support Nuo integrator
+        let allowanceTarget;
+        switch (swapVersion) {
+            case SwapVersion.V0:
+                allowanceTarget = this._contractAddresses.forwarder;
```
dekz, line 167: If the sellToken is `ETH` the allowance target is `NULL_ADDRESS`, i.e no allowances needed to sell ETH via the Forwarder.

If the sellToken is any ERC20Token, then the allowanceTarget is `erc20Proxy`.

There are no cases where you need to set the allowance on the forwarder.

src/services/swap_service.ts
```diff
@@ -336,18 +365,35 @@ export class SwapService {
...
+                    // forwarder contract. If it's not, then we want to request quotes with the taker set to the
+                    // API's takerAddress query parameter, which in this context is known as `from`.
+                    takerAddress = isETHSell ? this._contractAddresses.forwarder : from || '';
+                    break;
+                case SwapVersion.V1:
+                    // If this is an ETH sell, then we want to request quotes with the taker address
+                    // as the ExchangeProxy so that it can automatically unwrap WETH to ETH. If it's not,
+                    // then we want to request quotes with the taker set to the API's takerAddress query
+                    // parameter, which in this context is known as `from`.
+                    takerAddress = isETHSell ? await this._getExchangeProxyFlashWalletAsync() : from || '';
```
dekz, line 386: This is ALWAYS the ExchangeProxy flash wallet in the new architecture. It always is the taker of orders (both in ETH sells and DAI sells)

src/services/swap_service.ts
```diff
@@ -379,16 +425,34 @@ export class SwapService {
...
     ): Promise<SwapQuoteResponsePartialTransaction> {
-        const extensionContractType = isETHSell ? ExtensionContractType.Forwarder : ExtensionContractType.None;
+        let extensionContractType;
+        switch (swapVersion) {
+            case SwapVersion.V0:
+                extensionContractType = isFromETH ? ExtensionContractType.Forwarder : ExtensionContractType.None;
+                break;
+            case SwapVersion.V1:
+                extensionContractType =
+                    isFromETH || isToETH ? ExtensionContractType.ExchangeProxy : ExtensionContractType.None;
```
dekz, line 440: This should always be `ExtensionContractType.ExchangeProxy`.

The EP allows us to do more magic (withdraw WETH to ETH, deposit ETH to WETH), but it comes at a constant cost. We always use the EP in v1 (never the default). EP (FlashWallet) is always the taker.

src/services/swap_service.ts
```diff
@@ -161,17 +161,19 @@ export class SwapService {
...
+                    allowanceTarget = this._contractAddresses.erc20Proxy;
+                    break;
+                case SwapVersion.V1:
+                    allowanceTarget = this._contractAddresses.exchangeProxyAllowanceTarget;
+                    break;
+                default:
+                    allowanceTarget = NULL_ADDRESS;
+                    break;
+            }
         }
```
dekz: mm more like

```suggestion
            let erc20AllowanceTarget = NULL_ADDRESS;
            switch (swapVersion) {
                case SwapVersion.V0:
                    erc20AllowanceTarget = this._contractAddresses.erc20Proxy;
                    break;
                case SwapVersion.V1:
                    erc20AllowanceTarget = this._contractAddresses.exchangeProxyAllowanceTarget;
                    break;
                default:
                    erc20AllowanceTarget = NULL_ADDRESS;
                    break;
            }
            const allowanceTarget = isETHSell ? NULL_ADDRESS : erc20AllowanceTarget;
```

xianny:
deploy staging

test/swap_test.ts
```diff
@@ -183,7 +198,10 @@ describe(SUITE_NAME, () => {
                 },
             );
         });
-        it('should return a Forwarder transaction for sellToken=ETH', async () => {
+        it.skip('should return a ExchangeProxy transaction for sellToken=ETH', async () => {
```
dekz, line 201: this should be good to not skip now.

.github/workflows/ci-build.yml
```diff
@@ -24,7 +24,7 @@ jobs:
         node-version: ${{ matrix.node-version }}
     
     - name: Cache node modules
-      uses: actions/cache@v1
```
xianny, line 27: upload the node_modules cache even if the job fails

xianny:
deploy staging

src/routers/swap_router.ts
```diff
@@ -8,10 +8,16 @@ import { SwapService } from '../services/swap_service';
...
-    router.get('/', asyncHandler(SwapHandlers.rootAsync.bind(SwapHandlers)));
-    router.get('/quote', asyncHandler(handlers.getSwapQuoteAsync.bind(handlers)));
-    router.get('/tokens', asyncHandler(handlers.getSwapTokensAsync.bind(handlers)));
-    router.get('/prices', asyncHandler(handlers.getTokenPricesAsync.bind(handlers)));
-    router.get('/price', asyncHandler(handlers.getSwapPriceAsync.bind(handlers)));
+    router.get('/v0', asyncHandler(SwapHandlers.rootAsync.bind(SwapHandlers)));
+    router.get('/v0/quote', asyncHandler(handlers.getSwapQuoteV0Async.bind(handlers)));
+    router.get('/v0/tokens', asyncHandler(handlers.getSwapTokensAsync.bind(handlers)));
+    router.get('/v0/prices', asyncHandler(handlers.getTokenPricesAsync.bind(handlers)));
+    router.get('/v0/price', asyncHandler(handlers.getSwapPriceV0Async.bind(handlers)));
```
dorothy-zbornak, line 15: Why not have all the handlers take a `SwapVersion` as the first parameter and just do like `handlers.getSwapPriceAsync.bind(handlers, SwapVersion.V0)` so there isn't as much code dupe.

src/services/swap_service.ts
```diff
@@ -127,9 +134,10 @@ export class SwapService {
         let conservativeBestCaseGasEstimate = new BigNumber(worstCaseGas).plus(gasTokenGasCost);
         if (!skipValidation && from) {
             // Force a revert error if the takerAddress does not have enough ETH.
-            const txDataValue = isETHSell
-                ? BigNumber.min(value, await this._web3Wrapper.getBalanceInWeiAsync(from))
-                : value;
+            const txDataValue =
+                isETHSell || isETHBuy
+                    ? BigNumber.min(value, await this._web3Wrapper.getBalanceInWeiAsync(from))
+                    : value;
```
dorothy-zbornak, line 140: why do we do this? and why does it matter if we're buying ETH?

dekz: It doesn't matter if we're buying, so `isETHBuy` should be removed.

It once mattered when we were selling as geth allowed any `value` and was not validated. So a user wouldn't see any issue with selling 100 ETH (with no ETH balance) until hitting metamask.

Latest geth now validates `value` so this entire check can disappear soon.

dekz:
deploy staging

BMillman19:
deploy staging

BMillman19:
NOTE: this commit: https://github.com/0xProject/0x-api/pull/262/commits/785d31b8c3752741bdd4aa5fdd82a2dac399362e is missing tests that check for the presence of `allowanceTarget` in the `/meta_transaction/v0/quote` and `/meta_transaction/v0/price` endpoints.

SIDE NOTE: should the `allowanceTarget` field that is present in a couple of different interfaces in `types.ts` be an optional property at this point?

BMillman19:
deploy staging

dekz:
deploy staging

dekz:
deploy staging

dekz:
deploy staging

xianny:
deploy staging

mzhu25:
deploy staging

opaolini:
deploy staging

mzhu25:
deploy staging

mzhu25:
deploy staging

dekz:
deploy staging

dekz:
### Realized prices
image

### Revert rate
image

### Response time 
image

dekz:
deploy staging

dekz:
## Typescript 3.9.6 update
### Response time
image

dekz:
deploy production

PR #267, opened 2020-06-23
feat: Unique quote ID

steveklebanoff:
# Description

Hijacks the timestamp calldata to be treated as unique id: `[10 random numbers][timestamp in seconds]`

This will allow us to more easily attribute data in our data pipeline, and communicate unique ids for quotes to RFQT market makers

# Testing Instructions

# Checklist

### DIFF
--- a/src/utils/number_utils.ts
+++ b/src/utils/number_utils.ts
@@ -5,15 +5,15 @@ export const numberUtils = {
         const max = Math.floor(maximumSpecified);
         return Math.floor(Math.random() * (max - min + 1)) + min; // The maximum is inclusive and the minimum is inclusive
     },
-    // creates a random number of desired length by stringing together
-    // random integers from 1-9, guaranteeing the
-    // the .toString().length of the return value to equal to numberLength parameter
-    randomNumberOfLength: (numberLength: number): number => {
+    // creates a random hex number of desired length by stringing together
+    // random integers from 1-15, guaranteeing the
+    // result is a hex number of the given length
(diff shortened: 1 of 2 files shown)


src/utils/number_utils.ts
```diff
@@ -0,0 +1,19 @@
...
+    // from MDN https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Math/random
+    randomNumberInclusive: (minimumSpecified: number, maximumSpecified: number): number => {
+        const min = Math.ceil(minimumSpecified);
+        const max = Math.floor(maximumSpecified);
+        return Math.floor(Math.random() * (max - min + 1)) + min; // The maximum is inclusive and the minimum is inclusive
+    },
+    // creates a random number of desired length by stringing together
+    // random integers from 1-9, guaranteeing the
+    // the .toString().length of the return value to equal to numberLength parameter
+    randomNumberOfLength: (numberLength: number): number => {
```
steveklebanoff, line 11: I know there are cooler and faster ways to achieve this (i.e. https://stackoverflow.com/a/3437180) but I wanted to guarantee that `.toString().length` = 10 so that we could correctly grab the following `timestamp` even in 300 years when `unixTimestampInSeconds.toString().length` is 11 😉 

image

steveklebanoff:
Upon further review of the data pipeline, I'm going to push up some commits that revise this to take the `nanoseconds` + `random modified` approach as I believe this will make things easier on the postgres side

PR #272, opened 2020-06-30
fix: Update to ethgasstation proxy endpoint

dekz:
https://github.com/0xProject/0x-monorepo/pull/2614


dekz:
deploy staging

dekz:
deploy production

PR #275, opened 2020-07-03
fix: Changes latency to be capped at 600ms, added some extra typing

PirosB3:
# Description

We noticed that RFQT latency on `api.0x.org` can be too high at times and may cause downstream latency to our customers, so we are going to cap HTTP RFQT timeouts to 600ms.  This PR sets the `makerEndpointMaxResponseTimeMs` parameter that can be optionally specified in `RfqtRequestOpts`, and that defaults to 1 second. We are intentionally enforcing this parameter to be `600`.

⚠️ We noticed that 0x API passes to Asset Swapper some parameters that do not exist anymore. More precisely:

- `skipBuyRequests` is an old relic of when our RFQT bot only supported sell requests. skipBuyRequests is not used at all by AssetSwapper, since it’s not present in [RfqtRequestOpts](https://github.com/0xProject/0x-monorepo/blob/85ca926fd012d9233c01a0a2a58e8709d5dabe32/packages/asset-swapper/src/types.ts#L202).
- `slippagePercentage` was removed [here](https://github.com/0xProject/0x-monorepo/pull/2513) and passing this parameter to SwapQuoter will have no effect at all.

Can I get a sign off from @dekz and/or @dorothy-zbornak that the removal of these 2 parameters does not cause any issue downstream? and that these parameters are simply technical debt? 

# Testing Instructions

Once I have a first approval that my PR does not introduce new issues, I will try it in staging

# Checklist

-   [x] Get a sign-off from @dekz and/or @dorothy-zbornak re. removal of the 2 parameters
-   [x] Test in staging
-   [ ] Merge and deploy to Prod

### DIFF
--- a/src/config.ts
+++ b/src/config.ts
@@ -283,6 +283,8 @@ const feeSchedule: { [key in ERC20BridgeSource]: BigNumber } = Object.assign(
     })),
 );
 
+export const RFQT_REQUEST_MAX_RESPONSE_MS = 600;
+
 export const ASSET_SWAPPER_MARKET_ORDERS_OPTS: Partial<SwapQuoteRequestOpts> = {
     excludedSources: EXCLUDED_SOURCES,
     bridgeSlippage: DEFAULT_QUOTE_SLIPPAGE_PERCENTAGE,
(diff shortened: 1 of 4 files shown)


dekz:
yeah the removal of those params is fine.

PirosB3:
deploy staging

PirosB3:
deploy prod

PR #292, opened 2020-07-23
fix: synchronization false unless tests

dekz:
Disables synchronization unless we're running in test env


dekz:
deploy staging

dekz:
deploy production

PR #298, opened 2020-07-27
feat: Log QuoteReport to Kibana

steveklebanoff:
Logs QuoteReports for RFQT firm quotes so that they can be consumed by Kibana and later put into metabase.

When a firm quote is to be submitted by the taker, it is logged with a `uniqueIdString` to be associated with the random  number + timestamp when the fill occurs on chain (See https://github.com/0xProject/0x-api/pull/281)

This does not work for meta txns, because the additional calldata is generated as the time that the meta txn service submits the fill on-chain, which is a different request than when the 'order salad' was created. In that case, we log the `zeroExTransactionHash` so we can associate it in metabase.

Also, we log the sources in "chunks", as outputting the entire object was too big to be parsed

# TODO
- [x] Try on staging
- [x] Update to latest version of asset-swapper
- [x] Understand if inconsistencies on staging v1 price and quote are due to this
- [x] Consider adding additional metadata (i.e. api key or affiliate address)
- [ ] Consider putting back `apiKey` check prior to logging back for meta_txn 

# Description

# Testing Instructions

# Checklist


package.json
```diff
@@ -128,7 +128,7 @@
     },
     "dependencies": {
         "@0x/assert": "^3.0.4",
-        "@0x/asset-swapper": "0xProject/gitpkg-registry#0x-asset-swapper-v4.6.0-ae2a6fb68",
+        "@0x/asset-swapper": "0xProject/gitpkg-registry#0x-asset-swapper-v4.7.0-aae1a6179",
```
steveklebanoff, line 131: ⚠️  need to figure out why I needed to update this to get the `QuoteReport` included ⚠️

src/services/swap_service.ts
```diff
@@ -208,6 +208,7 @@ export class SwapService {
             sources: serviceUtils.convertSourceBreakdownToArray(sourceBreakdown),
             orders: serviceUtils.cleanSignedOrderFields(orders),
             allowanceTarget,
+            quoteReport,
```
steveklebanoff, line 211: ⚠️ Need to make sure this never gets returned to the end user ⚠️

steveklebanoff:
deploy staging

steveklebanoff:
deploy staging

package.json
```diff
@@ -128,7 +128,7 @@
     },
     "dependencies": {
         "@0x/assert": "^3.0.4",
-        "@0x/asset-swapper": "0xProject/gitpkg-registry#0x-asset-swapper-v4.6.0-ae2a6fb68",
+        "@0x/asset-swapper": "0xProject/gitpkg-registry#0x-asset-swapper-v4.6.0-766888ec4",
```
steveklebanoff, line 131: ⚠️ This updates to https://github.com/0xProject/0x-monorepo/pull/2646 ⚠️ 

Will need to modify this once 2646 is merged

steveklebanoff:
deploy staging

package.json
```diff
@@ -128,7 +128,7 @@
     },
     "dependencies": {
         "@0x/assert": "^3.0.4",
-        "@0x/asset-swapper": "0xProject/gitpkg-registry#0x-asset-swapper-v4.6.0-ae2a6fb68",
+        "@0x/asset-swapper": "0xProject/gitpkg-registry#0x-asset-swapper-v4.6.0-498a50b22",
```
steveklebanoff, line 131: ⚠️ will have to update this to `master` or remove this change prior to merging ⚠️

steveklebanoff:
deploy staging

steveklebanoff:
deploy staging

steveklebanoff:
deploy staging

steveklebanoff:
deploy staging

src/services/meta_transaction_service.ts
```diff
@@ -206,6 +216,11 @@ export class MetaTransactionService {
             )
             .callAsync();
 
+        // log quote report and associate with txn hash if this is an RFQT firm quote
```
alexkroeger, line 219: Does this mean we don't get a report if the meta TX didn't consider an RFQT quote

src/utils/service_utils.ts
```diff
@@ -58,7 +58,13 @@ export const serviceUtils = {
         return attributedSwapQuote;
     },
 
-    attributeCallData(data: string, affiliateAddress?: string): string {
+    attributeCallData(
+        data: string,
+        affiliateAddress?: string,
+    ): {
+        affiliatedData: string;
```
feuGeneA, line 65: Yet another naming nit for you :grin: 
I think this should be `affiliateData` (drop the extra `d`).  I realize you just copied it from the variable already in use in the function, but I feel you should correct it there as well, now that the name is being promoted into the interface.

Note that the method giving this value is `affiliateCallDataEncoder.encode()` (no extra `d` there), and you can also note in `git blame` that that variable was added to this existing method, and it seems it was simply typo-ed in that addition.

src/services/swap_service.ts
feuGeneA: You did this, over in the handler, with the `cleanedQuote`, so this comment can be resolved, right?

feuGeneA: You did this, over in the handler, with the `cleanedQuote`, so this comment can be resolved, right?

steveklebanoff: Correct 👍

src/services/meta_transaction_service.ts
steveklebanoff: right now all meta-txn calls are RFQT enabled b/c they all come from matcha. 

for non-meta-txn calls I _do_ limit it only to RFQT firm quotes (https://github.com/0xProject/0x-api/blob/ba495a78e305113478396539e74577f84e2f3fba/src/handlers/swap_handlers.ts#L67) since there will be a lot of verbose log lines emitted. 

if we find this data useful for firm quotes we can consider opening up to all requests (as long as get confirmation from Oskar that our log serve can handle it)

src/types.ts
```diff
@@ -389,6 +390,7 @@ export interface SwapQuoteResponsePartialTransaction {
     to: string;
     data: string;
     value: BigNumber;
+    uniqueIdString: string;
```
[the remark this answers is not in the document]
steveklebanoff: ah, great question, as this does look a little unexpected.

the timestamp that we encode in the calldata is generated in a bit of a strange way to help support our data pipeline to be able to differentiate the 'timestamp' vs the 'random number', see https://github.com/0xProject/0x-api/pull/281.

the `string` suffix in this param is met to indiciate that this is a special "string representation" of that items that make up the encoded timestamp 

the `uniqueIdString` looks like ( `${randomNumber}-${timestampInSeconds}`) which is noteably different than the BigNumber sent out as the calldata. i could imagine us using this string representation in the future when we communicate the order salad to market makers

that being said -- this does all feel a bit confusing, and i'm open to any alternative suggestions

src/types.ts
```diff
@@ -410,7 +412,9 @@ export interface GetSwapQuoteResponse extends SwapQuoteResponsePartialTransactio
     gas: BigNumber;
     estimatedGas: BigNumber;
     estimatedGasTokenRefund: BigNumber;
+    uniqueIdString: string;
```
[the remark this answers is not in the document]
steveklebanoff: great catch -- thank you 939135fd72e13cf980bfe8ee753deff7666cc99f

src/services/meta_transaction_service.ts
```diff
@@ -111,7 +119,7 @@ export class MetaTransactionService {
             rfqt: _rfqt,
         };
 
-        let swapQuote;
+        let swapQuote: MarketSellSwapQuote | MarketBuySwapQuote | undefined;
```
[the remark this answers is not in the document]
steveklebanoff: nope was not required -- but I noticed `swapQuote` was `any` and wanted to add some typesafety here.  this helped me feel more confident that my changes worked

src/utils/quote_report_utils.ts
```diff
@@ -0,0 +1,53 @@
+import { QuoteReport } from '@0x/asset-swapper';
+import _ = require('lodash');
+
+import { NUMBER_SOURCES_PER_LOG_LINE } from '../constants';
+import { logger } from '../logger';
+
+interface QuoteReportSubmissionByTaker {
```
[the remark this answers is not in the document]
steveklebanoff: i like the name nits -- naming is hard & it's important other people can read this code :)

thanks for the suggestion, going to take it! 3dae495

src/utils/quote_report_utils.ts
```diff
@@ -0,0 +1,53 @@
...
+    quoteReport: QuoteReport;
+    submissionBy: 'taker';
+    uniqueIdString: string;
+}
+interface QuoteReportSubmissionByMetaTxn {
+    quoteReport: QuoteReport;
+    submissionBy: 'metaTxn';
+    zeroExTransactionHash: string;
+}
+type LogQuoteReportOptions = QuoteReportSubmissionByTaker | QuoteReportSubmissionByMetaTxn;
```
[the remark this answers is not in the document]
steveklebanoff: great call -- i blindly appended `Options` to the fn name but you are right that my param name choice suggests something different, 3dae495

src/utils/service_utils.ts
steveklebanoff: I think the `d` at the end actually is an accurate representation of what it is -- it takes `calldata` and appends affiliate data onto it, so it in my mind it is calldata that has been `affiliated` :)

feuGeneA: Makes sense to me, thanks for clarifying.

src/types.ts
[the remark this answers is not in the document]
feuGeneA: How about `decodedUniqueId`?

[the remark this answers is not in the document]
steveklebanoff: i'm down -- c5593db

steveklebanoff:
deploy staging

steveklebanoff:
deploy staging

PR #307, opened 2020-08-04
fix: return input in depth-chart

dekz:
Returns the input values rather than the output values. This is better aligned with the way it is to be used.

For example, sell BAL, buy DAI, it returns the price and the amount of BAL that can be sold at that price.
You can take that amount and perform a swap quote either selling or buying the amount returned.. 

Previous it returned the amount of DAI available at that price.

# Description

# Testing Instructions

# Checklist


dekz:
deploy production

dekz:
deploy production

PR #320, opened 2020-08-14
feat: Allow indicative RFQ-T quotes without takerAddress

feuGeneA:
# Description

Changed RFQ-T enablement constraints such that a `takerAddress` query parameter is no longer required when coming through the `/price` endpoint.

# Testing Instructions

Use [Matcha on staging](https://api-staging.matcha.xyz/markets/DAI/USDC), and don't connect your wallet, and request some pricing, and check the API staging logs to ensure that RFQ-T liquidity was accessed.

# Checklist

-   [x] Implement.
-   [x] Test manually on staging.
-   [x] Merge [monorepo PR](https://github.com/0xProject/0x-monorepo/pull/2684) and re-pin the dependency herein to the `development` branch.

### DIFF
--- a/package.json
+++ b/package.json
@@ -126,7 +126,7 @@
     },
     "dependencies": {
         "@0x/assert": "^3.0.4",
-        "@0x/asset-swapper": "0xProject/gitpkg-registry#0x-asset-swapper-v4.6.0-db0f3a878",
+        "@0x/asset-swapper": "0xProject/gitpkg-registry#0x-asset-swapper-v4.6.0-7b0a1c363",
         "@0x/connect": "^6.0.4",
         "@0x/contract-addresses": "^4.11.0",
         "@0x/contract-wrappers": "^13.7.0",
--- a/package.json
+++ b/package.json
@@ -126,7 +126,7 @@
     },
     "dependencies": {
         "@0x/assert": "^3.0.4",
-        "@0x/asset-swapper": "0xProject/gitpkg-registry#0x-asset-swapper-v4.6.0-b065d8015",
+        "@0x/asset-swapper": "0xProject/gitpkg-registry#0x-asset-swapper-v4.6.0-e9b0ac882",
         "@0x/connect": "^6.0.4",
         "@0x/contract-addresses": "^4.11.0",
         "@0x/contract-wrappers": "^13.7.0",
--- a/package.json
+++ b/package.json
@@ -126,7 +126,7 @@
     },
     "dependencies": {
         "@0x/assert": "^3.0.4",
-        "@0x/asset-swapper": "0xProject/gitpkg-registry#0x-asset-swapper-v4.6.0-e9b0ac882",
+        "@0x/asset-swapper": "0xProject/gitpkg-registry#0x-asset-swapper-v4.6.0-db0f3a878",
         "@0x/connect": "^6.0.4",
         "@0x/contract-addresses": "^4.11.0",
         "@0x/contract-wrappers": "^13.7.0",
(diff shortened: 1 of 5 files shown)


feuGeneA:
deploy staging

feuGeneA:
deploy staging

feuGeneA:
deploy staging

feuGeneA:
deploy staging

feuGeneA:
deploy staging

src/handlers/swap_handlers.ts
```diff
@@ -403,14 +403,24 @@ const parseGetSwapQuoteRequestParams = (
...
-        takerAddress && apiKey
-            ? {
-                  intentOnFilling: endpoint === 'quote' && req.query.intentOnFilling === 'true',
-                  isIndicative: endpoint === 'price',
-                  nativeExclusivelyRFQT,
-              }
-            : undefined;
+    const rfqt: Pick<RfqtRequestOpts, 'intentOnFilling' | 'isIndicative' | 'nativeExclusivelyRFQT'> = (() => {
+        if (apiKey) {
+            if (takerAddress) {
```
PirosB3, line 408: could you make the `if` statement the following? (just a suggestion)

```
if (endpoint === 'quote' && takerAddress) {
                return {
                    intentOnFilling: req.query.intentOnFilling === 'true',
                    isIndicative: false,
                    nativeExclusivelyRFQT,
                };
} else if (endpoint === 'price') {
                return {
                    intentOnFilling: false,
                    isIndicative: true,
                    nativeExclusivelyRFQT,
                };
} else {
    return undefined;
}
```

src/services/swap_service.ts
```diff
@@ -471,7 +471,7 @@ export class SwapService {
                 },
             ]);
         }
-        if (apiKey !== undefined && (isETHSell || from !== undefined)) {
+        if (apiKey !== undefined && (isETHSell || from !== undefined || (rfqt && rfqt.isIndicative))) {
```
PirosB3, line 474: Can you explain this flow to me? why is this change needed?

feuGeneA: This conditional says, "if there's an API key AND (if either it's a forwarder tx (`isETHSell`) OR there's a taker address (`from`) OR it's an indicative quote".

Without this addition, an indicative quote request without a taker address cannot flow through here, so `_rfqt` remains undefined, so RFQT never gets enabled downstream.

src/handlers/swap_handlers.ts
feuGeneA: Great suggestion, thank you.  Addressed in https://github.com/0xProject/0x-api/pull/320/commits/e63ee3b02ede33bd30f566ae8db6825d296523d9

package.json
```diff
@@ -126,7 +126,7 @@
     },
     "dependencies": {
         "@0x/assert": "^3.0.4",
-        "@0x/asset-swapper": "0xProject/gitpkg-registry#0x-asset-swapper-v4.6.0-a2f0d5eed",
+        "@0x/asset-swapper": "0xProject/gitpkg-registry#0x-asset-swapper-v4.6.0-db0f3a878",
```
PirosB3, line 129: Should this contain the hash of the latest `master` when https://github.com/0xProject/0x-monorepo/pull/2684#pullrequestreview-478852431 is merged?

src/services/swap_service.ts
PirosB3: This now makes sense for me, however, I'm a little concerned that whoever next touches this code won't understand the logic. Is it worth adding some comments explaining the reasoning behind L479? alternatively, we could refactor the `if` statement to use self-explaining variablenames. Example:

```
if (hasAPIKey && (isETHSell || hasTakerAddress || isIndicativeQuote)) {

}
```

package.json
feuGeneA: Yup, done in https://github.com/0xProject/0x-api/pull/320/commits/c38fca77f088a4dbf328d17c194e02dffbb265ea

feuGeneA:
deploy staging

PR #327, opened 2020-08-20
feat: update fee schedule for v1

dekz:
# Description
* Updates the fee schedule for the latest EP
* Uses gross gas for just the source and adds in the base cost later. So the path optimizer doesn't pay the full hit of overhead.
* Removes the buffer if the sources are more deterministic

# Testing Instructions

# Checklist


dekz:
deploy production

dekz:
deploy production

PR #342, opened 2020-08-25
feat: bump pino, use pino.destination, allow async logging

opaolini:
# Description

This PR:
- Bumps `pino` to its latest version.
- Uses a buffered version of `pino.destination` which improves performance.
- Allows to enable `async` logging through environment variables.

# Testing Instructions

# Checklist


README.md
```diff
@@ -93,6 +93,10 @@ To get a local development version of `0x-api` running:
...
+| `LOG_ASYNC_ENABLED`                    | Default: `false`                                                | A boolean determining whether or not to enable async logging.                                                                                                                                             |
+| `LOG_BUFFER_BYTES`                     | Default: `0`                                                    | The log write buffer size.                                                                                                                                                                                |
+| `LOG_ASYNC_FLUSH_INTERVAL_SEC`         | Default: `10`                                                   | The log flush interval in seconds when `LOG_ASYNC_ENABLED` is set to true.                                                                                                                                |
```
PirosB3, line 99: So logs could be delayed by up to 10 seconds? That's fine for me, but is there any process or alarm that depends on data being available much faster?

src/logger.ts
```diff
@@ -1,16 +1,39 @@
...
+    },
+    pinoDestination,
+);
+
+if (LOG_ASYNC_ENABLED) {
+    // asynchronously flush every 10 seconds to keep the buffer empty
+    // in periods of low activity
+    setInterval(() => {
+        logger.flush();
+    }, ONE_SECOND_MS * LOG_ASYNC_FLUSH_INTERVAL_SEC).unref();
```
PirosB3, line 34: What is "unref" exactly?

src/runners/utils.ts
```diff
@@ -50,13 +50,16 @@ export function createDefaultServer(
                 await dependencies.connection.close();
             }
             if (!server.listening) {
+                logger.flush();
```
PirosB3, line 53: It looks like all these codepaths lead to flush at the end. Is there a reason to do multiple flushes?

src/logger.ts
opaolini: When called the `Timeout` object will not require the Node.js event loop to remain active. If there is no other activity keeping the event loop running, the process may exit before the Timeout object's callback is invoked. Which means this will not prevent nodejs process from exiting before callback is invoked.

[0] - https://nodejs.org/api/timers.html#timers_timeout_unref

README.md
opaolini: For the logs to be delayed:
- `sync` has to be set to `false` (enabling async logging)
- there has to be minimal activity to not fill up the given `LOG_BUFFER` (for example `4kb`) and flush automatically.
- The external services that depend on timing can use the `pino`'s timestamp contained in the log object, not on the flush time.

src/runners/utils.ts
opaolini: You can see the `logger.flush()` precedes `process.exit`, we can make it a bit nice by lifting it to a function:
```
const exitWithCode = (exitCode) => {
                logger.flush();
                process.exit(exitCode);

};
```

src/logger.ts
PirosB3: cool!

opaolini:
deploy staging

PR #350, opened 2020-09-08
fix: throw error on NaN Integer type ENV variables

kimpers:
# Description

Noticed some issues with missing Integer type env variables. `parseInt(undefined, 10` returns `NaN` instead of throwing, which makes the check not error out even if the value is missing.

# Testing Instructions

# Checklist


src/config.ts
```diff
@@ -507,6 +507,9 @@ function assertEnvVarType(name: string, value: any, expectedType: EnvVarType): a
         case EnvVarType.Integer:
             try {
                 returnValue = parseInt(value, 10);
+                if (isNaN(returnValue)) {
+                    throw new Error();
```
dekz, line 511:
```suggestion
                    throw new Error(`${name} must be a valid integer, found ${value}.`);
```

kimpers: It's wrapped in a catch below which throws with an appropriate message, I don't think the message here has any impact.

opaolini:
I think we should just have:
```
        case EnvVarType.Integer:
                returnValue = parseInt(value, 10);
                if (isNaN(returnValue)) {
                    throw new Error(`${name} must be a valid integer, found ${value}.`);
                }
            return returnValue;
```

And maybe just have the:
```
        case EnvVarType.KeepAliveTimeout:
            try {
                returnValue = parseInt(value, 10);
            } catch (err) {
                throw new Error(`${name} must be a valid integer, found ${value}.`);
            }
            return returnValue;
        case EnvVarType.ChainId:
            try {
                returnValue = parseInt(value, 10);
            } catch (err) {
                throw new Error(`${name} must be a valid integer, found ${value}.`);
            }
            return returnValue;
```

As
```
        case EnvVarType.Integer:
        case EnvVarType.KeepAliveTimeout:
        case EnvVarType.ChainId:
                returnValue = parseInt(value, 10);
                if (isNaN(returnValue)) {
                    throw new Error(`${name} must be a valid integer, found ${value}.`);
                }
            return returnValue;
```

opaolini (approved):

kimpers:
deploy staging

kimpers:
deploy production

PR #352, opened 2020-09-10
feat: Quote report based price comparison

kimpers:
# Description
This adds an optional `prices` attribute in the `/quote` and `/price` responses with the estimated prices for filling the entire order at any of the considered external sources.

Depends on 0xProject/0x-monorepo#2702

# Testing Instructions

# Checklist

-   [X] Prefix PR title with `[WIP]` if necessary.


kimpers:
deploy staging

package.json
```diff
@@ -127,7 +127,7 @@
     },
     "dependencies": {
         "@0x/assert": "^3.0.4",
-        "@0x/asset-swapper": "0xProject/gitpkg-registry#0x-asset-swapper-v4.6.0-0ba79b060",
+        "@0x/asset-swapper": "0xProject/gitpkg-registry#0x-asset-swapper-v4.6.0-26ba3fac4",
```
kimpers, line 130: **TODO**: update to `development` branch build when 0xProject/0x-monorepo#2702 is merged

src/handlers/meta_transaction_handlers.ts
```diff
@@ -171,7 +183,23 @@ export class MetaTransactionHandlers {
...
+                const prices = priceComparisonUtils.getPriceComparisonFromQuote(CHAIN_ID, swapVersion, params, {
+                    ...metaTransactionPrice,
+                    buyTokenAddress,
+                    sellTokenAddress,
+                });
+
+                if (prices) {
+                    priceResponse = {
+                        ...metaTransactionPriceResponse,
+                        prices,
```
kimpers, line 198: Not sure if the attribute should be `prices` or `comparisons` Any thoughts on this? 🤔

src/utils/price_comparison_utils.ts
```diff
@@ -0,0 +1,121 @@
...
+                    .sort((a, b) =>
+                        isSelling ? b.makerAmount.comparedTo(a.makerAmount) : a.takerAmount.comparedTo(b.takerAmount),
+                    ),
+                'liquiditySource',
+            );
+
+            const sourcePrices: SourceComparison[] = uniqueRelevantSources.map(source => {
+                const { liquiditySource, makerAmount, takerAmount } = source;
+                const gas = new BigNumber(
+                    gasSchedule[swapVersion][source.liquiditySource]!((source as BridgeReportSource).fillData),
```
dekz, line 96: Note v0 gas schedule is total gross including 0x overhead. (E.g 300k Uniswap)
v1 gas schedule is just the individual source cost estimate. (E.g 150k Uniswap where we add the base cost in later, 150 Uni  + 150 Base)

dekz (approved):
This looks good to me, made a few comments.

dekz:
Let's get this rebased off master and deployed to staging, I'll play around with it a bit.

src/utils/price_comparison_utils.ts
```diff
@@ -0,0 +1,121 @@
...
+
+import { getTokenMetadataIfExists } from './token_metadata_utils';
+
+// Exclude internal sources from comparison
+// NOTE: asset-swapper does not return fillData for Native & MultiHop sources
+const excludedLiquiditySources = new Set([
+    ERC20BridgeSource.Native,
+    ERC20BridgeSource.MultiHop,
+    ERC20BridgeSource.MultiBridge,
+    ERC20BridgeSource.LiquidityProvider,
```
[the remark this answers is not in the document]
kimpers: For Native 0x & Multibridge I currently do not pass back the `fillData` from asset-swapper, I could easily add them but I was not sure how we felt about providing these comparison values for our proprietary sources. It's not something that Matcha would be displaying but we could return them if we want. @dekz do you prefer returning all sources in the prices response?

src/utils/price_comparison_utils.ts
kimpers: Good point! It probably makes more sense to always use v1 gas schedule as it is the most accurate representation of the gas cost for each source independently of 0x overhead 👍

src/utils/price_comparison_utils.ts
```diff
@@ -0,0 +1,121 @@
...
+                    .slice()
+                    .sort((a, b) =>
+                        isSelling ? b.makerAmount.comparedTo(a.makerAmount) : a.takerAmount.comparedTo(b.takerAmount),
+                    ),
+                'liquiditySource',
+            );
+
+            const sourcePrices: SourceComparison[] = uniqueRelevantSources.map(source => {
+                const { liquiditySource, makerAmount, takerAmount } = source;
+                const gas = new BigNumber(
```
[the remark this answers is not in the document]
kimpers: Yeah, Matcha already uses the ETH/USD for calculating the tx cost so it shouldn't be any extra work there.

kimpers:
> Let's get this rebased off master and deployed to staging, I'll play around with it a bit.

Updated the branch and deployed it to the Matcha staging 👍

src/handlers/meta_transaction_handlers.ts
```diff
@@ -369,6 +393,11 @@ const parseGetTransactionRequestParams = (req: express.Request): GetTransactionR
         req.query.excludedSources === undefined
             ? []
             : parseUtils.parseStringArrForERC20BridgeSources((req.query.excludedSources as string).split(','));
+
+    // tslint:disable-next-line:boolean-naming
+    const includePriceComparisons =
```
steveklebanoff, line 398: nit:

 ```
const includePriceComparisons = req.query.includePriceComparisons ?  req.query.includePriceComparisons === 'true' : false;
```

File: src/services/meta_transaction_service.ts [hunk 7]
```diff
@@ -228,7 +232,8 @@ export class MetaTransactionService {
             .callAsync();
 
         // log quote report and associate with txn hash if this is an RFQT firm quote
-        if (quoteReport) {
+        const shouldLogQuoteReport = quoteReport && params.apiKey !== undefined;
```
steveklebanoff, line 235: i might be missing something: but I expected every meta txn request to always have an api key -- is that the case?

src/types.ts
```diff
@@ -421,6 +421,13 @@ export interface GetSwapQuoteResponse extends SwapQuoteResponsePartialTransactio
     estimatedGasTokenRefund: BigNumber;
     allowanceTarget?: string;
     quoteReport?: QuoteReport;
+    prices?: SourceComparison[];
+}
+
+export interface SourceComparison {
+    name: ERC20BridgeSource | '0x';
```
steveklebanoff, line 428: is this just to support saying `0x` instead of `Native`?

could we leave this as `name: ERC20BridgeSource;` and do the rename of `Native` -> `0x` after calling the function that returns the `SourceComparison`?  That would feel more natural to me

src/types.ts
```diff
@@ -421,6 +421,13 @@ export interface GetSwapQuoteResponse extends SwapQuoteResponsePartialTransactio
...
+    price: BigNumber | null;
```
steveklebanoff, line 429: any strong reason to use nulls instead of undefined? 

if we used undefined we could do `price?: BigNumber;` instead of `price: BigNumber | null;` which feels more consistent w/ our other code

src/handlers/meta_transaction_handlers.ts
steveklebanoff: would it make sense to call this `priceComparisons` since the query param is `includePriceComparisons `?

src/utils/quote_report_utils.ts
```diff
@@ -18,7 +18,18 @@ type QuoteReportLogOptions = QuoteReportForTakerTxn | QuoteReportForMetaTxn;
 
 export const quoteReportUtils = {
```
steveklebanoff, line 19: now that we are logging indicative and firm quotes -- would it be easy to log whether or not this was firm or indicative in the `logQuoteReport` fn ? i think this would be useful, but we also would have to update the metabase consumer to support this

src/utils/price_comparison_utils.ts
```diff
@@ -0,0 +1,110 @@
...
+    quoteReport?: { sourcesConsidered: QuoteReportSource[] };
+}
+
+export const priceComparisonUtils = {
+    getPriceComparisonFromQuote(
+        chainId: ChainId,
+        params: PartialRequestParams,
+        quote: PartialQuote,
+    ): SourceComparison[] | undefined {
+        if (!quote.quoteReport) {
```
steveklebanoff, line 44: could we just make `quoteReport` required in `PartialQuote`, and not call this fn if `quoteReport` is `undefined`?

src/utils/price_comparison_utils.ts
```diff
@@ -0,0 +1,110 @@
...
+            logger.error('Missing quote report, cannot calculate price comparison');
+            return undefined;
+        }
+
+        // NOTE: don't fail quote request if comparison calculations error out
```
steveklebanoff, line 49: nice safeguard 👍

src/utils/price_comparison_utils.ts
```diff
@@ -0,0 +1,110 @@
...
+        try {
+            const buyToken = getTokenMetadataIfExists(quote.buyTokenAddress, chainId);
+            const sellToken = getTokenMetadataIfExists(quote.sellTokenAddress, chainId);
+
+            const isSelling = !!params.sellAmount;
+
+            const { sourcesConsidered } = quote.quoteReport;
+
+            // Filter matching amount samples with a valid result
+            const fullTradeSources = sourcesConsidered.filter(s =>
```
steveklebanoff, line 59: nice little fn w/ a clear name 💯

src/utils/price_comparison_utils.ts
```diff
@@ -0,0 +1,110 @@
...
+                    ? s.takerAmount.isEqualTo(quote.sellAmount) && s.makerAmount.isGreaterThan(ZERO)
+                    : s.makerAmount.isEqualTo(quote.buyAmount) && s.takerAmount.isGreaterThan(ZERO),
+            );
+
+            // Sort by amount the user will receive and deduplicate to only take the best option for e.g. Kyber
+            const uniqueSources = _.uniqBy(
+                fullTradeSources
+                    .slice()
+                    .sort((a, b) =>
+                        isSelling ? b.makerAmount.comparedTo(a.makerAmount) : a.takerAmount.comparedTo(b.takerAmount),
```
steveklebanoff, line 70: this felt a bit hard to read -- could we instead do something like:

```
const uniqueSources = _.uniqBy(fullTradeSources.slice());

const sortedSources = isSellling ? _.sortBy(uniqueSources, ...) : _.sortBy(uniqueSources, ...)
# or
const sortedSources = isSellling ? uniqueSources.sort(...) : uniqueSources.sourt(...)
```

src/utils/price_comparison_utils.ts
```diff
@@ -0,0 +1,110 @@
...
+                    ),
+                'liquiditySource',
+            );
+
+            const sourcePrices: SourceComparison[] = uniqueSources.map(source => {
+                const { liquiditySource, makerAmount, takerAmount } = source;
+                let gas: BigNumber;
+                if (liquiditySource === ERC20BridgeSource.Native) {
```
steveklebanoff, line 78: is there an easy way to ensure this stays up to date when we add new liquidity sources?  or perhaps an existing helper fn we could use?

src/utils/price_comparison_utils.ts
```diff
@@ -0,0 +1,110 @@
...
+                    : unitTakerAmount.dividedBy(unitMakerAmount).decimalPlaces(buyToken.decimals);
+
+                return {
+                    name: renameIfNativeSource(liquiditySource),
+                    price,
+                    gas,
+                };
+            });
+
+            // Add null values for all sources we don't have a result for so that we always have a full result set in the response
```
steveklebanoff, line 102: just curious -- why is this needed?

test/price_comparison_utils_test.ts
```diff
@@ -0,0 +1,350 @@
...
+
+const WETH = getTokenMetadataIfExists('WETH', ChainId.Mainnet);
+const DAI = getTokenMetadataIfExists('DAI', ChainId.Mainnet);
+const USDC = getTokenMetadataIfExists('USDC', ChainId.Mainnet);
+const buyAmount = new BigNumber('23318242912334152626');
+const sellAmount = new BigNumber('70100000000000000');
+
+const SUITE_NAME = 'priceComparisonUtils';
+
+describe(SUITE_NAME, () => {
```
steveklebanoff, line 20: nice tests!

steveklebanoff (changes requested):
a few minor comments but overall very nice work!

src/utils/quote_report_utils.ts
kimpers: I might missing something but nothing should have changed regarding when we log reports.

We are only logging from the `/quote` endpoints if an api key is supplied. Is that not correct?

src/types.ts
steveklebanoff: makes sense 👍

src/utils/price_comparison_utils.ts
```diff
@@ -55,15 +55,14 @@ export const priceComparisonUtils = {
...
-                    .slice()
-                    .sort((a, b) =>
-                        isSelling ? b.makerAmount.comparedTo(a.makerAmount) : a.takerAmount.comparedTo(b.takerAmount),
-                    ),
-                'liquiditySource',
-            );
+            // NOTE: Sort sources by the best outcome for the user
+            // if the user is selling we want to maximize the maker amount they will receive
+            // if the user is buying we want to minimize the taker amount they have to pay
+            const sortedSources = isSelling
```
steveklebanoff, line 61: thanks for this change and additional comments 🙏

src/types.ts
steveklebanoff: thanks for this change!

src/utils/quote_report_utils.ts
steveklebanoff: you are right -- my bad, i misread what was happening -- got a bit mixed up with us generating quote repots but not logging them.  thanks for the clarification.

src/utils/price_comparison_utils.ts
steveklebanoff: great change -- thanks!

src/utils/price_comparison_utils.ts
steveklebanoff: i see -- makes sense -- thanks

steveklebanoff (approved):

kimpers:
deploy staging

kimpers:
deploy production

kimpers:
deploy production

PR #356, opened 2020-09-16
feat: add uniswap VIP support

dorothy-zbornak:
Also add `includedSources` query param support to end points. I hate how much work this totally uncalled for feature took. :weary: 

# Description

# Testing Instructions

# Checklist

### DIFF
--- a/src/config.ts
+++ b/src/config.ts
@@ -402,9 +402,11 @@ export const GAS_SCHEDULE_V1: FeeSchedule = {
     },
     [ERC20BridgeSource.MultiBridge]: () => 3.5e5,
     [ERC20BridgeSource.UniswapV2]: fillData => {
-        let gas = 1.5e5;
-        if ((fillData as UniswapV2FillData).tokenAddressPath.length > 2) {
-            gas += 5e4;
+        // TODO: Different base cost if to/from ETH.
+        let gas = 100e3;
+        const path = (fillData as UniswapV2FillData).tokenAddressPath;
+        if (path.length > 2) {
+            gas += Math.max(0, path.length - 2) * 50e3; // +50k for each hop.
         }
         return gas;
     },
--- a/src/handlers/meta_transaction_handlers.ts
+++ b/src/handlers/meta_transaction_handlers.ts
@@ -66,6 +66,7 @@ export class MetaTransactionHandlers {
             buyAmount,
             slippagePercentage,
             excludedSources,
+            includedSources,
         } = parseGetTransactionRequestParams(req);
         try {
             const metaTransactionQuote = await this._metaTransactionService.calculateMetaTransactionQuoteAsync({
@@ -77,6 +78,7 @@ export class MetaTransactionHandlers {
                 from: takerAddress,
                 slippagePercentage,
                 excludedSources,
+                includedSources,
                 apiKey,
             });
             res.status(HttpStatus.OK).send(metaTransactionQuote);
@@ -136,6 +138,7 @@ export class MetaTransactionHandlers {
             buyAmount,
             slippagePercentage,
             excludedSources,
+            includedSources,
         } = parseGetTransactionRequestParams(req);
         try {
             const metaTransactionPrice = await this._metaTransactionService.calculateMetaTransactionPriceAsync(
@@ -148,6 +151,7 @@ export class MetaTransactionHandlers {
                     from: takerAddress,
                     slippagePercentage,
                     excludedSources,
+                    includedSources,
                     apiKey,
                 },
                 'price',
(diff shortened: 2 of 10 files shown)


src/config.ts
```diff
@@ -422,6 +423,9 @@ export const GAS_SCHEDULE_V1: FeeSchedule = {
     [ERC20BridgeSource.MultiHop]: fillData => {
         const firstHop = (fillData as MultiHopFillData).firstHopSource;
         const secondHop = (fillData as MultiHopFillData).secondHopSource;
+        if (!firstHop || !secondHop) {
```
dorothy-zbornak, line 426: This code path will get called even if `MultiHop` samples nothing now (ie, it's disabled), so we need this catch.

src/config.ts
```diff
@@ -350,6 +350,7 @@ export const GAS_SCHEDULE_V0: FeeSchedule = {
     [ERC20BridgeSource.Bancor]: () => 4.5e5,
     [ERC20BridgeSource.MStable]: () => 8.5e5,
     [ERC20BridgeSource.Mooniswap]: () => 3.5e5,
+    [ERC20BridgeSource.MultiHop]: () => 0,
```
dorothy-zbornak, line 353: This code path will get called even if `MultiHop` samples nothing now (ie, it's disabled), so we need this stub.

src/utils/parse_utils.ts
```diff
@@ -41,85 +44,82 @@ export const parseUtils = {
...
             return {
-                excludedSources: parseUtils.parseStringArrForERC20BridgeSources(request.excludedSources.split(',')),
+                excludedSources: parseUtils.parseStringArrForERC20BridgeSources(excludedIds),
+                includedSources: [],
                 nativeExclusivelyRFQT: false,
             };
         }
 
-        if (request.includedSources !== undefined) {
-            // Only RFQT is eligible as of now
```
dorothy-zbornak, line 53: So now `RFQT` can coexist with other sources in `includedSources`.

dekz (approved):

src/handlers/swap_handlers.ts
```diff
@@ -183,6 +183,10 @@ export class SwapHandlers {
                 req.query.excludedSources === undefined
                     ? []
                     : parseUtils.parseStringArrForERC20BridgeSources((req.query.excludedSources as string).split(',')),
+            includedSources:
+                req.query.includedSources === undefined
+                    ? []
+                    : parseUtils.parseStringArrForERC20BridgeSources((req.query.includedSources as string).split(',')),
```
PirosB3, line 189: `includedSources` can only be `RFQT`. Doesn't this added code mean that it could be any ERC20 source?

dorothy-zbornak: Yes. We now support an `includedSources` option in AS.

dorothy-zbornak: This change should still allow you to ask for RFQT only.

src/utils/parse_utils.ts
```diff
@@ -41,85 +44,82 @@ export const parseUtils = {
...
+                        field: 'intentOnFilling',
+                        code: ValidationErrorCodes.RequiredField,
+                        reason: ValidationErrorReasons.RequiresIntentOnFilling,
                     },
                 ]);
             }
+
+            return {
+                excludedSources: [],
+                includedSources: parseUtils.parseStringArrForERC20BridgeSources(['0x', ...includedIds]),
```
PirosB3, line 103: `RFQT` is not part of ERC20Source, will it get removed when `parseStringArrForERC20BridgeSources` is called?

src/utils/parse_utils.ts
```diff
@@ -41,85 +44,82 @@ export const parseUtils = {
...
+                nativeExclusivelyRFQT: !includedIds.includes('0x') && !includedIds.includes('Native'),
```
PirosB3, line 104: Could we simply make RFQT a special edge case here? for example, as of today there is no taker that requests exclusive RFQT liquidity alongside another on-chain source. If we feel good about making this assumption, we can add the following checks and statements in the code:

- If `RFQT` is added to `includedSources`, then `includedIds.length` must == 1
- `nativeExclusivelyRFQT: includedIds.includes("RFQT")`. So you could keep `L104` as `nativeExclusivelyRFQT: true`.

src/utils/parse_utils.ts
```diff
@@ -24,9 +24,12 @@ export const parseUtils = {
         request: ParseRequestForExcludedSourcesParams,
         validApiKeys: string[],
         endpoint: 'price' | 'quote',
-    ): { excludedSources: ERC20BridgeSource[]; nativeExclusivelyRFQT: boolean } {
+    ): { excludedSources: ERC20BridgeSource[]; includedSources: ERC20BridgeSource[]; nativeExclusivelyRFQT: boolean } {
```
PirosB3, line 27: This return type has grow a little, have you considered making it it's own type?

src/handlers/meta_transaction_handlers.ts
```diff
@@ -369,6 +373,10 @@ const parseGetTransactionRequestParams = (req: express.Request): GetTransactionR
         req.query.excludedSources === undefined
             ? undefined
             : parseUtils.parseStringArrForERC20BridgeSources((req.query.excludedSources as string).split(','));
+    const includedSources =
```
PirosB3, line 376: I'm really happy that you added this! thanks!

PirosB3 (changes requested):
Sorry for the delay! I ran a first review :)

src/utils/parse_utils.ts
dorothy-zbornak: Yes.

src/utils/parse_utils.ts
dorothy-zbornak: Idk, I could see simbot playing with a mix of sources and excluding open orderbook. The behavior for a taker passing in just `includedSources=RFQT` is still the same, right? What is the motivation for not allowing other included sources to coexist with RFQT?

dekz:
deploy staging

dekz:
deploy production

PR #365, opened 2020-09-30
chore: Upgrade TS and migrate to using strict mode

kimpers:
# Description
Our current TS configuration has disabled strict mode, which means that among other things TS will not error on accessing potentially undefined values. This makes it easy for run-time errors to slip out in production with code not properly handling an undefined property.

This proposal is a first step towards a fully strict mode TS codebase, it uses the non-null assertion operator (❗) to circumvent some of our current optional type mismatches. We can make a continuous effort to gradually remove these overrides as we tighten our internal typings. With this, at least the new code we write will have higher correctness 🤓 

# Testing Instructions

# Checklist

### DIFF
--- a/src/config.ts
+++ b/src/config.ts
@@ -6,6 +6,7 @@ import {
     DODOFillData,
     ERC20BridgeSource,
     FeeSchedule,
+    FillData,
     MultiHopFillData,
     OrderPrunerPermittedFeeTypes,
     RfqtMakerAssetOfferings,
@@ -381,9 +382,9 @@ export const GAS_SCHEDULE: FeeSchedule = {
         const firstHop = (fillData as MultiHopFillData).firstHopSource;
         const secondHop = (fillData as MultiHopFillData).secondHopSource;
         const firstHopGas =
-            GAS_SCHEDULE[firstHop.source] === undefined ? 0 : GAS_SCHEDULE[firstHop.source](firstHop.fillData);
+            GAS_SCHEDULE[firstHop.source] === undefined ? 0 : GAS_SCHEDULE[firstHop.source]!(firstHop.fillData);
         const secondHopGas =
-            GAS_SCHEDULE[secondHop.source] === undefined ? 0 : GAS_SCHEDULE[secondHop.source](secondHop.fillData);
+            GAS_SCHEDULE[secondHop.source] === undefined ? 0 : GAS_SCHEDULE[secondHop.source]!(secondHop.fillData);
         return new BigNumber(firstHopGas)
             .plus(secondHopGas)
             .plus(30e3)
@@ -402,8 +403,8 @@ const FEE_SCHEDULE: FeeSchedule = Object.assign(
     ...(Object.keys(GAS_SCHEDULE) as ERC20BridgeSource[]).map(k => ({
         [k]:
             k === ERC20BridgeSource.Native
-                ? fillData => PROTOCOL_FEE_MULTIPLIER.plus(GAS_SCHEDULE[k](fillData))
-                : fillData => GAS_SCHEDULE[k](fillData),
+                ? (fillData: FillData) => PROTOCOL_FEE_MULTIPLIER.plus(GAS_SCHEDULE[k]!(fillData))
+                : (fillData: FillData) => GAS_SCHEDULE[k]!(fillData),
     })),
 );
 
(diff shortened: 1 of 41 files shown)


tslint.json
```diff
@@ -1,6 +1,7 @@
 {
     "extends": ["@0x/tslint-config"],
     "rules": {
-        "custom-no-magic-numbers": [true, 0, 1, 2, 10]
+        "custom-no-magic-numbers": [true, 0, 1, 2, 10],
+        "no-unnecessary-type-assertion": false
```
kimpers, line 5: I had to abuse the non-null assertion operator (`!`) quite a bit in order to avoid having to do major refactors to fix our potentially undefined types sprinkled in several places. TSLint does not like this as it does not change the underlying type, just makes it non-nullable.

But at least we have a clear idea now where we are overriding potentially undefined properties as opposed to before where TS wasn't even checking that. In my opinion, this is strictly better, and we can make an effort to gradually clean this up.

test/rfqt_test.ts
```diff
@@ -137,7 +137,7 @@ describe(SUITE_NAME, () => {
                             {
                                 ...DEFAULT_RFQT_RESPONSE_DATA,
                                 responseData: { signedOrder },
-                            },
+                            } as any,
```
kimpers, line 140: We are just mocking a subset of the required data for the test so it's fine to coerce it to `any` here I think

opaolini:
deploy staging

src/config.ts
```diff
@@ -356,11 +357,23 @@ export const GAS_SCHEDULE_V0: FeeSchedule = {
 const FEE_SCHEDULE_V0: FeeSchedule = Object.assign(
     {},
     ...(Object.keys(GAS_SCHEDULE_V0) as ERC20BridgeSource[]).map(k => ({
-        [k]: fillData => PROTOCOL_FEE_MULTIPLIER.plus(GAS_SCHEDULE_V0[k](fillData)),
+        [k]: (fillData: FillData) => PROTOCOL_FEE_MULTIPLIER.plus(GAS_SCHEDULE_V0[k]!(fillData)),
     })),
 );
 
-export const ASSET_SWAPPER_MARKET_ORDERS_V0_OPTS: Partial<SwapQuoteRequestOpts> = {
+// NOTE: Need to explicitly specify which subset of SwapQuoteRequestOpts this will contain
```
dekz, line 364: damnit typescript get smarter

src/config.ts
```diff
@@ -429,9 +442,11 @@ export const GAS_SCHEDULE_V1: FeeSchedule = {
         const firstHop = (fillData as MultiHopFillData).firstHopSource;
         const secondHop = (fillData as MultiHopFillData).secondHopSource;
         const firstHopGas =
-            GAS_SCHEDULE_V1[firstHop.source] === undefined ? 0 : GAS_SCHEDULE_V1[firstHop.source](firstHop.fillData);
+            GAS_SCHEDULE_V1[firstHop.source] === undefined ? 0 : GAS_SCHEDULE_V1[firstHop.source]!(firstHop.fillData);
```
dekz, line 445: is this because it doesn't think `firstHop.source` is an `ERC20BridgeSource`? Or because the type of `GAS_SCHEDULE_V1` doesn't explicitly say it contains a values for all `ERC20BridgeSource`.

src/services/swap_service.ts
```diff
@@ -505,7 +507,8 @@ export class SwapService {
             includePriceComparisons,
         } = params;
         let _rfqt: RfqtRequestOpts | undefined;
-        const isAllExcluded = Object.values(ERC20BridgeSource).every(s => excludedSources.includes(s));
+        const isAllExcluded =
+            excludedSources && Object.values(ERC20BridgeSource).every(s => excludedSources.includes(s));
```
dekz, line 511: excludedSources can be undefined? I thought we set the defaults and it is defaulted to `[]`

dekz (approved):
I'm happy with this and aiming for tighter code. Need to find smart ways to inform typescript that certain values are definitely not undefined (like after passing a `utils.isNil` assertion the first time)

Thanks for taking this on

src/config.ts
kimpers: It seems like TS is not able to infer that `GAS_SCHEDULE_V1[firstHop.source]` will always be defined from the preceding ternary. I will update to a more TS friendly expression 👍

src/services/swap_service.ts
kimpers: We had some intermediate parameter type definitions where it was an optional parameter (even though it was always passed). I have now updated it to be non-nullable.

kimpers:
deploy staging

kimpers:
deploy production

PR #393, opened 2020-10-20
Fix bug when excludedSources=Native,Mesh

xianny:
# Description

Small bug with `Server Error` when `excludedSources=0x,Mesh`. [example request](https://api.matcha.0x.org/swap/v1/price?affiliateAddress=0x86003b044f70dac0abc80ac8957305b6370893ed&buyToken=ETH&excludedSources=0x,Mesh&includePriceComparisons=false&sellAmount=1000000000000000000&sellToken=0x6b175474e89094c44da98b954eedeac495271d0f&skipValidation=false&slippagePercentage=0.01)

[This line in parse utils](https://github.com/0xProject/0x-api/blob/master/src/utils/parse_utils.ts#L54-L59) sets `nativeExclusivelyRFQT=true` when excluded sources includes `Mesh`:
```
    if (excludedIds.length > 0) {
            return {
                excludedSources: parseUtils.parseStringArrForERC20BridgeSources(excludedIds),
                includedSources: [],
                // Exclude open orderbook if 'Mesh' is excluded.
                nativeExclusivelyRFQT: excludedIds.includes('Mesh'),
            };
        }
```

However, asset-swapper later throws an error when `nativeExclusivelyRFQT=true` AND native/0x source is excluded:
```
        // If RFQT is enabled and `nativeExclusivelyRFQT` is set, then `ERC20BridgeSource.Native` should
        // never be excluded.
        if (
            opts.rfqt &&
            opts.rfqt.nativeExclusivelyRFQT === true &&
            !sourceFilters.isAllowed(ERC20BridgeSource.Native)
        ) {
            throw new Error('Native liquidity cannot be excluded if "rfqt.nativeExclusivelyRFQT" is set');
        }
```
https://github.com/0xProject/0x-monorepo/blob/development/packages/asset-swapper/src/swap_quoter.ts#L658-L666

Solution: Modify parse utils to only set `nativeExclusivelyRFQT=true` when `Mesh` is excluded but `Native` is NOT.

# Testing Instructions

# Checklist

### DIFF
--- a/src/utils/parse_utils.ts
+++ b/src/utils/parse_utils.ts
@@ -56,7 +56,8 @@ export const parseUtils = {
                 excludedSources: parseUtils.parseStringArrForERC20BridgeSources(excludedIds),
                 includedSources: [],
                 // Exclude open orderbook if 'Mesh' is excluded.
-                nativeExclusivelyRFQT: excludedIds.includes('Mesh'),
+                nativeExclusivelyRFQT:
+                    excludedIds.includes('Mesh') && !excludedIds.includes('0x') && !excludedIds.includes('Native'),
             };
         }
 


dorothy-zbornak (approved):
Thanks for figuring this out!

PR #402, opened 2020-10-28
chore: Move to more stable 0x packages

dekz:
# Description

Bridge addresses are now in Asset-swapper, so there is little need to import a custom contract-addresses.
Order utils has been published and there is no need to use the private gitpkg version

# Testing Instructions

# Checklist

### DIFF
--- a/package.json
+++ b/package.json
@@ -120,22 +120,17 @@
         "tslint": "^6.1.3",
         "typescript": "^4.0.3"
     },
-    "resolutions": {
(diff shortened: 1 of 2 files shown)


dorothy-zbornak (approved):
:clap: :clap:

dorothy-zbornak:
PS, it looks like there was a `protocol` publish since your last commit so might need to regen lockfile.

PR #403, opened 2020-10-28
feat: support persistent orders for Matcha

xianny:
# Description

Support additional data for limit orders in Matcha, to provide a sense of continuity for users submitting orders through Matcha.

Currently, orders are ephemeral. SRA only returns orders that are "live" on Mesh, i.e. fillable. We want users on Matcha to be able to see their cancelled, filled, unfunded, and expired orders. This PR provides the ability to post "persistent" orders to a new endpoint, and retrieve them via the usual `GET /orders` endpoint. 

## Details

Adds `unfillable: boolean` query param in in `GET sra/orders`.
Adds a new endpoint: `POST sra/order/persistent` - protected by API key
Adds a new env variable: `SRA_PERSISTENT_ORDER_POSTING_WHITELISTED_API_KEYS`.
Adds a new table to database: `PersistentSignedOrderEntity`

When a user posts to `sra/order/persistent`, the order is posted to Matcha as usual, but also added to the database as a `PersistentSignedOrder`. 
Since the order is on the Matcha network, it will also be stored in the `SignedOrderEntity` table, like any other order. When Matcha event updates are received, both the `SignedOrder` and `PersistentOrder` table will be updated. 
There are a set of Matcha events that we consider "removal" events, which render an order unfillable. When we receive one of these events, the order will be removed from the `SignedOrder` table, but only updated in the `PersistentOrder` table.
The result is that regular orders are pruned as usual, but persistent orders stick around. 

This PR doesn't include a mechanism for removing persistent orders. We don't anticipate a large number of these (just Matcha limit orders for now). We still need to determine what the desired criteria for removal will be. 

To retrieve persistent orders, the client requests `GET /sra/orders?unfillable=true`. This will return all open orderbook orders (fillable) as well as any persistent orders that matcha the query params. 

The default behaviour if no argument is provided for `unfillable` is to return only fillable open orderbook orders (no change). 

Depends on https://github.com/0xProject/tools/pull/1

### POST /sra/orders/persistent
1. Post to mesh
2. Save to Persistent table
3. Sync mesh
4. Add to SignedOrders table

### Websocket Subscription
5. order event update comes in from Mesh
6. tag orders with appropriate lifecycle according to the event status
6. Accepted/Updated gets added to SRA (not Persistent)
7. Removed gets removed from SRA (not Persistent)
8. PersistentUpdated gets updated in Persistent (with open orderbook filtered out)

Example 1:
5. CANCEL comes in, removed from Mesh
6. Remove from SignedOrders table
7. Update in Persistent table

Example 2:
5. EXPIRED comes in, removed from Mesh
6. Remove from SignedOrders table
7. Update in Persistent table

### Mesh connection dropped, syncing orderbook
5. Post everything from SignedOrders table to Mesh because we missed mesh updates for a while
6. Receive Mesh response with Accepted and Rejected orders
7. Nothing happens to accepted
8. Remove Rejected from SignedOrders table
9. Update Rejected in Persistent table, using the Rejected reason (filter out open orderbook orders)
No more updates to Rejected from Mesh, stays in Persistent forever

# Testing Instructions

# Checklist

-   [ ] Create new table for `persistent_signed_orders`


xianny:
deploy staging

src/entities/PersistentSignedOrderEntity.ts
```diff
@@ -0,0 +1,93 @@
...
+            makerFee?: string;
+            takerFee?: string;
+            makerFeeAssetData?: string;
+            takerFeeAssetData?: string;
+            makerAssetAmount?: string;
+            takerAssetAmount?: string;
+            salt?: string;
+            signature?: string;
+            remainingFillableTakerAssetAmount?: string;
+            orderState?: OrderEventEndState;
```
kimpers, line 87: Do all of these attributes need to be optional? Wouldn't most of them be required for a valid order?

If we can tighten up the model schema we can avoid needing to check for null/undefined whenever using a property on the model.

src/services/order_watcher_service.ts
```diff
@@ -116,6 +127,42 @@ export class OrderWatcherService {
...
+                        orderState,
+                    });
+                });
+
+                // 3. Wait for results
+                await Promise.allSettled(updatePromises).then(results => {
+                    // Group by success or failure. Open orderbook orders should fail.
+                    const { fulfilled, rejected } = results.reduce(
+                        (acc, r, i) => {
+                            r.status === 'fulfilled' ? acc.fulfilled.push(i) : acc.rejected.push(i);
```
kimpers, line 154: NIT(feel free to ignore): maybe a bit more readable to store the object rather than the index the object has in the array?

Could even be a neat one-liner 😄 
```typescript
const [fulfilled, rejected] = _.partition(results, result => r.status === 'fulfilled');
```

src/utils/mesh_utils.ts
```diff
@@ -69,25 +87,42 @@ export const meshUtils = {
...
+                    persistentUpdated.push(apiOrder);
+                    break;
                 case OrderEventEndState.StoppedWatching:
+                    removed.push(apiOrder);
+                    persistentUpdated.push(apiOrder);
+                    break;
                 case OrderEventEndState.Unfunded: {
                     removed.push(apiOrder);
+                    persistentUpdated.push(apiOrder);
                     break;
```
kimpers, line 112: NIT(just my opinon feel free to ignore 😄 ): since we are doing the same thing in all the case statements I would just let it fall through to the bottom case so we don't have to make changes in many different places in the future
```suggestion
                case OrderEventEndState.Invalid:
                case OrderEventEndState.Cancelled:
                case OrderEventEndState.Expired:
                case OrderEventEndState.FullyFilled:
                case OrderEventEndState.StoppedWatching:
                case OrderEventEndState.Unfunded: {
                    removed.push(apiOrder);
                    persistentUpdated.push(apiOrder);
                    break;
```

tsconfig.json
```diff
@@ -1,8 +1,8 @@
 {
     "compilerOptions": {
         "module": "commonjs",
-        "lib": ["es2017", "dom"],
-        "target": "es6",
+        "lib": ["es2020", "dom"],
+        "target": "es2019",
```
kimpers, line 5: Oskar will be so happy about this 🎉  No more transpiling promises to generator functions 🙏  

But do we not need to bump the node version in the Dockerfile for this to work in prod?

src/entities/PersistentSignedOrderEntity.ts
```diff
@@ -0,0 +1,93 @@
+import { OrderEventEndState } from '@0x/mesh-rpc-client';
```
kimpers, line 1: Given that we have `synchronize` disabled in production I think we'd need to create a migration for this Entity in order for it to take effect in prod, or am I missing something? 🤔

kimpers (changes requested):

test/orderbook_service_test.ts
```diff
@@ -0,0 +1,226 @@
...
+        metaData: {
+            orderHash: orderHashUtils.getOrderHash(order),
+            remainingFillableTakerAssetAmount: remainingFillableAssetAmount || order.takerAssetAmount,
+            state: OrderEventEndState.Added,
+        },
+    };
+    return apiOrder;
+}
+
+describe.only(SUITE_NAME, () => {
```
kimpers, line 76: Just a reminder to remove the `only` before merging 👍 
```suggestion
describe(SUITE_NAME, () => {
```

src/entities/PersistentSignedOrderEntity.ts
xianny: The SignedOrder entity also has all optional fields. I think this is to guard against write errors.

src/utils/mesh_utils.ts
xianny: good point! changing

tsconfig.json
xianny: The image we're using [node:12-alpine](https://hub.docker.com/layers/node/library/node/12-alpine/images/sha256-c29fba1f3ea0c23ca3ab59367dc644a594c8186fe28067ea4ac7e51b4a0956a1?context=explore) was upgraded to node version 12.19.0,[ it looks like >12.9.0 supports es2020.](https://stackoverflow.com/questions/59787574/typescript-tsconfig-settings-for-node-js-12)

src/entities/PersistentSignedOrderEntity.ts
xianny: We used `synchronize` to add the last table and it was meh. I'll create a migration file.

test/orderbook_service_test.ts
xianny: thanks, good catch!

kimpers (approved):
Awesome work on this! 👍

xianny:
deploy staging

xianny:
deploy staging

src/routers/sra_router.ts
```diff
@@ -45,6 +45,12 @@ export function createSRARouter(orderBook: OrderBookService): express.Router {
      * This is an additional endpoint not a part of the official SRA standard
      */
     router.post('/orders', asyncHandler(handlers.postOrdersAsync.bind(handlers)));
+    /**
+     * POST Persistent order endpoint submits an order that will be persisted even after cancellation or expiration.
+     * Requires an API Key
+     * This is an additional endpoint not a part of the official SRA standard
+     */
+    router.post('/order/persistent', asyncHandler(handlers.postPersistentOrderAsync.bind(handlers)));
```
dekz, line 53: what are your thoughts on this custom endpoint versus just using `POST /order` with an API Key, iff there is an API key we also save it to `PersistentOrders`.

migrations/1604516429383-CreatePersistentSignedOrder.ts
```diff
@@ -0,0 +1,42 @@
+import { OrderEventEndState } from '@0x/mesh-rpc-client';
+import { MigrationInterface, QueryRunner, Table } from 'typeorm';
+
+const OrderEventEndStateStrings = Object.keys(OrderEventEndState).filter(x => isNaN(parseInt(x, 10)));
+
+export class CreatePersistentSignedOrder1604516429383 implements MigrationInterface {
+    // tslint:disable-next-line
+    public async up(queryRunner: QueryRunner): Promise<void> {
```
dekz, line 8: do we also create the indexes here?
Everytime we ran with `synchronize` true our indexes would be deleted >< I guess since they weren't explicitly defined.

But we'd want indexes on:
```
maker_address
maker_asset_data
taker_asset_data
fee_recipient_address
```

src/services/orderbook_service.ts
```diff
@@ -128,8 +133,31 @@ export class OrderBookService {
         const { fresh, expired } = orderUtils.groupByFreshness(apiOrders, SRA_ORDER_EXPIRATION_BUFFER_SECONDS);
         alertOnExpiredOrders(expired);
 
-        // Post-filters
-        const filteredApiOrders = orderUtils.filterOrders(fresh, ordersFilterParams);
+        // Join with persistent orders
+        let persistentOrders: APIOrderWithMetaData[] = [];
+        if (ordersFilterParams.unfillable === true) {
```
dekz, line 138: My only thought here is we want to avoid sending too much data, to prevent ourselves from being taken down.

We might want to enforce that `unfillable` can only be used in conjunction when `maker_address`  is filtered as well.

Or what ever the most common behaviour is expected.

I have concern of just an `GET /sra/orders?unfillable=true` returning large payloads.

dekz (changes requested):

xianny:
deploy staging

xianny:
deploy staging

xianny:
deploy staging

xianny:
deploy staging

xianny:
deploy staging

xianny:
deploy staging

xianny:
deploy staging

src/services/orderbook_service.ts
xianny: doneskiii

xianny:
deploy staging

xianny:
deploy staging

xianny:
deploy production

PR #427, opened 2020-11-23
fix: handling mesh order event end state

xianny:
# Description

Fixes a bug in parsing mesh order events and rejection codes. Previously would occasionally return an undefined order state, which results in an uncaught error when saving to the database. 

Also includes a bit of a refactor. Trying to make the order watcher service more readable.
# Testing Instructions

# Checklist

    -   [ ] SRA/Limit orders
    -   [ ] Swap endpoints
    -   [ ] Depth charts

### DIFF
--- a/src/utils/mesh_utils.ts
+++ b/src/utils/mesh_utils.ts
@@ -23,20 +23,25 @@ export const meshUtils = {
     orderInfoToAPIOrder: (
         orderEvent: OrderEvent | AcceptedOrderInfo | RejectedOrderInfo | OrderInfo,
     ): APIOrderWithMetaData => {
-        const remainingFillableTakerAssetAmount = (orderEvent as OrderEvent).fillableTakerAssetAmount
-            ? (orderEvent as OrderEvent).fillableTakerAssetAmount
-            : ZERO;
-        return {
+        const { orderHash, fillableTakerAssetAmount, signedOrder } = orderEvent as OrderEvent;
+        const apiOrder: APIOrderWithMetaData = {
             // orderEvent.signedOrder comes from mesh with string fields, needs to be serialized into SignedOrder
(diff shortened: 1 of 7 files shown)


xianny:
deploy staging

xianny:
Tested this out on staging vs production. 

Expired order disappeared from staging after a minute, but stayed on production.

xianny:
deploy staging

src/utils/mesh_utils.ts
```diff
@@ -23,20 +23,25 @@ export const meshUtils = {
...
+        // populate order state
+        apiOrder.metaData.state = (orderEvent as OrderEvent).endState;
+        if (apiOrder.metaData.state === undefined) {
+            const r = orderEvent as RejectedOrderInfo;
+            apiOrder.metaData.state =
+                r.status && r.status.code
+                    ? meshUtils.rejectedCodeToOrderState(r.status.code)
+                    : OrderEventEndState.Invalid;
+        }
+        return apiOrder;
```
xianny, line 44: This is the new logic. Previously would sometimes return undefined and then throw an error when saving to db.

src/utils/mesh_utils.ts
```diff
@@ -92,22 +95,20 @@ export const meshUtils = {
...
-                    removed.push(apiOrder);
-                    persistentUpdated.push(apiOrder);
+                    removed.push(order);
                     break;
                 }
                 case OrderEventEndState.Unexpired:
                 case OrderEventEndState.FillabilityIncreased:
                 case OrderEventEndState.Filled: {
-                    updated.push(apiOrder);
-                    persistentUpdated.push(apiOrder);
```
xianny, line 103: Instead of including persistent orders here, which Mesh doesn't know about, we'll handle the db update logic exclusively in OrderWatcherService.

kimpers (approved):
Nice work! 👍

PR #432, opened 2020-11-25
fix: Postgres Firm Quote Validator

PirosB3:
Implements the backend process that performs soft validation of firm quotes. This code is to be run during an API request and it should implement the `RfqtFirmQuoteValidator`interface.


src/constants.ts
```diff
@@ -86,3 +86,10 @@ export const MARKET_DEPTH_END_PRICE_SLIPPAGE_PERC = 20;
 
 // Logging
 export const NUMBER_SOURCES_PER_LOG_LINE = 12;
+
+// RFQ Quote Validator expiration threshold
+
+// NOTE: Until worker component is created, we set an infinite expiry
+// rate which will allow us to start writing out records to the cache
+// table without causing all quotes to be invalid.
+export const RFQ_FIRM_QUOTE_CACHE_EXPIRY = Number.POSITIVE_INFINITY;
```
BMillman19, line 95: Is this safe?

src/services/firm_quote_validator.ts
```diff
@@ -0,0 +1,147 @@
...
+        };
+    }
+
+    constructor(chainCacheRepository: Repository<MakerBalanceChainCacheEntity>, cacheExpiryThresholdMs: number = THRESHOLD_CACHE_EXPIRED_MS) {
+        this._chainCacheRepository = chainCacheRepository;
+        this._cacheExpiryThresholdMs = cacheExpiryThresholdMs;
+    }
+
+    // tslint:disable-next-line: prefer-function-over-method
+    public async getRFQTTakerFillableAmounts(quotes: SignedOrder[]): Promise<BigNumber[]> {
```
BMillman19, line 61: sometimes we use `RFQT` in all caps and sometimes we use `Rfqt`

BMillman19: Also, don't we usually add `Async` as a suffix to async methods? maybe the linter doesnt catch that anymore..

src/services/firm_quote_validator.ts
```diff
@@ -0,0 +1,147 @@
...
+        // TODO: Handle error on query
+
+        // Ensure that all quotes have the same exact maker token.
+        const makerTokenAddressesSet = new Set(
+            quotes.map(quote => {
+                const decodedAssetData = assetDataUtils.decodeAssetDataOrThrow(quote.makerAssetData) as ERC20AssetData;
+                return decodedAssetData.tokenAddress;
+            })
+        );
+        const makerTokenAddresses = Array.from(makerTokenAddressesSet);
```
BMillman19, line 71: do we need to convert to an array here? can we just use the `.size()` accessor on the set?

src/services/firm_quote_validator.ts
```diff
@@ -0,0 +1,147 @@
...
+    }
+
+    // tslint:disable-next-line: prefer-function-over-method
+    public async getRFQTTakerFillableAmounts(quotes: SignedOrder[]): Promise<BigNumber[]> {
+        // TODO: Handle error on query
+
+        // Ensure that all quotes have the same exact maker token.
+        const makerTokenAddressesSet = new Set(
+            quotes.map(quote => {
+                const decodedAssetData = assetDataUtils.decodeAssetDataOrThrow(quote.makerAssetData) as ERC20AssetData;
```
BMillman19, line 67: Do we want to throw here or fail gracefully i.e. return `quotes.map(_quote => BIG_NUMBER_ZERO)`?

src/services/firm_quote_validator.ts
```diff
@@ -0,0 +1,147 @@
...
+            return quotes.map(_quote => BIG_NUMBER_ZERO);
+        }
+        const makerTokenAddress = makerTokenAddresses[0];
+
+        // Fetch balances and create a lookup table
+        const makerLookup: {[key: string]: BigNumber} = {};
+        const makerAddresses = Array.from(new Set(quotes.map(quote => quote.makerAddress)));
+        const cacheResults = await this._chainCacheRepository.find({
+            where: [{
+                tokenAddress: makerTokenAddresses[0],
```
BMillman19, line 83: we already pull out the first element above and name it `makerTokenAddress`

src/services/firm_quote_validator.ts
```diff
@@ -0,0 +1,147 @@
...
+        const makerTokenAddresses = Array.from(makerTokenAddressesSet);
+        if (makerTokenAddresses.length !== 1) {
+            logger.error(`Quotes array was empty or found multiple maker token addresses within one single RFQ batch: ${JSON.stringify(makerTokenAddresses)}. Rejecting the batch`);
+            return quotes.map(_quote => BIG_NUMBER_ZERO);
+        }
+        const makerTokenAddress = makerTokenAddresses[0];
+
+        // Fetch balances and create a lookup table
+        const makerLookup: {[key: string]: BigNumber} = {};
+        const makerAddresses = Array.from(new Set(quotes.map(quote => quote.makerAddress)));
```
BMillman19, line 80: at first its a bit confusing to see the conversions between Set and Array, maybe we can just add a quick comment about how this step is to ensure `makerAddresses` has only unique addresses

src/services/firm_quote_validator.ts
```diff
@@ -0,0 +1,147 @@
...
+        if (makerAddressesToAddToCache.length > 0) {
+            logger.info(`Adding new addresses to cache: ${JSON.stringify(makerAddressesToAddToCache)}`);
+            await this._chainCacheRepository
+                .createQueryBuilder()
+                .insert()
+                .values(
+                    makerAddressesToAddToCache.map(makerAddress => {
+                        return {
+                            makerAddress,
+                            tokenAddress: makerTokenAddresses[0],
```
BMillman19, line 108: just use `makerTokenAddress` here

src/services/firm_quote_validator.ts
```diff
@@ -0,0 +1,147 @@
...
+        }
+        return takerFillableAmounts;
+    }
+
+    private _calculateMakerBalanceFromResult(result: MakerBalanceChainCacheEntity, makerTokenAddress: string, nowUnix: number): BigNumber {
+        if (!result.timeOfSample) {
+            // If a record exists but a time of sample does not yet exist, this means that the cache entry has not yet been
+            // populated by the worker process. This may be due to a new address being added a few minutes ago, but it could
+            // also be due to a bug in the worker.
+            const timeFirstSeen = result.timeFirstSeen ? result.timeFirstSeen.getTime() : 0;
```
BMillman19, line 124: is this in ms?

migrations/1606305812748-AddMakerBalanceCacheIndex.ts
```diff
@@ -0,0 +1,13 @@
+import { MigrationInterface, QueryRunner } from 'typeorm';
+
+export class AddMakerBalanceCacheIndex1606305812748 implements MigrationInterface {
```
alexkroeger, line 3: Are you sure this index is necessary? Those are the primary key fields, so it shouldn't scan the whole table if you specify both.

src/services/firm_quote_validator.ts
```diff
@@ -0,0 +1,147 @@
...
+                return quote.takerAssetAmount;
+            }
+
+            // Order is fully fillable, because Maker has 100% of the assets
+            if (makerTokenBalanceForMaker.gte(quote.makerAssetAmount)) {
+                return quote.takerAssetAmount;
+            }
+
+            // Order is empty, return zero
+            if (quote.makerAssetAmount.lte(0)) {
```
BMillman19, line 36: should this be `makerTokenBalanceForMaker.lte(0)` ?

PirosB3: no this is checking that `makerAssetAmount` is greater than 0 because if not it will raise a "divide by 0" error on L41. If `makerTokenBalanceForMaker` is 0 then `partialFillableAmount` will gracefully become 0

src/services/firm_quote_validator.ts
PirosB3: oh lol somehow I missed that

src/services/firm_quote_validator.ts
```diff
@@ -0,0 +1,147 @@
...
+import { In } from 'typeorm';
+import { Repository } from 'typeorm/repository/Repository';
+
+import { ONE_SECOND_MS } from '../constants';
+import { MakerBalanceChainCacheEntity } from '../entities/MakerBalanceChainCacheEntity';
+import { logger } from '../logger';
+
+
+const BIG_NUMBER_ZERO = new BigNumber(0);
+const THRESHOLD_CACHE_EXPIRED_MS = 2 * 60 * ONE_SECOND_MS;
```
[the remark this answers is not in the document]
PirosB3: done!

src/services/firm_quote_validator.ts
PirosB3: Good question, if ERC721 orders are passing through RFQT, I would rather raise an error that will bubble up. This should straight up never happen, what do you think?

src/services/firm_quote_validator.ts
PirosB3: yes it is, do you think I should specify in the naming (`timeOfFirstSeenMs`?)

src/services/firm_quote_validator.ts
```diff
@@ -0,0 +1,147 @@
+import { BigNumber, RfqtFirmQuoteValidator, SignedOrder } from '@0x/asset-swapper';
```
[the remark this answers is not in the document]
PirosB3: great ideal, will do the rename in a separate commit

src/services/firm_quote_validator.ts
PirosB3: Thinking about this more, it really doesn't matter to the query if duplicate address entries are present. If you think of the SQL that this query produces, it's simply a list of addresses in an `IN` clause. If the addresses in that `IN` clause are duplicate, there is no effect.

Would you rather I remote the `Set` stuff, given that it would simplify the code and results would be the same?

src/services/firm_quote_validator.ts
```diff
@@ -0,0 +1,147 @@
...
+import { assetDataUtils, ERC20AssetData } from '@0x/order-utils';
+import { In } from 'typeorm';
+import { Repository } from 'typeorm/repository/Repository';
+
+import { ONE_SECOND_MS } from '../constants';
+import { MakerBalanceChainCacheEntity } from '../entities/MakerBalanceChainCacheEntity';
+import { logger } from '../logger';
+
+
+const BIG_NUMBER_ZERO = new BigNumber(0);
```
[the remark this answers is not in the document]
PirosB3: done

src/services/firm_quote_validator.ts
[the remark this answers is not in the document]
PirosB3: done

src/services/firm_quote_validator.ts
```diff
@@ -0,0 +1,147 @@
...
+const THRESHOLD_CACHE_EXPIRED_MS = 2 * 60 * ONE_SECOND_MS;
+
+
+export class PostgresBackedFirmQuoteValidator implements RfqtFirmQuoteValidator {
```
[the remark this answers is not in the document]
PirosB3: made naming consistent

src/app.ts
```diff
@@ -254,13 +262,14 @@ function createMetaTransactionRateLimiterFromConfig(
  */
 export function createSwapServiceFromOrderBookService(
     orderBookService: OrderBookService,
+    rfqFirmQuoteValidator: RfqtFirmQuoteValidator,
```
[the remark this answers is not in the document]
PirosB3: done

src/services/firm_quote_validator.ts
PirosB3: done

src/services/firm_quote_validator.ts
PirosB3: done

migrations/1606305812748-AddMakerBalanceCacheIndex.ts
PirosB3: That's a great thought, let me remove the index and try running thr query

PirosB3: https://gist.github.com/PirosB3/1548153af1e5b3939498cf42e44aef62

Sir, you are totally right! the `EXPLAIN `cost is the same.

```sql
api=# EXPLAIN SELECT * FROM maker_balance_chain_cache WHERE token_address = '0xfoo' AND maker_address IN ('a', 'b', 'c');
                                          QUERY PLAN
-----------------------------------------------------------------------------------------------
 Bitmap Heap Scan on maker_balance_chain_cache  (cost=4.17..11.29 rows=1 width=112)
   Recheck Cond: ((token_address)::text = '0xfoo'::text)
   Filter: ((maker_address)::text = ANY ('{a,b,c}'::text[]))
   ->  Bitmap Index Scan on "PK_13d68ce2e19b661c60a8499808c"  (cost=0.00..4.17 rows=3 width=0)
         Index Cond: ((token_address)::text = '0xfoo'::text)
(5 rows)
```

Will drop the new index

PirosB3: @alexkroeger maybe this is irrelevant and it's over-optimizing (not a PG expert), but I'm surprised that the query that I run only partially hits the index.

```
api=# EXPLAIN SELECT * FROM maker_balance_chain_cache WHERE token_address = '0xfoo' AND maker_address IN ('a', 'b', 'c') AND maker_address IS NOT NULL;
                                                   QUERY PLAN
----------------------------------------------------------------------------------------------------------------
 Index Scan using maker_balance_chain_cache_idx on maker_balance_chain_cache  (cost=0.28..8.30 rows=1 width=56)
   Index Cond: (((token_address)::text = '0xfoo'::text) AND (maker_address IS NOT NULL))
   Filter: ((maker_address)::text = ANY ('{a,b,c}'::text[]))
(3 rows)
```

Is there any reason why PG does a linear scan (Which is what I interpret as a `Filter`) after indexing on `token_address`?

src/constants.ts
PirosB3: It is, because of [this](https://github.com/0xProject/0x-api/pull/432/files#diff-fa95b01194bc9cd84720be5a299890448459cfb7c4b7faa452681c9f11f65907R42) safety pin

PirosB3 (commented):
Ready for another round of reviews

migrations/1606305812748-AddMakerBalanceCacheIndex.ts
alexkroeger: I think the `maker_address is not null` may be throwing it off. Primary keys can't be null so you can get rid of that.

package.json
```diff
@@ -125,7 +125,7 @@
     },
     "dependencies": {
         "@0x/assert": "^3.0.17",
-        "@0x/asset-swapper": "^5.2.0",
+        "@0x/asset-swapper": "0xProject/gitpkg-registry#0x-asset-swapper-v5.2.0-0c08353b2",
```
alexkroeger, line 128: Can you link to the asset-swapper changes? was it just the naming change?

PirosB3: Yes indeed! https://github.com/0xProject/protocol/pull/63

migrations/1606305812748-AddMakerBalanceCacheIndex.ts
PirosB3: oh sorry, I added that to see if I could fool the planner to use the index. The query plan is (nearly) the same when running: `SELECT * FROM maker_balance_chain_cache WHERE token_address = '0xfoo' AND maker_address IN ('a', 'b', 'c')`

src/services/postgres_rfqt_firm_quote_validator.ts
```diff
@@ -0,0 +1,188 @@
...
+    ) {
+        this._chainCacheRepository = chainCacheRepository;
+        this._cacheExpiryThresholdMs = cacheExpiryThresholdMs;
+    }
+
+    // tslint:disable-next-line: prefer-function-over-method
+    public async getRfqtTakerFillableAmountsAsync(quotes: SignedOrder[]): Promise<BigNumber[]> {
+        // TODO: Handle error on query
+
+        // Ensure that all quotes have the same exact maker token.
```
alexkroeger, line 77: Is it necessary to enforce this? Maybe we'll use RFQT in multi-hop at some point and will request different tokens

src/services/postgres_rfqt_firm_quote_validator.ts
```diff
@@ -0,0 +1,188 @@
...
+        makerLookup: { [key: string]: BigNumber },
+    ): { makerAddressesToAddToCache: string[]; takerFillableAmounts: BigNumber[] } {
+        const makerAddressesToAddToCacheSet: Set<string> = new Set();
+        const takerFillableAmounts = quotes.map(quote => {
+            const makerTokenBalanceForMaker: BigNumber | undefined = makerLookup[quote.makerAddress];
+
+            // TODO: Add Prometheus hooks
+            if (makerTokenBalanceForMaker === undefined) {
+                makerAddressesToAddToCacheSet.add(quote.makerAddress);
+                return quote.takerAssetAmount;
```
alexkroeger, line 27: is it worth taking the hit and pulling the balance on-chain?

src/services/postgres_rfqt_firm_quote_validator.ts
```diff
@@ -0,0 +1,188 @@
...
+    private readonly _chainCacheRepository: Repository<MakerBalanceChainCacheEntity>;
+    private readonly _cacheExpiryThresholdMs: number;
+
+    private static _calculateTakerFillableAmountsFromQuotes(
+        quotes: SignedOrder[],
+        makerLookup: { [key: string]: BigNumber },
+    ): { makerAddressesToAddToCache: string[]; takerFillableAmounts: BigNumber[] } {
+        const makerAddressesToAddToCacheSet: Set<string> = new Set();
+        const takerFillableAmounts = quotes.map(quote => {
+            const makerTokenBalanceForMaker: BigNumber | undefined = makerLookup[quote.makerAddress];
```
alexkroeger, line 22: What happens if balance is `null`?

src/services/postgres_rfqt_firm_quote_validator.ts
```diff
@@ -0,0 +1,188 @@
...
+            // Order is empty, return zero
+            if (quote.makerAssetAmount.lte(0)) {
+                return ZERO;
+            }
+
+            // Order is partially fillable, because Maker has a fraction of the assets
+            const partialFillableAmount = makerTokenBalanceForMaker
+                .times(quote.takerAssetAmount)
+                .div(quote.makerAssetAmount)
+                .integerValue(BigNumber.ROUND_DOWN);
```
alexkroeger, line 44: Will this round to an even integer?

src/services/postgres_rfqt_firm_quote_validator.ts
```diff
@@ -0,0 +1,188 @@
...
+        // Finally, adjust takerFillableAmount based on maker balances
+        const {
+            makerAddressesToAddToCache,
+            takerFillableAmounts,
+        } = PostgresRfqtFirmQuoteValidator._calculateTakerFillableAmountsFromQuotes(quotes, makerLookup);
+
+        // If any new addresses were found, add new addresses to cache.
+        // NOTE: since this insertion happens on the web processes, we need to gracefully handle conflict
+        // that can happen if two threads try to insert the same entry at the same time. This is why we add
+        // the "ON CONFLICT" clause.
```
alexkroeger, line 124: nice 👍

src/services/postgres_rfqt_firm_quote_validator.ts
```diff
@@ -0,0 +1,188 @@
...
+                return new BigNumber(Number.POSITIVE_INFINITY);
+            }
+        } else if (nowUnix - result.timeOfSample.getTime() > this._cacheExpiryThresholdMs) {
+            // In this case a cache entry exists, but it's simply too old and this should never really happen unless the worker is stuck.
+            logger.error(
+                `Cache entry for maker ${result.makerAddress} and token ${
+                    result.tokenAddress
+                } was last refreshed on ${result.timeOfSample.getTime()} which is more than ${
+                    this._cacheExpiryThresholdMs
+                } ms ago. Assuming worker is stuck and setting maker balance to 0.`,
```
alexkroeger, line 174: This seems like a harsh assumption. Maybe we want to assume it's fillable in that case?

PirosB3: discussed

src/services/postgres_rfqt_firm_quote_validator.ts
PirosB3: discussed

src/services/postgres_rfqt_firm_quote_validator.ts
PirosB3: discussed, will talk to Blake on Monday

src/constants.ts
PirosB3, line 95: Will not merge until [Kroeger's PR](https://github.com/0xProject/0x-api/pull/434) is merged and runs smoothly in prod

alexkroeger (approved):

src/services/postgres_rfqt_firm_quote_validator.ts
```diff
@@ -142,11 +135,65 @@ export class PostgresRfqtFirmQuoteValidator implements RfqtFirmQuoteValidator {
...
+
+            // Order is fully fillable, because Maker has 100% of the assets
+            if (makerTokenBalanceForMaker.gte(quote.makerAssetAmount)) {
+                ORDER_FULLY_FILLABLE.labels(this._workerId);
+                return quote.takerAssetAmount;
+            }
+
+            // Order is empty, return zero
+            if (quote.makerAssetAmount.lte(0)) {
+                ORDER_NOT_FILLABLE.labels(this._workerId).inc();
```
alexkroeger, line 161: do all the counters need this `inc()` call?

PirosB3: yes

PirosB3: timers have `set()`, counters have `inc()`

alexkroeger: it's missing on two of the prometheus calls above ^

src/services/postgres_rfqt_firm_quote_validator.ts
```diff
@@ -142,11 +135,65 @@ export class PostgresRfqtFirmQuoteValidator implements RfqtFirmQuoteValidator {
...
+        makerLookup: { [key: string]: BigNumber },
+    ): { makerAddressesToAddToCache: string[]; takerFillableAmounts: BigNumber[] } {
+        const makerAddressesToAddToCacheSet: Set<string> = new Set();
+        const takerFillableAmounts = quotes.map(quote => {
+            const makerTokenBalanceForMaker: BigNumber | undefined = makerLookup[quote.makerAddress];
+
+            // TODO: Add Prometheus hooks
+            if (makerTokenBalanceForMaker === undefined) {
+                makerAddressesToAddToCacheSet.add(quote.makerAddress);
+                ORDER_FULLY_FILLABLE.labels(this._workerId);
```
alexkroeger, line 149: should this be a separate counter? it's more like `not validated`

src/runners/rfq_maker_balance_cache_runner.ts
```diff
@@ -19,7 +19,7 @@ import { createResultCache, ResultCache } from '../utils/result_cache';
 const DELAY_WHEN_NEW_BLOCK_FOUND = ONE_SECOND_MS * 5;
 const DELAY_WHEN_NEW_BLOCK_NOT_FOUND = ONE_SECOND_MS;
 // tslint:disable-next-line:custom-no-magic-numbers
-const CACHE_MAKER_TOKENS_FOR_MS = ONE_SECOND_MS * 30;
+const CACHE_MAKER_TOKENS_FOR_MS = Math.floor(RFQ_FIRM_QUOTE_CACHE_EXPIRY / 30);
```
alexkroeger, line 22: / 4

src/services/postgres_rfqt_firm_quote_validator.ts
alexkroeger: also missing inc()

src/services/postgres_rfqt_firm_quote_validator.ts
```diff
@@ -0,0 +1,237 @@
...
+                return quote.takerAssetAmount;
+            }
+
+            // Order is fully fillable, because Maker has 100% of the assets
+            if (makerTokenBalanceForMaker.gte(quote.makerAssetAmount)) {
+                ORDER_FULLY_FILLABLE.labels(this._workerId);
```
alexkroeger, line 155: missing `inc()`

src/runners/rfq_maker_balance_cache_runner.ts
PirosB3: fixed

src/services/postgres_rfqt_firm_quote_validator.ts
PirosB3: fixed

PR #433, opened 2020-11-25
fix: price comparisons gas estimations with ETH to token rates

kimpers:
# Description
This fixes so that price comparisons take into account gas costs when picking the best sample for each sampled provider. Additionally it adds the fields to `ethToInputRate` and `ethToOutputRate` the responses of the `/quote` and `/price` endpoints. This was previously a pre-requisite for calculating savings, but now it may not be necessary. However, the fields could still prove useful for future data explorations. Maybe we should keep them in the responses, wdyt? 

# Testing Instructions

# Checklist

    -   [ ] SRA/Limit orders
    -   [ ] Swap endpoints
    -   [ ] Depth charts


src/types.ts
```diff
@@ -408,6 +408,8 @@ export interface GetSwapQuoteResponse extends SwapQuoteResponsePartialTransactio
     from?: string;
     gas: BigNumber;
     estimatedGas: BigNumber;
+    ethToInputRate: BigNumber;
+    ethToOutputRate: BigNumber;
```
kimpers, line 412: TODO: should this be `null` when rate is 0?

src/handlers/meta_transaction_handlers.ts
```diff
@@ -242,6 +242,8 @@ export class MetaTransactionHandlers {
                 protocolFee: metaTransactionPrice.protocolFee,
                 minimumProtocolFee: metaTransactionPrice.minimumProtocolFee,
                 allowanceTarget: metaTransactionPrice.allowanceTarget,
+                ethToInputRate: metaTransactionPrice.ethToInputRate,
+                ethToOutputRate: metaTransactionPrice.ethToOutputRate,
```
xianny, line 246: Can we rename to `sellTokenToEth` and `buyTokenToEth` for consistency? (or eth to x), I think we'll probably forget what input/output is supposed to mean

xianny (commented):
The math looks good to me! Would just love to disambiguate the ethToX variable names.

xianny (approved):

src/handlers/meta_transaction_handlers.ts
kimpers: Yeah, that's a great idea. Will update tomorrow 👍

src/utils/price_comparison_utils.ts
```diff
@@ -118,8 +113,57 @@ export const priceComparisonUtils = {
...
+            const ethUnitAmount = new BigNumber(10).pow(ethToken.decimals);
+            // NOTE: Sort sources by the best outcome for the user
+            // if the user is selling we want to maximize the maker amount they will receive
+            // if the user is buying we want to minimize the taker amount they have to pay
+            const sortedSources = isSelling
+                ? tradeSourcesWithGas.slice().sort((a, b) => {
+                      const aGasCostMakerAssetUnitAmount = a.gas
+                          .times(quote.gasPrice)
+                          .dividedBy(ethUnitAmount)
+                          .times(quote.ethToBuyTokenRate);
```
xianny, line 136: Hey, should this be `dividedBy` instead of `times`?

Example:

1 eth = 3 buyTokens
ethToBuyTokenRate = 1:3 = 0.3333
gasInEth = 1
gasInTokens should be 3

**_with .times, wrong!_**
gasInTokens
= gasInEth * ethToBuyTokenRate
= 1 * 0.33333
= 0.333333 // wrong

**_with .dividedBy, correct!_**
gasInTokens
= gasInEth / ethToBuyTokenRate
= 1 / 0.33333
= 3 // expected

xianny: I double checked in asset-swapper and I think the calculation here is correct, but the naming in asset-swapper is wrong, and this value actually represents XTokenToEthRate.

This is where the rate is sampled:
https://github.com/0xProject/protocol/blob/development/packages/asset-swapper/src/utils/market_operation_utils/index.ts#L166-L169

The comment here suggests it's "ETH -> maker token" ratio, e.g. 1 ETH to 4 tokens = 1:4 = 0.25

But the function that it's calling actually returns `(median maker token quote amount) / eth amount (1)`, i.e. 4 instead of 0.25:

https://github.com/0xProject/protocol/blob/development/packages/asset-swapper/src/utils/market_operation_utils/sampler_operations.ts#L936

kimpers: Awesome find! Thanks for digging into it 🙏  The naming is definitely incorrect. I have now updated it to be `sellTokenToEthRate` and `buyTokenToEthRate`. Other than the naming I think the calculations here are correct but please let me know if you see any issue 🙏

xianny:
merged in https://github.com/0xProject/0x-api/pull/439 instead

PR #434, opened 2020-11-30
feat: created a runner for rfq maker balance caching

alexkroeger:
# Description

Created a new runner that caches RFQ market maker balances.

Notable design choices:
- cached maker-token pairs for 10 minutes to reduce DB reads
- using a deployed contract that iterates over address-token pairs to make batch balance calls reads more efficient and accessible from plain JSON rpc (as opposed to using a bespoke balance cache from a third party)
- limit to 1000 balance reads per call to avoid out of gas errors

# Testing Instructions

# Checklist

-   [x] Test changes on the staging environment with [Matcha API Staging](https://api-staging.matcha.xyz).

    -   [ ] SRA/Limit orders
    -   [ ] Swap endpoints
    -   [ ] Depth charts


src/constants.ts
```diff
@@ -86,3 +86,59 @@ export const MARKET_DEPTH_END_PRICE_SLIPPAGE_PERC = 20;
...
+
+// ABI for balance checker
+export const BALANCE_CHECKER_ABI = [
+    {
+        constant: true,
+        inputs: [
+            { name: 'user', type: 'address' },
+            { name: 'token', type: 'address' },
+        ],
+        name: 'tokenBalance',
```
PirosB3, line 98: do we need `tokenBalance` or is it just an internal function?

src/constants.ts
```diff
@@ -86,3 +86,59 @@ export const MARKET_DEPTH_END_PRICE_SLIPPAGE_PERC = 20;
 
 // Logging
 export const NUMBER_SOURCES_PER_LOG_LINE = 12;
+
+// ABI for balance checker
+export const BALANCE_CHECKER_ABI = [
```
PirosB3, line 91: Given that this contract now exists in `Protocol` and has 0x bindings, should we use the 0x tooling instead of `ethers`?

src/runners/rfq_maker_balance_cache_runner.ts
```diff
@@ -0,0 +1,169 @@
...
+    if (!MAKER_TOKEN_CACHE) {
+        logUtils.log('updating cache');
+        MAKER_TOKEN_CACHE = createResultCache<any[]>(
+            () =>
+                connection
+                    .getRepository(MakerBalanceChainCacheEntity)
+                    .createQueryBuilder('maker_balance_chain_cache')
+                    .select(['maker_balance_chain_cache.tokenAddress', 'maker_balance_chain_cache.makerAddress'])
+                    .getMany(),
+            TEN_MINUTES_MS,
```
PirosB3, line 107: This may be a problem, unfortunately.

In the cache validator component, we assume a cache entry is invalid if more than 5 minutes have passed since the entry was first seen and no balances exist. If we cache results for 10 minutes in your code, this may cause new address entries to be seen as stale. In [this code](https://github.com/0xProject/0x-api/pull/432/files#diff-fa95b01194bc9cd84720be5a299890448459cfb7c4b7faa452681c9f11f65907R151), `_cacheExpiryThresholdMs` will default to 5 seconds. The code is changeable of course, so let's just sync up and discuss a solution that will work with both of our codes.

However, I may suggest thinking if a cache is really needed? this query is fairly simple, the results should have very low cardinality (~100 rows max) and, unlike the API `swap` request when we verify maker balances using this table, there is no "added pressure" on the DB or on request latency if this PG query takes more time than expected. Would it probably be better to simply add a gauge (using Prometheus) and simply benchmark latency? if latency becomes a problem, we can fix the issue?

PirosB3 (changes requested):
Started this review, have not finished yet

src/constants.ts
alexkroeger: It's callable. I'm just using the entire contract ABI here

src/constants.ts
```diff
@@ -86,3 +86,59 @@ export const MARKET_DEPTH_END_PRICE_SLIPPAGE_PERC = 20;
...
+        ],
+        name: 'balances',
+        outputs: [{ name: '', type: 'uint256[]' }],
+        payable: false,
+        stateMutability: 'view',
+        type: 'function',
+    },
+    { payable: true, stateMutability: 'payable', type: 'fallback' },
+];
+export const BALANCE_CHECKER_ADDRESS = '0x8d975E1a936BC9AaBB5d6cea9adCc5C3838070a7';
```
PirosB3, line 144: Won't work outside of `mainnet`. This will fail on our Kovan deployment

src/runners/rfq_maker_balance_cache_runner.ts
```diff
@@ -0,0 +1,169 @@
...
+    balanceCheckerContractInterface: ethers.Contract,
+    balancesCallInput: BalancesCallInput,
+): Promise<string[]> {
+    // due to gas contraints limit the call to 1K balance checks
+    const addressesChunkedArray = _.chunk(balancesCallInput.addresses, MAX_BALANCE_CHECKS_PER_CALL);
+    const tokensChunkedArray = _.chunk(balancesCallInput.tokens, MAX_BALANCE_CHECKS_PER_CALL);
+
+    const balances = await Promise.all(
+        addressesChunkedArray.map(async (addresses, i) => {
+            return balanceCheckerContractInterface.functions.balances(addresses, tokensChunkedArray[i]);
```
PirosB3, line 140: Not sure it matters here, but I wanted to point it out: there may be a race condition between the moment you read the block number to the moment you read the balances. Example: you read block number `x` but you then record balances for block number `x+1`

src/runners/rfq_maker_balance_cache_runner.ts
```diff
@@ -0,0 +1,169 @@
...
+        }),
+    );
+
+    const balancesFlattened = Array.prototype.concat.apply([], balances);
+
+    return balancesFlattened.map((bal: any) => bal.toString());
+}
+
+async function _updateErc20BalancesAsync(
```
PirosB3, line 149: outside of a class, you are fine not having to add an underscore prefix to a function call

src/runners/rfq_maker_balance_cache_runner.ts
```diff
@@ -0,0 +1,169 @@
...
+        } else {
+            await delay(DELAY_WHEN_NEW_BLOCK_NOT_FOUND);
+        }
+    }
+}
+
+// NOTE: this only returns a partial entity class, just token address and maker address
+// Cache the query results to reduce reads from the DB
+let MAKER_TOKEN_CACHE: ResultCache<MakerBalanceChainCacheEntity[]>;
+const _getMakerTokensAsync = async (connection: Connection) => {
```
PirosB3, line 97: Just to make it clearer, would you mind being explicit about the return type?

src/runners/rfq_maker_balance_cache_runner.ts
```diff
@@ -0,0 +1,169 @@
...
+    }
+    return (await MAKER_TOKEN_CACHE.getResultAsync()).result;
+};
+
+function _splitValues(makerTokens: MakerBalanceChainCacheEntity[]): BalancesCallInput {
+    const functionInputs: BalancesCallInput = { addresses: [], tokens: [] };
+
+    return makerTokens.reduce(({ addresses, tokens }, makerToken) => {
+        return {
+            addresses: addresses.concat(makerToken.makerAddress as string),
```
PirosB3, line 118: Instead of casting to string, you may be able to do `makerAddress!`

src/runners/rfq_maker_balance_cache_runner.ts
```diff
@@ -0,0 +1,169 @@
...
+            TEN_MINUTES_MS,
+        );
+    }
+    return (await MAKER_TOKEN_CACHE.getResultAsync()).result;
+};
+
+function _splitValues(makerTokens: MakerBalanceChainCacheEntity[]): BalancesCallInput {
+    const functionInputs: BalancesCallInput = { addresses: [], tokens: [] };
+
+    return makerTokens.reduce(({ addresses, tokens }, makerToken) => {
```
PirosB3, line 116: Just wanted to provide a suggested alternative implementation, however, your implementation is correct so don't feel you need to change anything unless you prefer :)

```
const addresses = [];
const tokens = [];
for (const makerToken of makerTokens) {
    addresses.push(makerToken.makerAddress);
    tokens.push(makerToken. tokenAddress);
}
return {tokens, addresses};
```

On another note, can we change the name `makerTokens` to be called something more like `balanceEntries`.

src/runners/rfq_maker_balance_cache_runner.ts
```diff
@@ -0,0 +1,169 @@
...
+async function _getErc20BalancesAsync(
+    balanceCheckerContractInterface: ethers.Contract,
+    balancesCallInput: BalancesCallInput,
+): Promise<string[]> {
+    // due to gas contraints limit the call to 1K balance checks
+    const addressesChunkedArray = _.chunk(balancesCallInput.addresses, MAX_BALANCE_CHECKS_PER_CALL);
+    const tokensChunkedArray = _.chunk(balancesCallInput.tokens, MAX_BALANCE_CHECKS_PER_CALL);
+
+    const balances = await Promise.all(
+        addressesChunkedArray.map(async (addresses, i) => {
```
PirosB3, line 139: why is this `async`?

src/runners/rfq_maker_balance_cache_runner.ts
```diff
@@ -0,0 +1,169 @@
...
+): Promise<ethers.Contract> {
+    return new ethers.Contract(BALANCE_CHECKER_ADDRESS, BALANCE_CHECKER_ABI, provider);
+}
+
+async function _getErc20BalancesAsync(
+    balanceCheckerContractInterface: ethers.Contract,
+    balancesCallInput: BalancesCallInput,
+): Promise<string[]> {
+    // due to gas contraints limit the call to 1K balance checks
+    const addressesChunkedArray = _.chunk(balancesCallInput.addresses, MAX_BALANCE_CHECKS_PER_CALL);
```
PirosB3, line 135: I appreciate that you are being mindful about the payload size. I agree this is something we need to be mindful about, in fact the current implementation could only really work if we have hundreds of addresses and not thousands. I would really encourage us to add a log warning somewhere if the number of addresses end up growing more than that number, since we may want to think about another way of doing this.

src/runners/rfq_maker_balance_cache_runner.ts
PirosB3, line 139: have you considered using `_.zip` here instead of iterating over 1 array and indexing in the other? just a thought

src/runners/rfq_maker_balance_cache_runner.ts
```diff
@@ -0,0 +1,169 @@
...
+    return balancesFlattened.map((bal: any) => bal.toString());
+}
+
+async function _updateErc20BalancesAsync(
+    balancesCallInput: BalancesCallInput,
+    balances: string[],
+    connection: Connection,
+    updateTime: Date,
+): Promise<void> {
+    const toSave = await Promise.all(
```
PirosB3, line 155: why are you using a `Promise` here?

src/runners/rfq_maker_balance_cache_runner.ts
PirosB3:
```typescript
_.zip(addressesChunkedArray, tokensChunkedArray).map(([addressesChunk, tokensChunk]) => {
    return balanceCheckerContractInterface.functions.balances(addressesChunk, tokensChunk);
});
```

src/runners/rfq_maker_balance_cache_runner.ts
```diff
@@ -0,0 +1,169 @@
...
+            dbEntity.makerAddress = addr;
+            dbEntity.tokenAddress = balancesCallInput.tokens[i];
+            dbEntity.balance = new BigNumber(balances[i]);
+            dbEntity.timeOfSample = updateTime;
+
+            return dbEntity;
+        }),
+    );
+
+    await connection.getRepository(MakerBalanceChainCacheEntity).save(toSave);
```
PirosB3, line 168: What does the query look like for this statement? Have you tried running `TypeORM` in debug mode and testing this out>

src/constants.ts
alexkroeger: 👍

src/constants.ts
alexkroeger: will work once we replace with asset-swapper tooling

src/runners/rfq_maker_balance_cache_runner.ts
alexkroeger: returns the entity, add type

src/runners/rfq_maker_balance_cache_runner.ts
alexkroeger: reduce to 30s, keep tabs on read latency

PirosB3:
```typescript
const MAKER_CACHE_RESULT_COUNT = new Counter({
    name: 'maker_cache_result_count',
    help: 'Add me here',
    labelNames: [],
});
```
```

File: src/runners/rfq_maker_balance_cache_runner.ts (hunk 5 above)
Comment (alexkroeger) [reply src/runners/rfq_maker_balance_cache_runner.ts:140]: talked about it, shouldn't be an issue

**Comment by PirosB3** [comment]
If 2 workers were to run this code at the same time, would that be a problem?

File: src/runners/rfq_maker_balance_cache_runner.ts (hunk 13 above)
Comment (alexkroeger) [reply src/runners/rfq_maker_balance_cache_runner.ts:168]: `advanced-console`

File: src/runners/rfq_maker_balance_cache_runner.ts (hunk 10 above)
Comment (alexkroeger) [reply src/runners/rfq_maker_balance_cache_runner.ts:139]: This contains an async function

File: src/runners/rfq_maker_balance_cache_runner.ts (hunk 12 above)
Comment (alexkroeger) [reply src/runners/rfq_maker_balance_cache_runner.ts:155]: unnecessary, removing

File: src/runners/rfq_maker_balance_cache_runner.ts (hunk 11 above)
Comment (alexkroeger) [reply src/runners/rfq_maker_balance_cache_runner.ts:135]: Discussed, keeping this in for protection against high gas usage

File: src/runners/rfq_maker_balance_cache_runner.ts (hunk 9 above)
Comment (alexkroeger) [reply src/runners/rfq_maker_balance_cache_runner.ts:116]: discussed, leaving as is

**Comment by alexkroeger** [comment]
deploy staging

File: src/runners/rfq_maker_balance_cache_runner.ts [hunk 14]
```diff
@@ -0,0 +1,222 @@
...
+                    .getRepository(MakerBalanceChainCacheEntity)
+                    .createQueryBuilder('maker_balance_chain_cache')
+                    .select(['maker_balance_chain_cache.tokenAddress', 'maker_balance_chain_cache.makerAddress'])
+                    .getMany(),
+            CACHE_MAKER_TOKENS_FOR_MS,
+        );
+    }
+    const results = (await MAKER_TOKEN_CACHE.getResultAsync()).result;
+
+    MAKER_BALANCE_CACHE_RESULT_COUNT.set(results.length);
```
PirosB3, line 143: Just to keep things consistent, would you mind adding the same label here too? thanks

src/runners/rfq_maker_balance_cache_runner.ts
```diff
@@ -0,0 +1,222 @@
...
+    provider: SupportedProvider,
+): BalanceCheckerContract {
+    return new BalanceCheckerContract(contractAddress, provider, { gas: BALANCE_CHECKER_GAS_LIMIT });
+}
+
+async function getErc20BalancesAsync(
+    balanceCheckerContractInterface: BalanceCheckerContract,
+    balancesCallInput: BalancesCallInput,
+    // HACK: allow for testing on ganache without override
+    codeOverride: boolean,
```
PirosB3, line 174: nice!

src/runners/rfq_maker_balance_cache_runner.ts
```diff
@@ -0,0 +1,222 @@
...
+): Promise<string[]> {
+    // due to gas contraints limit the call to 1K balance checks
+    const addressesChunkedArray = _.chunk(balancesCallInput.addresses, MAX_BALANCE_CHECKS_PER_CALL);
+    const tokensChunkedArray = _.chunk(balancesCallInput.tokens, MAX_BALANCE_CHECKS_PER_CALL);
+
+    const balanceCheckerByteCode = _.get(artifacts.BalanceChecker, 'compilerOutput.evm.deployedBytecode.object');
+
+    const balances = await Promise.all(
+        _.zip(addressesChunkedArray, tokensChunkedArray).map(async ([addressesChunk, tokensChunk]) => {
+            return !codeOverride
```
PirosB3, line 184:
```typescript
const txOpts = codeOverride? {
    overrides: {
        [RANDOM_ADDRESS]: {
            code: balanceCheckerByteCode,
        },
    },     
}: {}
return balanceCheckerContractInterface.balances(addressesChunk!, tokensChunk!).callAsync(txOpts, BlockParamLiteral.Latest)
```

PirosB3: ^ alternative implementation

src/runners/rfq_maker_balance_cache_runner.ts
```diff
@@ -0,0 +1,222 @@
...
+                                  code: balanceCheckerByteCode,
+                              },
+                          },
+                      },
+                      BlockParamLiteral.Latest,
+                  );
+        }),
+    );
+
+    const balancesFlattened = Array.prototype.concat.apply([], balances);
```
PirosB3, line 199: Alternative implementation using [flatten](https://lodash.com/docs/4.17.15#flatten)

```typescript
const balancesFlattened = _.flatten(balances)
```

test/rfq_maker_balance_cache_test.ts
```diff
@@ -0,0 +1,120 @@
...
+        provider = web3Factory.getRpcProvider(ganacheConfigs);
+
+        web3Wrapper = new Web3Wrapper(provider);
+
+        const accounts = await web3Wrapper.getAvailableAddressesAsync();
+        const deployer = accounts[0];
+        makerAddress1 = accounts[1];
+        makerAddress2 = accounts[2];
+
+        zrx = await DummyERC20TokenContract.deployFrom0xArtifactAsync(
```
PirosB3, line 49: Wow! congrats great stuff here

test/rfq_maker_balance_cache_test.ts
```diff
@@ -0,0 +1,120 @@
...
+            provider,
+            { from: deployer, gas: 10000000 },
+            {},
+            '0x Protocol Token',
+            'ZRX',
+            new BigNumber(18),
+            new BigNumber(1000000),
+        );
+
+        await zrx.mint(new BigNumber(100)).awaitTransactionSuccessAsync({ from: makerAddress1 });
```
PirosB3, line 60: I know this may not change anything, but we may want to simply keep the decimals into consideration here. You are minting 1e-16 tokens. Did you want to mint `100000000000000000000` tokens?

```typescript
Web3Wrapper.toBaseUnitAmount(100, 18)
```

PirosB3: same comment for below

PirosB3 (approved):
A few comments here and there, again I don't think that you need to address these comments and we should be good for liftoff!

test/rfq_maker_balance_cache_test.ts
alexkroeger: the balance pipeline never deals with decimals, so I'm not too worried it. I think prioritizing simplicity here is better

test/rfq_maker_balance_cache_test.ts
alexkroeger: thanks!

src/runners/rfq_maker_balance_cache_runner.ts
alexkroeger: I'll keep it as is for now

src/runners/rfq_maker_balance_cache_runner.ts
alexkroeger: 👍

PR #469, opened 2021-01-22
[DO NOT MERGE] fix: modified all the staking queries to join native_fills instead

sui414:
- this commit is to update staking queries to reflect v4 protocol fees collected.
- before merge we must add the v4 scrapers & backfill all records into native_fills table.

- tested all the queries locally on hashalytics

### DIFF
--- a/src/queries/staking_queries.ts
+++ b/src/queries/staking_queries.ts
@@ -322,20 +322,16 @@ export const poolTotalProtocolFeesGeneratedQuery = `
                   case when fwe.protocol_version='v4' then fwe.pool
                       else esps.pool_id end
                    as pool_id
-                  , SUM(protocol_fee_paid) / 1e18 AS total_protocol_fees
-                  , COUNT(*) AS number_of_fills
-              FROM fills_with_epochs fwe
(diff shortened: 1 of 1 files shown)


src/queries/staking_queries.ts
```diff
@@ -480,17 +473,16 @@ export const nextEpochPoolsStatsQuery = `
...
-            FROM events.fill_events fe
-            LEFT JOIN staking.pool_info pi ON fe.maker_address = ANY(pi.maker_addresses)
+            FROM events.native_fills fe
             JOIN staking.current_epoch ce
                 ON fe.block_number > ce.starting_block_number
                 OR (fe.block_number = ce.starting_block_number AND fe.transaction_index >= ce.starting_transaction_index)
             WHERE
                 -- fees not accruing to a pool do not count
-                pool_id IS NOT NULL
+                pool IS NOT NULL
```
alexkroeger, line 485: This only works for v4 fills, we need the maker address matching for v3 fills

sui414:
deploy staging

xianny:
fixed in https://github.com/0xProject/0x-event-pipeline/pull/63

PR #499, opened 2021-03-03
fix: lowercase addresses in order update event data received from Mesh

kimpers:
# Description
There is an inconsistency in the address casing of v4 order data received from Mesh, it seems like query responses return strictly lower case addresses, whereas the subscription order event updates can return checksummed addresses. This lead to some issues with checksummed addresses temporarily getting into the DB when inserted from an `ADDED` event, but then lowercased when API was restarted and it does a full sync query to Mesh.

# Testing Instructions

# Checklist

    -   [ ] SRA/Limit orders
    -   [ ] Swap endpoints
    -   [ ] Depth charts


xianny (approved):
Nice catch!

PR #507, opened 2021-03-09
fix: expect sra pool filter to be a hex number just like in the order

kimpers:
# Description

If filtering SRA order by pool, it needs to be the padded hex number that is in the order (`0x0000000000000000000000000000000000000000000000000000000000000001`) so we should validate that it's a hex number.

# Testing Instructions

# Checklist

    -   [ ] SRA/Limit orders
    -   [ ] Swap endpoints
    -   [ ] Depth charts


xianny (approved):
👍 🙏

PR #525, opened 2021-03-29
wip: skipValidation

xianny:
# Description

# Testing Instructions

# Checklist

    -   [ ] SRA/Limit orders
    -   [ ] Swap endpoints
    -   [ ] Depth charts

### DIFF
--- a/src/handlers/sra_handlers.ts
+++ b/src/handlers/sra_handlers.ts
@@ -77,6 +77,7 @@ export class SRAHandlers {
         res.status(HttpStatus.OK).send(orderbookResponse);
     }
     public async postOrderAsync(req: express.Request, res: express.Response): Promise<void> {
+        const shouldSkipValidation = req.params.skipValidation === 'true';
         schemaUtils.validateSchema(req.body, schemas.sraPostOrderPayloadSchema);
         const signedOrder = unmarshallOrder(req.body);
         if (WHITELISTED_TOKENS !== '*') {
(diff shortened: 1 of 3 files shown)


src/handlers/sra_handlers.ts
```diff
@@ -77,19 +77,26 @@ export class SRAHandlers {
...
+        const shouldSkipValidation = req.query.skipValidation === 'true';
         schemaUtils.validateSchema(req.body, schemas.sraPostOrderPayloadSchema);
         const signedOrder = unmarshallOrder(req.body);
         if (WHITELISTED_TOKENS !== '*') {
             const allowedTokens: string[] = WHITELISTED_TOKENS;
             validateAssetTokenOrThrow(allowedTokens, signedOrder.makerToken, 'makerToken');
             validateAssetTokenOrThrow(allowedTokens, signedOrder.takerToken, 'takerToken');
         }
+        if (shouldSkipValidation) {
+            res.status(HttpStatus.OK).send();
```
PirosB3, line 89: Would this close the connection for the client? don't we need something like `close()`? or does `status()` close the connection too?

src/handlers/sra_handlers.ts
```diff
@@ -77,19 +77,26 @@ export class SRAHandlers {
...
     public async postOrderAsync(req: express.Request, res: express.Response): Promise<void> {
+        const shouldSkipValidation = req.query.skipValidation === 'true';
         schemaUtils.validateSchema(req.body, schemas.sraPostOrderPayloadSchema);
         const signedOrder = unmarshallOrder(req.body);
         if (WHITELISTED_TOKENS !== '*') {
             const allowedTokens: string[] = WHITELISTED_TOKENS;
             validateAssetTokenOrThrow(allowedTokens, signedOrder.makerToken, 'makerToken');
             validateAssetTokenOrThrow(allowedTokens, signedOrder.takerToken, 'takerToken');
         }
+        if (shouldSkipValidation) {
```
PirosB3, line 88: Just curious, is there any commenting that we could add to this code to explain why we are doing this? optionally, we could also change the parameter from `skipValidation` (which is used in `swap` to avoid the final `eth_call` that checks that the swap is successful) to something like `enqueueInsertion=true` - just ideas :)

PirosB3 (changes requested):

src/handlers/sra_handlers.ts
xianny: Nah, `send()` calls `end()` as well. We use `send()` instead of `end()` as a pattern because it generates an ETag header for cache validation.

src/handlers/sra_handlers.ts
xianny: Skipping Mesh validation also skips the on-chain maker balance validation, so it's kinda similar to the concept of Swap API `skipValidation` - no guarantee that your order "works". Just wanna emphasize that this implementation means the user may get a 200 OK even if Mesh rejects the order.

src/handlers/sra_handlers.ts
PirosB3: Thanks for clarifying

src/handlers/sra_handlers.ts
PirosB3: Yes, however the `skipValidation` parameter already exists right and have a specific purpose? don't you feel like, even if the current parameter could be interpreted the way you are suggesting, it just could be safer to change the name to something more explicit?

PR #535, opened 2021-04-13
feat: add RfqmService, implement fetching indicative RFQM quotes [MKR-29]  

phil-ociraptor:
# Description

Introduce the `RfqmService`, which will become the coordination layer for HTTP based RFQM flows (aka, everything except the queue consumer/worker)

# Testing Instructions

Unit tests provided. Introduced a new library called ts-mockito, it's a noticeable improvement over Sinon. In future commits, will introduce more integration tests, which is the standard form of testing in 0x-api.

# Checklist

    -   [ ] SRA/Limit orders
    -   [ ] Swap endpoints
    -   [ ] Depth charts


package.json
```diff
@@ -123,14 +123,15 @@
         "sinon": "^9.2.1",
         "supertest": "^4.0.2",
         "ts-mock-imports": "^1.3.0",
+        "ts-mockito": "^2.6.1",
```
alexkroeger, line 126: what is this for?

src/services/rfqm_service.ts
```diff
@@ -0,0 +1,126 @@
...
+        const indicativeQuotes = await this._quoteRequestor.requestRfqmIndicativeQuotesAsync(
+            makerToken,
+            takerToken,
+            assetFillAmount,
+            marketOperation,
+            undefined,
+            opts,
+        );
+
+        // Filter only quotes that are 100% filled, and sort by best price
```
alexkroeger, line 52:
```suggestion
        // Filter out quotes that cannot fill 100% of the requested amount, and sort by best price
```

src/services/rfqm_service.ts
```diff
@@ -0,0 +1,126 @@
...
+        const firmQuotes = await this._quoteRequestor.requestRfqmFirmQuotesAsync(
+            makerToken,
+            takerToken,
+            assetFillAmount,
+            marketOperation,
+            undefined,
+            opts,
+        );
+
+        // Filter only quotes that are 100% filled, and sort by best price
```
alexkroeger, line 112: same comment as above

package.json
phil-ociraptor: Had convo offline, but the gist is: 

- Adopt dependency injection (never creating dependencies in our constructors and always getting deps passed in)
- Write unit tests that pass in mocked deps
- Use `ts-mockito` to make making those mocks painless

src/services/rfqm_service.ts
```diff
@@ -0,0 +1,138 @@
...
+ * RfqmService is the coordination layer for HTTP based RFQM flows
+ */
+export class RfqmService {
+    constructor(private readonly _quoteRequestor: QuoteRequestor) {}
+
+    /**
+     * Fetch the best indicative quote available
+     */
+    public async fetchIndicativeQuoteAsync(params: FetchIndicativeQuoteParams): Promise<FetchIndicativeQuoteResponse> {
+        const { sellAmount, buyAmount, sellToken, buyToken, apiKey } = params;
```
xianny, line 22: Can we use a single `amount` param instead of `buyAmount` and `sellAmount`?

src/types.ts
```diff
@@ -354,6 +354,36 @@ export interface CalculateMetaTransactionQuoteParams extends SwapQuoteParamsBase
...
+export interface FetchFirmQuoteParams {
+    apiKey: string;
+    buyAmount?: BigNumber;
+    buyToken: string;
+    sellAmount?: BigNumber;
+    sellToken: string;
+    takerAddress?: string;
+}
+export interface FetchFirmQuoteResponse {}
+
```
xianny, line 384: Empty interface? Does it extend the indicative response?

src/types.ts
```diff
@@ -354,6 +354,36 @@ export interface CalculateMetaTransactionQuoteParams extends SwapQuoteParamsBase
 
 /** End /swap types */
 
+/** Start /rfqm types */
```
xianny, line 357: Let's put these in the RfqmService file, or perhaps in the handler if the same params/response are parsed/returned from there.

src/services/rfqm_service.ts
```diff
@@ -0,0 +1,138 @@
...
+            price: isSelling
+                ? bestQuote.makerAmount.div(bestQuote.takerAmount)
+                : bestQuote.takerAmount.div(bestQuote.makerAmount),
+        };
+    }
+
+    /**
+     * Fetch the best firm quote available, as a signable metatransaction
+     */
+    public async fetchFirmQuoteAsync(params: FetchFirmQuoteParams): Promise<FetchFirmQuoteResponse> {
```
xianny, line 89: I noticed the function implementation is pretty much identical. Do you anticipate them diverging, or will most of the business logic be handled by QuoteRequestor? If the latter, would suggest DRYing up this code. 

e.g. 
```
private async _fetchQuoteAsync(params: RfqmQuoteParams, isIndicative: boolean): { 
    // same implementation, just pass isIndicative directly through to opts
}
```

And `fetchFirmQuote` can do additional work to create a metatx.

xianny (commented):
Left some comments on implementation, looks good so far.

I'm a fan of dependency injection, could you add some notes about advantages of ts-mockito vs sinon though? 

Fine with introducing ts-mockito here, let's see how it goes :)

src/services/rfqm_service.ts
phil-ociraptor: Good idea - I mostly copied over the params from the swap endpoint, but better to start simple and add on this complexity if we really need it

src/types.ts
phil-ociraptor: Just a placeholder at the moment - but yes, a good chance it will ultimately extend the indicative quote response

src/services/rfqm_service.ts
phil-ociraptor: Good idea - will give this a shot

phil-ociraptor: Gave it a shot - but after wrestling with some of our types (so many union types), will punt on this for now.

I'm going to remove fetching firm quotes at the moment, as it is far from being complete

package.json
PirosB3: Why not use the already existing TypeMoq?

src/services/rfqm_service.ts
```diff
@@ -0,0 +1,95 @@
...
+    constructor(private readonly _quoteRequestor: QuoteRequestor) {}
+
+    /**
+     * Fetch the best indicative quote available.
+     */
+    public async fetchIndicativeQuoteAsync(params: FetchIndicativeQuoteParams): Promise<FetchIndicativeQuoteResponse> {
+        const { amount, sellToken, buyToken, apiKey } = params;
+
+        // Map params to the terminology of Quote Requestor
+        const marketOperation = MarketOperation.Sell;
```
PirosB3, line 36: Are we not also supporting buys?

src/services/rfqm_service.ts
```diff
@@ -0,0 +1,95 @@
+import { MarketOperation, QuoteRequestor } from '@0x/asset-swapper';
+import { BigNumber } from '@0x/utils';
+
+import { RFQT_REQUEST_MAX_RESPONSE_MS } from '../config';
+import { NULL_ADDRESS } from '../constants';
+
+export interface FetchIndicativeQuoteParams {
```
PirosB3, line 7: Could we move these to `types.ts`?

src/services/rfqm_service.ts
```diff
@@ -0,0 +1,95 @@
...
+        // Map params to the terminology of Quote Requestor
+        const marketOperation = MarketOperation.Sell;
+        const assetFillAmount = amount;
+        const makerToken = buyToken;
+        const takerToken = sellToken;
+
+        // Fetch quotes
+        const opts = {
+            takerAddress: NULL_ADDRESS,
+            txOrigin: NULL_ADDRESS, // TODO - set to worker registry
```
PirosB3, line 44: Suggest to update `config.ts` to read a new address from the environment files

src/services/rfqm_service.ts
```diff
@@ -0,0 +1,95 @@
...
+            apiKey,
+            intentOnFilling: false, // TODO - safe to hardcode?
+            isIndicative: true,
+            makerEndpointMaxResponseTimeMs: RFQT_REQUEST_MAX_RESPONSE_MS,
+            nativeExclusivelyRFQ: true,
+            altRfqAssetOfferings: {}, // TODO - set this to the alt rfq asset offerings
+            isLastLook: true,
+        };
+        const indicativeQuotes = await this._quoteRequestor.requestRfqmIndicativeQuotesAsync(
```
PirosB3, line 53: Where are we reading the new RFQM configuration map? we want to have a dictionary of RFQT market makers and a dictionary of RFQM market makers right? where are the latter?

src/services/rfqm_service.ts
```diff
@@ -0,0 +1,95 @@
...
+            undefined,
+            opts,
+        );
+
+        // Filter out quotes that:
+        // - are for the wrong pair
+        // - cannot fill 100 % of the requested amount
+        //
+        // And sort by best price
+        const sortedQuotes = indicativeQuotes
```
PirosB3, line 67: This is awersome! is there any chance we could move this to a separte function, so that we can properly unit test it?

src/services/rfqm_service.ts
```diff
@@ -0,0 +1,95 @@
...
+            .filter((q) => q.takerAmount.gte(assetFillAmount))
+            .sort((a, b) => {
+                const aPrice = a.makerAmount.div(a.takerAmount);
+                const bPrice = b.makerAmount.div(b.takerAmount);
+                return bPrice.minus(aPrice).toNumber();
+            });
+
+        // No quotes found
+        if (sortedQuotes.length === 0) {
+            return Promise.reject(new Error('No valid quotes'));
```
PirosB3, line 78: I wouldn't say this is an error right? could we just return empty quotes?

PirosB3: Where would this error bubble up to?

PirosB3 (changes requested):
Not finished! just switching code review app (VSCode)

src/services/rfqm_service.ts
phil-ociraptor: Had supported both, but after addressing a previous comment, I removed it to only support sells: https://github.com/0xProject/0x-api/pull/535#discussion_r615048092

Sounds like this might be a discussion to have with the Taker team. I get the sense there is a push to frame things in terms of sells - but I lack context to chime in.

src/services/rfqm_service.ts
phil-ociraptor: So, I had these in `types.ts` but after addressing a previous comment, moved them here: https://github.com/0xProject/0x-api/pull/535#discussion_r615051010

Happy to do either one, but would be ideal to settle on a single convention

src/services/rfqm_service.ts
phil-ociraptor: The RFQT Market Maker map and the RFQM Market Maker map are passed in to the quoteRequestor in the constructor.

QuoteRequestor is not constructed within the `RFQMService` as I am employing dependency injection. See #542 for where the construction occurs

src/services/rfqm_service.ts
phil-ociraptor: I've been thinking about how to pull out this logic, but want to wait, because this logic will be shared with firm quotes, but the types are not the same.

All said, the current unit test coverage does exercise this code! Perhaps not exhaustively, but I believe the major cases are there

Will revisit plucking out this code when I implement the firm quote endpoint

src/services/rfqm_service.ts
```diff
@@ -0,0 +1,95 @@
...
+
+        // Filter out quotes that:
+        // - are for the wrong pair
+        // - cannot fill 100 % of the requested amount
+        //
+        // And sort by best price
+        const sortedQuotes = indicativeQuotes
+            .filter((q) => q.takerToken === takerToken && q.makerToken === makerToken)
+            .filter((q) => q.takerAmount.gte(assetFillAmount))
+            .sort((a, b) => {
```
PirosB3, line 70: Can we also check reasonable expiration window? maybe it's safe to assume quote will live for at least 3 minutes?

src/services/rfqm_service.ts
phil-ociraptor: Makes sense, the empty quote is a good return value.

Right now, I believe the error is being caught by our error handler middleware `error_handling.ts` - which maps certain error types to specific HTTP error responses - but I haven't investigated that component super closely.

src/services/rfqm_service.ts
```diff
@@ -0,0 +1,95 @@
...
+        );
+
+        // Filter out quotes that:
+        // - are for the wrong pair
+        // - cannot fill 100 % of the requested amount
+        //
+        // And sort by best price
+        const sortedQuotes = indicativeQuotes
+            .filter((q) => q.takerToken === takerToken && q.makerToken === makerToken)
+            .filter((q) => q.takerAmount.gte(assetFillAmount))
```
PirosB3, line 69: Is this always taker amount? what if we are selling? then we should be checking `makerAmount` right?

phil-ociraptor:
GitHub is being weird and not letting me respond to the `TypeMoq` comment so I'll do it here. I believe TypeMoq isn't actually a dependency in 0x-api, although I do think I've seen it elsewhere.

Given the choice between TypeMoq and ts-mockito, went with ts-mockito because I'm familiar with mockito, and ts-mockito has more github stars and weekly npm downloads than typemoq.

That said, happy to change to typemoq if we've built up usage around it!

src/services/rfqm_service.ts
```diff
@@ -0,0 +1,95 @@
...
+
+        // Map params to the terminology of Quote Requestor
+        const marketOperation = MarketOperation.Sell;
+        const assetFillAmount = amount;
+        const makerToken = buyToken;
+        const takerToken = sellToken;
+
+        // Fetch quotes
+        const opts = {
+            takerAddress: NULL_ADDRESS,
```
PirosB3, line 43: Can we create some RFQM default opts here, and use it as a mixin?

```
const opts = {
    ...defaultRfqmOpts,
    apiKey,
    // And so forth..
}
```

src/services/rfqm_service.ts
PirosB3: filter by expiration

src/services/rfqm_service.ts
PirosB3: I see, ideally the RFQM service would raise particular custom exceptions (classes that inherit from `Error`), and the handler can catch them and return a fine error code and message. This way we can distinguish between 2xx, 4xx, 5xx responses

src/services/rfqm_service.ts
```diff
@@ -0,0 +1,95 @@
...
+        const bestQuote = sortedQuotes[0];
+
+        // Prepare response
+        // TODO: handle decimals properly in price
+        return {
+            buyAmount: bestQuote.makerAmount,
+            buyTokenAddress: bestQuote.makerToken,
+            sellAmount: bestQuote.takerAmount,
+            sellTokenAddress: bestQuote.takerToken,
+            price: bestQuote.makerAmount.div(bestQuote.takerAmount),
```
PirosB3, line 90: It would be really ideal if we could return a similar output to the `price` endpoint for RFQT. Something like

```
{
    "price": "0.000433262422",
    "gas": "160600",
    "buyTokenAddress": "0xa0b86991c6218b36c1d19d4a2e9eb0ce3606eb48",
    "buyAmount": "5000000000",
    "sellTokenAddress": "0xc02aaa39b223fe8d0a0e5c4f27ead9083c756cc2",
    "sellAmount": "2166312110000000000",
    "allowanceTarget": "0xdef1c0ded9bec7f1a1670819833240f027b25eff",
}
```

PirosB3 (changes requested):
Did my review here too! :)

phil-ociraptor:
Ok, Github is freaking out - it's including far too many changes in this diff that don't belong in this diff. Will close this and repoen this diff. 

Apologies for the inconvenience

PR #536, opened 2021-04-13
feat: add rfqm utils for generating a meta tx [MKR-15]

alexkroeger:
# Description

Add utils for creating a meta tx object for RFQM.

# Testing Instructions

Added unit tests for meta TX validator.

# Checklist

    -   [ ] SRA/Limit orders
    -   [ ] Swap endpoints
    -   [ ] Depth charts


src/utils/rfqm_blockchain_utils.ts
```diff
@@ -0,0 +1,40 @@
...
+
+// allow a wide range for gas price for flexibility
+const MIN_GAS_PRICE = new BigNumber(0);
+// 10K Gwei
+const MAX_GAS_PRICE = new BigNumber(1e13);
+
+export class RfqBlockchainUtils {
+    constructor(private readonly _exchangeProxy: IZeroExContract) {}
+
+    public generateMetatransaction(
```
phil-ociraptor, line 15: perhaps add a comment that this method is used in the RFQM flow when 0x submits the transaction on chain (as opposed to the Maker submitting on chain)

phil-ociraptor (commented):
Looking good!

src/utils/rfqm_blockchain_utils.ts
```diff
@@ -0,0 +1,61 @@
...
+    ): Promise<boolean> {
+        try {
+            await this._exchangeProxy.executeMetaTransaction(metaTx, metaTxSig).callAsync({ from: sender });
+            return true;
+        } catch (error) {
+            return false;
+        }
+    }
+
+    public generateMetaTxCallData(metaTx: MetaTransaction, metaTxSig: Signature): string {
```
phil-ociraptor, line 58: nit: for consistency of naming with other functions in this class, the name `generateMetaTransactionCallData` is preferred

phil-ociraptor (changes requested):
Overall looks great! Just some super minor nits and a couple suggestions

src/utils/rfqm_blockchain_utils.ts
```diff
@@ -0,0 +1,61 @@
...
+            verifyingContract: this._exchangeProxy.address,
+        });
+    }
+
+    public async isValidMetaTransactionAsync(
+        metaTx: MetaTransaction,
+        metaTxSig: Signature,
+        sender: string,
+    ): Promise<boolean> {
+        try {
```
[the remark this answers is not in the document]
alexkroeger: yea I figured expiration checks are pretty simple to do from the metaTx object itself

src/utils/rfqm_blockchain_utils.ts
```diff
@@ -0,0 +1,61 @@
...
+            await this._exchangeProxy.executeMetaTransaction(metaTx, metaTxSig).callAsync({ from: sender });
+            return true;
+        } catch (error) {
+            return false;
+        }
+    }
```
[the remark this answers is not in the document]
alexkroeger: I agree, I was thinking about how to better surface the error, I like the suggestion!

src/utils/rfqm_blockchain_utils.ts
```diff
@@ -0,0 +1,61 @@
+import { IZeroExContract } from '@0x/contracts-zero-ex';
+import { MetaTransaction, RfqOrder, Signature } from '@0x/protocol-utils';
+import { BigNumber } from '@0x/utils';
+
+import { NULL_ADDRESS } from '../constants';
+
+// allow a wide range for gas price for flexibility
+const MIN_GAS_PRICE = new BigNumber(0);
+// 10K Gwei
+const MAX_GAS_PRICE = new BigNumber(1e13);
```
PirosB3, line 10: Is this max gas price some kind of guardrail? Are we trying to avoid gas estimation from going crazy?

src/utils/rfqm_blockchain_utils.ts
```diff
@@ -0,0 +1,61 @@
...
+export class RfqBlockchainUtils {
+    constructor(private readonly _exchangeProxy: IZeroExContract) {}
+
+    // for use when 0x API operator submits an order on-chain on behalf of taker
+    public generateMetaTransaction(
+        rfqOrder: RfqOrder,
+        signature: Signature,
+        taker: string,
+        takerAmount: BigNumber,
+        chainId: number,
```
PirosB3, line 21: Can this be of type `ChainId`?

src/utils/rfqm_blockchain_utils.ts
```diff
@@ -0,0 +1,61 @@
...
+            .fillRfqOrder(rfqOrder, signature, takerAmount)
+            .getABIEncodedTransactionData();
+
+        return new MetaTransaction({
+            signer: taker,
+            sender: NULL_ADDRESS,
+            minGasPrice: MIN_GAS_PRICE,
+            maxGasPrice: MAX_GAS_PRICE,
+            expirationTimeSeconds: rfqOrder.expiry,
+            salt: new BigNumber(Date.now()),
```
PirosB3, line 34: Just checking here. `Date.now()` will return a number to the millisecond. Usually we set Unix timestamps in this field (these tend to be smaller). Do we see any downsides with using a bigger field with `Date.now()`? I don't, but just thought it could be valuable to open the discussion

src/utils/rfqm_blockchain_utils.ts
```diff
@@ -0,0 +1,61 @@
...
+            callData,
+            // may need this to be variable to handle ETH -> ERC20
+            value: new BigNumber(0),
```
PirosB3, line 37: There should be a `ZERO` constant in the codebase

src/utils/rfqm_blockchain_utils.ts
```diff
@@ -0,0 +1,61 @@
...
+            feeToken: rfqOrder.takerToken,
```
PirosB3, line 38: Shouldn't this be `NULL_ADDRESS`?

src/utils/rfqm_blockchain_utils.ts
```diff
@@ -0,0 +1,61 @@
...
+        });
+    }
+
+    public async validateMetaTransactionOrThrowAsync(
+        metaTx: MetaTransaction,
+        metaTxSig: Signature,
+        sender: string,
+    ): Promise<boolean> {
+        try {
+            await this._exchangeProxy.executeMetaTransaction(metaTx, metaTxSig).callAsync({ from: sender });
```
PirosB3, line 51: Could we parse the return value and ensure that the actual call data in the metatransaction was successful?

PirosB3: Is it worth hard-coding some other values during the `eth_call` (such as max gas used, gas price)? This will call the trade to rightfully revert even if the metatransaction is valid but something is funky with the worker.

PirosB3: (Example, worker is out of ETH, will cause trade to revert)

PirosB3 (approved):
Just a few comments, awesome overall!

test/rfq_blockchain_utils_test.ts
```diff
@@ -0,0 +1,162 @@
...
+    after(async () => {
+        await teardownDependenciesAsync(SUITE_NAME);
+    });
+
+    describe('validateMetaTransaction', () => {
+        it('returns true for a valid metatransaction', async () => {
+            const metaTx = rfqBlockchainUtils.generateMetaTransaction(rfqOrder, orderSig, taker, takerAmount, CHAIN_ID);
+            const metaTxSig = await metaTx.getSignatureWithProviderAsync(provider);
+
+            expect(await rfqBlockchainUtils.validateMetaTransactionOrThrowAsync(metaTx, metaTxSig, txOrigin)).to.eq(true);
```
PirosB3, line 137: It would be really interesting if you can take this test to the next level and actually asset that the swap was successful? Maybe by having `validateMetaTransactionOfThrowAsync` return the amount filled? Just an idea :)

phil-ociraptor (approved):
Looks good from my side! Just the small naming issue of `src/utils/rfqm_blockchain_utils.ts` being called `rfqm` instead of `rfq` remains.

src/utils/rfqm_blockchain_utils.ts
alexkroeger: having milliseconds as the salt for now should be fine, no downsides I can think of.

src/utils/rfqm_blockchain_utils.ts
alexkroeger: 👍

src/utils/rfqm_blockchain_utils.ts
alexkroeger: 👍

src/utils/rfqm_blockchain_utils.ts
alexkroeger: 👍

test/rfq_blockchain_utils_test.ts
alexkroeger: 👍

src/utils/rfqm_blockchain_utils.ts
alexkroeger: added some more flexibility

PR #539, opened 2021-04-14
fix: New API key structure

PirosB3:
This PR creates a new API key structure, where every API key can have multiple "features" enabled. The feature is built so that it will fallback on the existing key structure in order to support a phased migration of the configurations.

### New key structure
```json
{
    "foo": {"rfqm": false, "rfqt": true, "plp": true, "label": "Foo key" }, 
    "bar": {"rfqm": false, "rfqt": true, "plp": false, "label": "Bar key" }, 
    "baz": {"rfqm": false, "rfqt": false, "plp": true, "label": "Baz key" }, 
    "barf": {"rfqm": true, "rfqt": false, "plp": false, "label": "Barf key" }, 
}
```

### DIFF
--- a/src/config.ts
+++ b/src/config.ts
@@ -95,13 +95,13 @@ export const getApiKeyWhitelistWithFallback = (
         const keyMeta = deserialized[apiKey];
         switch (groupType) {
             case 'plp':
-                if (keyMeta.plp) result.push(apiKey);
+                if (keyMeta.plp) { result.push(apiKey); }
                 break;
             case 'rfqm':
-                if (keyMeta.rfqm) result.push(apiKey);
+                if (keyMeta.rfqm) { result.push(apiKey); }
                 break;
             case 'rfqt':
-                if (keyMeta.rfqt) result.push(apiKey);
+                if (keyMeta.rfqt) { result.push(apiKey); }
                 break;
             default:
                 throw new Error(`Unknown group type inputted: ${groupType}`);
--- a/src/config.ts
+++ b/src/config.ts
@@ -98,10 +98,10 @@ export const getApiKeyWhitelistWithFallback = (
                 if (keyMeta.plp) result.push(apiKey);
                 break;
             case 'rfqm':
-                if (keyMeta.rfqm ===sult.push(apiKey);
+                if (keyMeta.rfqm) result.push(apiKey);
                 break;
             case 'rfqt':
-                if (keyMeta.rfqt === tru.push(apiKey);
+                if (keyMeta.rfqt) result.push(apiKey);
                 break;
             default:
                 throw new Error(`Unknown group type inputted: ${groupType}`);
--- a/src/config.ts
+++ b/src/config.ts
@@ -30,9 +30,11 @@ import {
     QUOTE_ORDER_EXPIRATION_BUFFER_MS,
     TX_BASE_GAS,
 } from './constants';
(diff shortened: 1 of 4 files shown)


src/config.ts
```diff
@@ -220,9 +270,11 @@ export const ALT_RFQ_MM_PROFILE: string | undefined = _.isEmpty(process.env.ALT_
...
     : assertEnvVarType('ALT_RFQ_MM_PROFILE', process.env.ALT_RFQ_MM_PROFILE, EnvVarType.NonEmptyString);
 
-export const PLP_API_KEY_WHITELIST: string[] = _.isEmpty(process.env.PLP_API_KEY_WHITELIST_JSON)
-    ? []
-    : assertEnvVarType('PLP_API_KEY_WHITELIST_JSON', process.env.PLP_API_KEY_WHITELIST_JSON, EnvVarType.JsonStringList);
+export const PLP_API_KEY_WHITELIST: string[] = getApiKeyWhitelistWithFallback(
+    'PLP_API_KEY_WHITELIST_JSON',
+    'API_KEYS_ACL',
+    'plp'
+);
```
alexkroeger, line 278: might as well add RFQM list to config

src/config.ts
```diff
@@ -58,6 +60,56 @@ enum EnvVarType {
...
+export interface ApiKeyStructure {
+    [key: string]: {
+        label: string;
+        rfqt: boolean;
+        rfqm: boolean;
+        plp: boolean;
+    };
+}
+
+export const getApiKeyWhitelistWithFallback = (
```
alexkroeger, line 72: I personally think it's a bit overkill to support both old and new, we could just deploy with a new env var and just support new

alexkroeger: not blocking, just preference

PirosB3: So you would perform the rollout by:
1. Copying API keys to new format
2. Deploying new code
3. Removing old API keys?

My strategy is to:
1. Deploy the new code, make sure it works well
2. Update environment variables (move old keys to new keys)

alexkroeger: Yea my pref is to just add copy the API keys to the new format, delete the old env var, and deploy (all in one main-infra PR), test on staging first.

alexkroeger: If the new key is there it'll use the new key, so it either works or it doesn't, if it doesn't we roll back.

alexkroeger (approved):

src/config.ts
```diff
@@ -58,6 +60,56 @@ enum EnvVarType {
...
+
+    const result: string[] = [];
+    for (const apiKey of Object.keys(deserialized)) {
+        const keyMeta = deserialized[apiKey];
+        switch (groupType) {
+            case 'plp':
+                if (keyMeta.plp) result.push(apiKey);
+                break;
+            case 'rfqm':
+                if (keyMeta.rfqm ===sult.push(apiKey);
```
alexkroeger, line 101: should be result.push?

src/config.ts
```diff
@@ -58,6 +60,56 @@ enum EnvVarType {
...
+                break;
+            case 'rfqt':
+                if (keyMeta.rfqt === tru.push(apiKey);
```
alexkroeger, line 104: same comment here

PirosB3: Crappy linter

PR #542, opened 2021-04-15
feat: add RFQM handler and http runner [MKR-31]

phil-ociraptor:
# Description

This change brings the RFQM service to life, by adding endpoints that wrap the underlying logic. It also introduces integration tests

# Testing Instructions

# Checklist

    -   [ ] SRA/Limit orders
    -   [ ] Swap endpoints
    -   [ ] Depth charts


src/runners/http_rfqm_service_runner.ts
```diff
@@ -0,0 +1,73 @@
...
+    dependencies: AppDependencies,
+    config: HttpServiceConfig,
+    _app?: core.Express,
+): Promise<{ app: express.Application; server: Server }> {
+    const app = _app || express();
+    app.use(addressNormalizer);
+    app.use(cacheControl(DEFAULT_CACHE_AGE_SECONDS));
+    const server = createDefaultServer(config, app, logger, destroyCallback(dependencies));
+
+    app.get('/', rootHandler);
```
PirosB3, line 61: Just curious, do we even need this?

src/runners/http_rfqm_service_runner.ts
```diff
@@ -0,0 +1,73 @@
...
+ * Runs the Rfqm Service in isolation
+ */
+export async function runHttpRfqmServiceAsync(
+    dependencies: AppDependencies,
+    config: HttpServiceConfig,
+    _app?: core.Express,
+): Promise<{ app: express.Application; server: Server }> {
+    const app = _app || express();
+    app.use(addressNormalizer);
+    app.use(cacheControl(DEFAULT_CACHE_AGE_SECONDS));
```
PirosB3, line 58: What is this caching for? we are not caching RFQM quotes right?

src/handlers/rfqm_handlers.ts
```diff
@@ -0,0 +1,63 @@
...
+        res.status(HttpStatus.OK).send(indicativeQuote);
+    }
+}
+
+const parseFetchIndicativeQuoteParams = (req: express.Request): FetchIndicativeQuoteParams => {
+    // HACK - reusing the validation for Swap Quote as the interface here is a subset
+    schemaUtils.validateSchema(req.query, schemas.swapQuoteRequestSchema as any);
+    const apiKey = req.header('0x-api-key');
+    if (apiKey === undefined) {
+        throw new Error('Must access with an API key');
```
PirosB3, line 28: How do these error bubble up to the user exactly?

src/handlers/rfqm_handlers.ts
```diff
@@ -0,0 +1,63 @@
...
+    }
+
+    // Parse string params
+    const { takerAddress } = req.query;
+
+    // Parse tokens and eth wrap/unwraps
+    const sellTokenRaw = req.query.sellToken as string;
+    const buyTokenRaw = req.query.buyToken as string;
+    const isETHSell = isETHSymbolOrAddress(sellTokenRaw);
```
PirosB3, line 37: ETH sells or buys cannot be done on RFQM

src/handlers/rfqm_handlers.ts
```diff
@@ -0,0 +1,63 @@
...
+    const isETHBuy = isETHSymbolOrAddress(buyTokenRaw);
+    // TODO - should we ignore ETH?
+    // NOTE: Internally all ETH trades are for WETH, we just wrap/unwrap automatically
```
PirosB3, line 40: This cannot be done with RFQM, metatransactions don't wrap or unwrap ETH

src/handlers/rfqm_handlers.ts
```diff
@@ -0,0 +1,63 @@
...
+    constructor(public readonly _rfqmService: RfqmService) {}
+
+    public async getIndicativeQuoteAsync(req: express.Request, res: express.Response): Promise<void> {
+        const params = parseFetchIndicativeQuoteParams(req);
+        const indicativeQuote = await this._rfqmService.fetchIndicativeQuoteAsync(params);
+        res.status(HttpStatus.OK).send(indicativeQuote);
+    }
+}
+
+const parseFetchIndicativeQuoteParams = (req: express.Request): FetchIndicativeQuoteParams => {
```
PirosB3, line 23: Things that I'd love to validate, that I think should be added to here:
- Make sure that the API key is RFQM-enabled
- Make sure the request is not for buy/sell ETH

src/handlers/rfqm_handlers.ts
```diff
@@ -0,0 +1,63 @@
...
+import { FetchIndicativeQuoteParams } from '../types';
+import { schemaUtils } from '../utils/schema_utils';
+import { findTokenAddressOrThrowApiError, isETHSymbolOrAddress } from '../utils/token_metadata_utils';
+
+export class RfqmHandlers {
+    constructor(public readonly _rfqmService: RfqmService) {}
+
+    public async getIndicativeQuoteAsync(req: express.Request, res: express.Response): Promise<void> {
+        const params = parseFetchIndicativeQuoteParams(req);
+        const indicativeQuote = await this._rfqmService.fetchIndicativeQuoteAsync(params);
```
PirosB3, line 18: If an exception is raised here, how will it bubble up to the user?

PirosB3 (changes requested):
Awesome! I did my first batch of review here. Happy to work together if that helps solve some of the issues :)

src/handlers/rfqm_handlers.ts
PirosB3: - Make sure `takerAddress` is not a smart contract

src/runners/http_rfqm_service_runner.ts
phil-ociraptor: will remove

src/runners/http_rfqm_service_runner.ts
phil-ociraptor: will remove

src/handlers/rfqm_handlers.ts
phil-ociraptor: discussed error handling strategy offline, will implement more specific errors. HTTP errors at the handler level, business logic errors at the app/service level (which are caught and rethrown at the handler level)

src/handlers/rfqm_handlers.ts
phil-ociraptor: - 👍 
- 👍 
- skip for now, but add a TODO

src/handlers/rfqm_handlers.ts
phil-ociraptor: the catching and rethrowing logic goes here

src/handlers/rfqm_handlers.ts
phil-ociraptor: Was about to add the TODO, but then realized that an increasing number of smart contract wallets might cause issues with this assumption. Could be a good parking lot item to discuss

src/handlers/rfqm_handlers.ts
```diff
@@ -0,0 +1,86 @@
...
+import { schemas } from '../schemas';
+import { EMPTY_QUOTE_RESPONSE, FetchIndicativeQuoteParams, RfqmService } from '../services/rfqm_service';
+import { ConfigManager } from '../utils/config_manager';
+import { schemaUtils } from '../utils/schema_utils';
+import { findTokenAddressOrThrowApiError } from '../utils/token_metadata_utils';
+
+export class RfqmHandlers {
+    constructor(private readonly _rfqmService: RfqmService, private readonly _configManager: ConfigManager) {}
+
+    public async getIndicativeQuoteAsync(req: express.Request, res: express.Response): Promise<void> {
```
PirosB3, line 17: Could we add Prometheus counters for "not found" and Prometheus counters for every request?

src/handlers/rfqm_handlers.ts
```diff
@@ -0,0 +1,86 @@
...
+            buyAmount,
+            buyToken,
+            sellAmount,
+            sellToken,
+            takerAddress: takerAddress as string,
+        };
+    }
+}
+
+const validateNotNativeToken = (token: string, field: string): boolean => {
```
PirosB3, line 74: Since this function throws, can you call it `validateNotNativeTokenOrThrow`?

src/handlers/rfqm_handlers.ts
```diff
@@ -0,0 +1,86 @@
...
+        const sellTokenRaw = req.query.sellToken as string;
+        const buyTokenRaw = req.query.buyToken as string;
+        validateNotNativeToken(sellTokenRaw, 'sellToken');
+        validateNotNativeToken(buyTokenRaw, 'buyToken');
+
+        const sellToken = findTokenAddressOrThrowApiError(sellTokenRaw, 'sellToken', CHAIN_ID).toLowerCase();
+        const buyToken = findTokenAddressOrThrowApiError(buyTokenRaw, 'buyToken', CHAIN_ID).toLowerCase();
+
+        // Parse number params
+        const sellAmount =
```
PirosB3, line 59: Since you validated the schema above using JSON schema, you can avoid the if/else and assume that `sellAmount` exists

PirosB3: Same as `buyAmount`

src/handlers/rfqm_handlers.ts
```diff
@@ -0,0 +1,86 @@
...
+        const params = this._parseFetchIndicativeQuoteParams(req);
+        let indicativeQuote;
+        try {
+            indicativeQuote = await this._rfqmService.fetchIndicativeQuoteAsync(params);
+        } catch (err) {
+            req.log.error(err, 'Encountered an error while fetching an rfqm indicative quote');
+            throw new NotFoundError('Unablie to retrieve a price');
+        }
+
+        if (indicativeQuote === EMPTY_QUOTE_RESPONSE) {
```
PirosB3, line 27: After our previous discussion, this can simply change to `if indicativeQuote === null` ..

src/handlers/rfqm_handlers.ts
```diff
@@ -0,0 +1,86 @@
...
+    constructor(private readonly _rfqmService: RfqmService, private readonly _configManager: ConfigManager) {}
+
+    public async getIndicativeQuoteAsync(req: express.Request, res: express.Response): Promise<void> {
+        const params = this._parseFetchIndicativeQuoteParams(req);
+        let indicativeQuote;
+        try {
+            indicativeQuote = await this._rfqmService.fetchIndicativeQuoteAsync(params);
+        } catch (err) {
+            req.log.error(err, 'Encountered an error while fetching an rfqm indicative quote');
+            throw new NotFoundError('Unablie to retrieve a price');
```
PirosB3, line 24: I think we should be able to differentiate between 404 errors (when orders are not found), other 4xx errors (when something is invalid somewhere), and 5xx errors that are server errors

PirosB3 (changes requested):

src/handlers/rfqm_handlers.ts
phil-ociraptor: Will use the default `throw new Error` for the unexpected 500 errors. Over time, as we see patterns with those errors, we can codify them into proper error handling, but for now, I think this is sufficient to start.

src/handlers/rfqm_handlers.ts
```diff
@@ -0,0 +1,127 @@
...
+    public async getIndicativeQuoteAsync(req: express.Request, res: express.Response): Promise<void> {
+        RFQM_INDICATIVE_QUOTE_REQUEST.labels(req.header('0x-api-key') || 'N/A').inc();
+        const params = this._parseFetchIndicativeQuoteParams(req);
+        let indicativeQuote;
+        try {
+            indicativeQuote = await this._rfqmService.fetchIndicativeQuoteAsync(params);
+        } catch (err) {
+            req.log.error(err, 'Encountered an error while fetching an rfqm indicative quote');
+            RFQM_INDICATIVE_QUOTE_ERROR.labels(req.header('0x-api-key') || 'N/A').inc();
+            throw new Error('Unexpected error encountered');
```
PirosB3, line 45: Does this return JSON? I'd love to make sure that there is consistency in our output. If we return JSON when a response is successful, we should also try to respond with JSON when an error is encountered.

Can be as simple as:
```
{
    "error": errorMessage,
}
```

src/handlers/rfqm_handlers.ts
```diff
@@ -0,0 +1,127 @@
...
+    name: 'rfqm_handler_indicative_quote_error',
+    help: 'Request to fetch rfqm indicative quote resulted in error',
+    labelNames: ['apiKey'],
+});
+
+export class RfqmHandlers {
+    constructor(private readonly _rfqmService: RfqmService, private readonly _configManager: ConfigManager) {}
+
+    public async getIndicativeQuoteAsync(req: express.Request, res: express.Response): Promise<void> {
+        RFQM_INDICATIVE_QUOTE_REQUEST.labels(req.header('0x-api-key') || 'N/A').inc();
```
PirosB3, line 37: I notice a little duplication here with API key label, could we create a variable for this?

src/handlers/rfqm_handlers.ts
```diff
@@ -0,0 +1,127 @@
...
+        const { takerAddress } = req.query;
+
+        // Parse tokens
+        const sellTokenRaw = req.query.sellToken as string;
+        const buyTokenRaw = req.query.buyToken as string;
+        validateNotNativeTokenOrThrow(sellTokenRaw, 'sellToken');
+        validateNotNativeTokenOrThrow(buyTokenRaw, 'buyToken');
+
+        const { tokenAddress: sellToken, decimals: sellTokenDecimals } = getTokenMetadataOrThrow(
+            sellTokenRaw,
```
PirosB3, line 78: If the function throws here, what would it look like to the end user? Will it be a 5xx or a 4xx? Based on what I see, it would be a 5xx because it run through the "exception" return path.

Can we make sure that a 404 error is bubbled back to the user as 404?

src/utils/config_manager.ts
```diff
@@ -0,0 +1,13 @@
+// tslint:disable: prefer-function-over-method
+import { RFQM_API_KEY_WHITELIST } from '../config';
+
+/**
+ * ConfigManager is a simple wrapper around configs.
+ *
+ * It exists to provide a layer around our configs, which can then be mocked while writing tests
+ */
+export class ConfigManager {
```
PirosB3, line 9: I'm curious why you ended up creating a class here? It looks like the class has no local state. Have you considered simply using the whitelist constant in the code without `ConfigManager`?

PirosB3: I'm not opposed to it, I just noticed that there isn't any state on this class and therefore I am curious if we could simplify things by removing this class/file. Maybe I'm not seeing things right? Let me know your opinion.

PirosB3 (changes requested):
I think we are close! I left a few more comments. Most importantly, I think we do need some end-to-end testing here. A few things that come to my mind that we can test for:

- Test that buys and sells work correctly
- Ensure that `isLastLook` is passed to the makers, as expected
- When a token is not supported, make sure response is 4xx
- When an invalid API key is submitted, make sure response is 4xx
- When an unexpected error occurs, make sure response is 5xx

PirosB3:
hey @phil-ociraptor if it helps we should probably go ahead and merge this code into the main service branch?

src/utils/config_manager.ts
phil-ociraptor: I added this class to mock out `RFQM_API_KEY_WHITELIST` - otherwise, I would need to gain access to the environment variables before the tests run, which is tricky - not to mention a bit hard to read for future test maintainers.

In other words, this class only exists for testability. Otherwise, I do agree, it's not particularly useful

phil-ociraptor:
> I think we are close! I left a few more comments. Most importantly, I think we do need some end-to-end testing here. A few things that come to my mind that we can test for:
> 
> * Test that buys and sells work correctly
> * Ensure that `isLastLook` is passed to the makers, as expected
> * When a token is not supported, make sure response is 4xx
> * When an invalid API key is submitted, make sure response is 4xx
> * When an unexpected error occurs, make sure response is 5xx

Regarding the e2e tests, most of these e2e tests do exist presently in `rfqm_test.ts`, with the exception of returning a 500 when the response is unexpected. i.e. tests that look like this:
```
                    const appResponse = await request(app)
                        .get(`${RFQM_PATH}/price?${params.toString()}`)
                        .set('0x-api-key', API_KEY)
                        .expect(HttpStatus.NOT_FOUND)
                        .expect('Content-Type', /json/);

                    expect(appResponse.body.reason).to.equal('Not Found');
```

I can update their names so that they are more specific to the error code (good idea!). These integration tests also indirectly require that `isLastLook = true` - otherwise they will fail.

There is additional coverage in `rfqm_service_test.ts` for the more business logic related items (i.e. specific sections for buys and sells).

I'll add the test for the 500 and update the test names to be more obvious that they're testing HTTP error codes, but let me know if there's any other test you'd like to add!

src/handlers/rfqm_handlers.ts
phil-ociraptor: Ah, yes, can add another test for this one! Currently only have a test for WETH vs ETH, but not for a nonexistent token.

src/handlers/rfqm_handlers.ts
phil-ociraptor: Yup - just added a test for this specific case

src/handlers/rfqm_handlers.ts
phil-ociraptor: Nice catch

src/utils/config_manager.ts
PirosB3: that's a good reason! it's fine

src/runners/http_rfqm_service_runner.ts
```diff
@@ -0,0 +1,125 @@
...
+        const protocolFeeUtils = ProtocolFeeUtils.getInstance(
+            PROTOCOL_FEE_UTILS_POLLING_INTERVAL_IN_MS,
+            ETH_GAS_STATION_API_URL,
+        );
+        const metaTxWorkerRegistry = META_TX_WORKER_REGISTRY || NULL_ADDRESS;
+        const rfqmService = new RfqmService(quoteRequestor, protocolFeeUtils, contractAddresses, metaTxWorkerRegistry);
+
+        const configManager = new ConfigManager();
+
+        await runHttpRfqmServiceAsync(rfqmService, configManager, config);
```
PirosB3, line 77: Just checking, can you confirm that Prometheus is working correctly? are you able to read the metrics exposed by Prometheus? I noticed some runners need to explicitly enable Prometheus, do we need to do the same here?

PirosB3 (approved):
Thank you for this incredible work! and I appreciate you addressing all my comments. I have 2 more comments for you. After this, as long as we have done regression and as long as we are able to run `price` endpoint correctly, I would say let's MERGE!

src/runners/http_rfqm_service_runner.ts
phil-ociraptor: verified that `createDefaultServer` will do this automatically: https://github.com/0xProject/api-utils/blob/main/src/default_server.ts#L72

PR #544, opened 2021-04-16
feat: don't allow matcha to skip validation of firm quotes

alexkroeger:
# Description

Validation ensures that the quote provided will be valid for the provided taker.

There is a param that allows a swap quote request to skip validation checks. These changes set this to false always for Matcha.

# Testing Instructions

# Checklist

    -   [ ] SRA/Limit orders
    -   [ ] Swap endpoints
    -   [ ] Depth charts

### DIFF
--- a/src/config.ts
+++ b/src/config.ts
@@ -110,6 +110,28 @@ export const getApiKeyWhitelistWithFallback = (
     return result.sort();
 };
 
+export const getApiKeyFromLabel = (
+    label: string,
+): string | undefined => {
+    if (process.env.API_KEYS_ACL === undefined) {
+        return undefined;
+    }
+    let deserialized: ApiKeyStructure;
+    try {
+        deserialized = JSON.parse(process.env.API_KEYS_ACL);
+        schemaUtils.validateSchema(deserialized, schemas.apiKeySchema as any);
+    } catch (e) {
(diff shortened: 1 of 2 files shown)


src/config.ts
```diff
@@ -110,6 +110,28 @@ export const getApiKeyWhitelistWithFallback = (
...
 
+export const getApiKeyFromLabel = (
+    label: string,
+): string | undefined => {
+    if (process.env.API_KEYS_ACL === undefined) {
+        return undefined;
+    }
+    let deserialized: ApiKeyStructure;
+    try {
+        deserialized = JSON.parse(process.env.API_KEYS_ACL);
```
phil-ociraptor, line 121: Is there a way we could avoid paying the deserialization cost on every call to this function? 

Maybe just make this object value an exported constant or something

phil-ociraptor: We could actually generate two lookup maps at start: 

- One by api key (which is the default)
- Another map by the label value

alexkroeger: that's a good call, I think we'd only call this function once on startup tho

phil-ociraptor (approved):

alexkroeger:
deploy staging

PirosB3 (approved):

PR #550, opened 2021-04-21
feat: add RfqmService, implement fetching indicative RFQM quotes [MKR-29]

phil-ociraptor:
# Description

Introduce the `RfqmService`, which will become the coordination layer for HTTP based RFQM flows (aka, everything except the queue consumer/worker)

This PR is a second version of #535, which entered a corrupted state after a rebase.

# Testing Instructions

Unit tests provided. Introduced a new library called ts-mockito, it's a noticeable improvement over Sinon. In future commits, will introduce more integration tests, which is the standard form of testing in 0x-api.

# Checklist

    -   [ ] SRA/Limit orders
    -   [ ] Swap endpoints
    -   [ ] Depth charts


phil-ociraptor:
@PirosB3  - comments from last PR that have been addressed:

- [x] worker registry as an environment variable in `config.ts`
- [x] return the empty quote instead of throwing a generic error
- [x] added a check for expiration at least 3 minutes out
- [x] support buys and sells
- [x] use a mixin for the default rfqm opts
- [x] add `gas` and `allowanceTarget` to response

src/services/rfqm_service.ts
```diff
@@ -0,0 +1,130 @@
+import { AssetSwapperContractAddresses, MarketOperation, ProtocolFeeUtils, QuoteRequestor } from '@0x/asset-swapper';
+import { BigNumber } from '@0x/utils';
+
+import { META_TX_WORKER_REGISTRY, RFQT_REQUEST_MAX_RESPONSE_MS } from '../config';
+import { NULL_ADDRESS, ONE_MINUTE_MS } from '../constants';
+
+export interface FetchIndicativeQuoteParams {
```
alexkroeger, line 7: we'll need a fee param potentially, worth adding now?

src/services/rfqm_service.ts
```diff
@@ -0,0 +1,130 @@
...
+    apiKey: string;
+    buyAmount?: BigNumber;
+    sellAmount?: BigNumber;
+    buyToken: string;
+    sellToken: string;
+    takerAddress?: string;
+}
+
+export interface FetchIndicativeQuoteResponse {
```
alexkroeger, line 16: same comment about fee

src/services/rfqm_service.ts
```diff
@@ -0,0 +1,130 @@
...
+        const marketOperation = isSelling ? MarketOperation.Sell : MarketOperation.Buy;
+        const assetFillAmount = isSelling ? sellAmount! : buyAmount!;
+        const makerToken = buyToken;
+        const takerToken = sellToken;
+
+        // Fetch quotes
+        const opts = {
+            ...RFQM_DEFAULT_OPTS,
+            apiKey,
+            intentOnFilling: false, // TODO - safe to hardcode?
```
alexkroeger, line 72: should be fine

src/services/rfqm_service.ts
```diff
@@ -0,0 +1,130 @@
...
+
+        // No quotes found
+        if (sortedQuotes.length === 0) {
+            return Promise.resolve(EMPTY_QUOTE_RESPONSE);
+        }
+
+        const bestQuote = sortedQuotes[0];
+
+        // Prepare gas estimate
+        const gas = await this._protocolFeeUtils.getGasPriceEstimationOrThrowAsync();
```
alexkroeger, line 114: what value is returned here?

alexkroeger: it this is async can we request this in parallel?

src/services/rfqm_service.ts
```diff
@@ -0,0 +1,130 @@
...
+    takerAddress?: string;
+}
+
+export interface FetchIndicativeQuoteResponse {
+    buyAmount: BigNumber;
+    buyTokenAddress: string;
+    price: BigNumber;
+    sellAmount: BigNumber;
+    sellTokenAddress: string;
+    gas: BigNumber;
```
alexkroeger, line 22: what is the target value here?

src/config.ts
```diff
@@ -297,17 +300,21 @@ export const ALT_RFQ_MM_PROFILE: string | undefined = _.isEmpty(process.env.ALT_
...
     : assertEnvVarType(
           'RFQT_MAKER_ASSET_OFFERINGS',
           process.env.RFQT_MAKER_ASSET_OFFERINGS,
-          EnvVarType.RfqtMakerAssetOfferings,
+          EnvVarType.RfqMakerAssetOfferings,
       );
 
+export const META_TX_WORKER_REGISTRY: string | undefined = _.isEmpty(process.env.META_TX_WORKER_REGISTRY)
+    ? undefined
+    : assertEnvVarType('META_TX_WORKER_REGISTRY', process.env.META_TX_WORKER_REGISTRY, EnvVarType.NonEmptyString);
```
PirosB3, line 316: Can you use type `EthAddressHex`?

src/services/rfqm_service.ts
```diff
@@ -0,0 +1,130 @@
...
+    allowanceTarget?: string;
+}
+
+export const EMPTY_QUOTE_RESPONSE = {
+    buyAmount: new BigNumber(0),
```
PirosB3, line 27: Can you use the constant `ZERO`? It should be available within the 0x API

src/services/rfqm_service.ts
```diff
@@ -0,0 +1,130 @@
...
+    buyTokenAddress: '',
```
PirosB3, line 28: Can you simply omit invalid fields here?

src/services/rfqm_service.ts
```diff
@@ -0,0 +1,130 @@
...
+    constructor(
+        private readonly _quoteRequestor: QuoteRequestor,
+        private readonly _protocolFeeUtils: ProtocolFeeUtils,
+        private readonly _contractAddresses: AssetSwapperContractAddresses,
+    ) {}
+
+    /**
+     * Fetch the best indicative quote available.
+     */
+    public async fetchIndicativeQuoteAsync(params: FetchIndicativeQuoteParams): Promise<FetchIndicativeQuoteResponse> {
```
PirosB3, line 58: Instead of returning an empty quote (with invalid fields) can we simply return undefined or null?

src/services/rfqm_service.ts
phil-ociraptor: Chat w/ @PirosB3 - won't add the fee param now, but will keep the `gas` value as the fee is likely to be dependent on the gas value

src/services/rfqm_service.ts
```diff
@@ -0,0 +1,130 @@
...
+        const { sellAmount, buyAmount, sellToken, buyToken, apiKey } = params;
+
+        // Map params to the terminology of Quote Requestor
+        const isSelling = sellAmount !== undefined;
+        const marketOperation = isSelling ? MarketOperation.Sell : MarketOperation.Buy;
+        const assetFillAmount = isSelling ? sellAmount! : buyAmount!;
+        const makerToken = buyToken;
```
PirosB3, line 65: Sorry @phil-ociraptor I forgot how this works, bear with me. Does this code already account for buys and sells? For example, `makerToken` is `ETH` if the user is selling `DAI` for `ETH`. However the `makerToken` will be `DAI` if the user is buying `DAI` for `ETH`.

src/services/rfqm_service.ts
```diff
@@ -0,0 +1,130 @@
...
+        private readonly _quoteRequestor: QuoteRequestor,
+        private readonly _protocolFeeUtils: ProtocolFeeUtils,
+        private readonly _contractAddresses: AssetSwapperContractAddresses,
+    ) {}
+
+    /**
+     * Fetch the best indicative quote available.
+     */
+    public async fetchIndicativeQuoteAsync(params: FetchIndicativeQuoteParams): Promise<FetchIndicativeQuoteResponse> {
+        const { sellAmount, buyAmount, sellToken, buyToken, apiKey } = params;
```
PirosB3, line 59: Can we raise an exception if the metatransaction registry address is not set?

src/services/rfqm_service.ts
```diff
@@ -0,0 +1,130 @@
...
+        // And sort by best price
+        const now = new BigNumber(Date.now());
+        const sortedQuotes = indicativeQuotes
+            .filter((q) => q.takerToken === takerToken && q.makerToken === makerToken)
+            .filter((q) => {
+                const requestedAmount = isSelling ? q.takerAmount : q.makerAmount;
+                return requestedAmount.gte(assetFillAmount);
+            })
+            // tslint:disable-next-line: custom-no-magic-numbers
+            .filter((q) => q.expiry.gte(now.plus(ONE_MINUTE_MS * 3)))
```
PirosB3, line 98: Can we create a new constant for this `ONE_MINUTE_MS * 3`?

src/services/rfqm_service.ts
```diff
@@ -0,0 +1,130 @@
...
+            .sort((a, b) => {
+                // Want the most amount of maker tokens for each taker token
+                const aPrice = a.makerAmount.div(a.takerAmount);
+                const bPrice = b.makerAmount.div(b.takerAmount);
+                return bPrice.minus(aPrice).toNumber();
+            });
+
+        // No quotes found
+        if (sortedQuotes.length === 0) {
+            return Promise.resolve(EMPTY_QUOTE_RESPONSE);
```
PirosB3, line 108: Can we throw a `QuoteNotFoundError` or return undefined? I'd rather not return anything than return something that looks valid but is effectively invalid.

PirosB3 (changes requested):

src/services/rfqm_service.ts
phil-ociraptor: removing this obj

src/services/rfqm_service.ts
phil-ociraptor: removing this obj

src/services/rfqm_service.ts
phil-ociraptor: Yup, returning null

src/services/rfqm_service.ts
phil-ociraptor: Resolved offline

src/services/rfqm_service.ts
phil-ociraptor: I prefer doing this check outside of this class - having the check here would make the class hard to unit test (requires refactoring)

I'm thinking an ideal place for it might even be in the `http_runner` where we can ensure that the application has all of the necessary environment variables to ensure execution

src/services/rfqm_service.ts
PirosB3: Hey @alexkroeger just wanted to follow up here. We are going to get this code merged first and then quickly follow up with the fee upgrade. I'm biasing towards getting some code in since we have plenty of pending code.

src/services/rfqm_service.ts
```diff
@@ -0,0 +1,136 @@
...
+            return null;
+        }
+
+        // Get the best quote
+        const bestQuote = sortedQuotes[0];
+
+        // Prepare the price
+        const makerAmountInUnit = Web3Wrapper.toUnitAmount(bestQuote.makerAmount, makerTokenDecimals);
+        const takerAmountInUnit = Web3Wrapper.toUnitAmount(bestQuote.takerAmount, takerTokenDecimals);
+        const price = makerAmountInUnit.div(takerAmountInUnit);
```
PirosB3, line 123: Just checking - is price always maker over taker or could it be the other way round? How do other services (like swap service) do it?

PirosB3 (changes requested):

src/services/rfqm_service.ts
phil-ociraptor: Good call! this is not how swap service currently handles it - will update to take into account selling vs buying

src/services/rfqm_service.ts
```diff
@@ -0,0 +1,136 @@
...
+    buyTokenAddress: string;
+    gas: BigNumber;
+    price: BigNumber;
+    sellAmount: BigNumber;
+    sellTokenAddress: string;
+}
+
+const RFQM_DEFAULT_OPTS = {
+    takerAddress: NULL_ADDRESS,
+    txOrigin: META_TX_WORKER_REGISTRY || NULL_ADDRESS,
```
PirosB3, line 32: As discussed, can we raise an error if no metatransaction registry is set?

PirosB3 (approved):

PR #573, opened 2021-05-11
feat: add rfqm /quote endpoint [MKR-39]

phil-ociraptor:
# Description

Add the `/quote` endpoint for the rfqm flow. This is largely the same as the `/price` endpoint.

# Testing Instructions

Unit and integration tests added

# Checklist

    -   [ ] SRA/Limit orders
    -   [ ] Swap endpoints
    -   [ ] Depth charts


src/handlers/rfqm_handlers.ts
```diff
@@ -66,6 +84,60 @@ export class RfqmHandlers {
         res.status(HttpStatus.OK).send(indicativeQuote);
     }
 
+    public async getFirmQuoteAsync(req: express.Request, res: express.Response): Promise<void> {
+        const apiKeyLabel = req.header('0x-api-key') || 'N/A';
```
alexkroeger, line 88: should we just throw an error here if there's no api key?

src/services/rfqm_service.ts
```diff
@@ -28,6 +30,29 @@ export interface FetchIndicativeQuoteResponse {
     sellAmount: BigNumber;
     sellTokenAddress: string;
 }
+export interface FetchFirmQuoteParams {
+    apiKey: string;
+    buyAmount?: BigNumber;
+    buyToken: string;
+    buyTokenDecimals: number;
+    gasPrice?: BigNumber;
```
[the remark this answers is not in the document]
phil-ociraptor: yeah good point! I copied this logic from the `/swap` endpoint, but you're right, this doesn't make sense. Will remove

src/services/rfqm_service.ts
```diff
@@ -28,6 +30,29 @@ export interface FetchIndicativeQuoteResponse {
...
+    sellAmount?: BigNumber;
+    sellToken: string;
+    sellTokenDecimals: number;
+    takerAddress: string;
+}
+
+export interface FetchFirmQuoteResponse {
```
alexkroeger, line 45: I think we should make this more generic, as we may want to change the order format.

Instead of `metaTransaction` I would propose a `payload` key, that could be a MetaTransaction object but could be something different, like an order to sign.

I think we should return the order hash as well, and the `metaTranactionHash` field could be optional. It might even be worth adding a new key called `identifiers` which could look like:

```
identifiers: {
    orderHash: string;
    metaTransactionHash?: string;
}
```

src/services/rfqm_service.ts
```diff
@@ -130,4 +156,97 @@ export class RfqmService {
...
+        );
+
+        // Get the best quote
+        const bestQuote = getBestQuote(
+            firmQuotes,
+            isSelling,
+            takerToken,
+            makerToken,
+            assetFillAmount,
+            RFQM_MINUMUM_EXPIRY_DURATION_MS,
```
alexkroeger, line 213: typo here:

```suggestion
            RFQM_MINIMUM_EXPIRY_DURATION_MS,
```

src/services/rfqm_service.ts
```diff
@@ -130,4 +156,97 @@ export class RfqmService {
...
+        // Generate the Meta Transaction and its hash
+        const rfqOrder = bestQuote.order as RfqOrder;
+        const metaTransaction = this._blockchainUtils.generateMetaTransaction(
+            rfqOrder,
+            bestQuote.signature,
+            takerAddress,
+            bestQuote.order.takerAmount,
+            CHAIN_ID,
+        );
+        const metaTransactionHash = metaTransaction.getHash();
```
alexkroeger, line 230: Is it worth checking here that the metaTransaction is fillable?

Once the taker signs and submits, they'll expect a fill.

src/handlers/rfqm_handlers.ts
alexkroeger: where do we check that the API is valid?

test/rfqm_test.ts
```diff
@@ -476,4 +490,179 @@ describe(SUITE_NAME, () => {
             );
         });
     });
+
+    describe('rfqm/v1/quote', async () => {
```
alexkroeger, line 494: should test for invalid API key

src/handlers/rfqm_handlers.ts
phil-ociraptor: regarding throwing an error: that's a good idea, I can't think of a usecase where we would allow for rfqm liquidity without a key. Will add this check

regarding the API key validity - this is some kind of hack code, but it happens in `_parseFetchIndicativeQuoteParams`, which is used by `_parseFetchFirmQuoteParams`

src/services/rfqm_service.ts
phil-ociraptor: This is a really good idea - will noodle on this a bit. I think there's two "flavors" of RFQ "M" to consider where:

- M = metatransaction, where the metatransaction needs to be signed
- M = maker, where the roles are flipped vs RFQT, and as you mentioned, an order needs to be signed

The identifiers also does make a lot of sense. I like that too.

src/services/rfqm_service.ts
phil-ociraptor: LUL

src/services/rfqm_service.ts
phil-ociraptor: Yes, you're right, we should check early. This also opens a small question, which is, if the metaTransaction is not fillable and there are other metaTransactions from MMs, should we try the others as fallbacks? How many fallbacks to consider?

Will see if there's a simple solution that can be coded up

src/handlers/rfqm_handlers.ts
alexkroeger: cool, I would prefer both `_parseFetchFirmQuoteParams` and `_parseFetchIndicativeQuoteParams` call some third function that contains common logic, but not necessary

alexkroeger (approved):
👍

PR #577, opened 2021-05-14
feat: Add RfqmQuote  [MKR-42]

phil-ociraptor:
# Description

# Testing Instructions

I have tested the change locally, including the migration, but would be highly preferable to try running these migrations against something like staging.

Unit and integration tests are working.

# Checklist

    -   [ ] SRA/Limit orders
    -   [ ] Swap endpoints
    -   [ ] Depth charts


src/entities/index.ts
```diff
@@ -1,8 +1,9 @@
...
-export { TransactionEntity } from './TransactionEntity';
 export { KeyValueEntity } from './KeyValueEntity';
-export { TransactionEntityOpts } from './types';
-export { PersistentSignedOrderEntity } from './PersistentSignedOrderEntity';
 export { MakerBalanceChainCacheEntity } from './MakerBalanceChainCacheEntity';
-export { SignedOrderV4Entity } from './SignedOrderV4Entity';
+export { PersistentSignedOrderEntity } from './PersistentSignedOrderEntity';
 export { PersistentSignedOrderV4Entity } from './PersistentSignedOrderV4Entity';
+export { RfqmQuoteEntity } from './RfqmQuoteEntity';
+export { SignedOrderEntity } from './SignedOrderEntity';
```
alexkroeger, line 6: just a re-ordering?

src/services/rfqm_service.ts
```diff
@@ -242,7 +252,17 @@ export class RfqmService {
...
         const metaTransactionHash = metaTransaction.getHash();
 
-        // TODO: Save a record of the metatransactionHash (indexed), orderHash, and fee (a json blob)
+        // TODO: Save the makerUri and integratorId
+        // Save the RfqmQuote
+        await this._connection.getRepository(RfqmQuoteEntity).insert(
+            new RfqmQuoteEntity({
+                orderHash,
+                metaTransactionHash,
+                chainId: CHAIN_ID.toString(),
```
alexkroeger, line 261: are we going with numeric representation here? this will be '1'. might be worth being descriptive, e.g. 'Ethereum Mainnet'

test/rfqm_test.ts
```diff
@@ -97,12 +101,17 @@ describe.skip(SUITE_NAME, () => {
             rfqBlockchainUtilsMock.generateMetaTransaction(anything(), anything(), anything(), anything(), anything()),
         ).thenReturn(MOCK_META_TX);
         const rfqBlockchainUtils = instance(rfqBlockchainUtilsMock);
+
+        const connection = await getDBConnectionAsync();
```
alexkroeger, line 105: we might need to reset the DB at the end to avoid state affecting other tests

migrations/1620879168566-CreateRfqmQuoteTable.ts
```diff
@@ -0,0 +1,34 @@
...
+        await queryRunner.createTable(
+            new Table({
+                name: 'rfqm_quotes',
+                columns: [
+                    { name: 'order_hash', type: 'varchar', isPrimary: true },
+                    { name: 'metatransaction_hash', type: 'varchar', isUnique: true, isNullable: true },
+                    { name: 'created_at', type: 'timestamptz', default: 'NOW()' },
+                    { name: 'chain_id', type: 'varchar' },
+                    { name: 'integrator_id', type: 'varchar', isNullable: true },
+                    { name: 'maker_uri', type: 'varchar', isNullable: true },
```
alexkroeger, line 16: this is essential to RFQM later in the process, could go not nullable

src/entities/RfqmQuoteEntity.ts
```diff
@@ -0,0 +1,52 @@
+import { RfqOrder } from '@0x/asset-swapper';
+import { Column, Entity, Index, PrimaryColumn } from 'typeorm';
+
+import { RfqmFee } from '../services/rfqm_service';
+
+@Entity({ name: 'rfqm_quotes' })
+export class RfqmQuoteEntity {
+    @PrimaryColumn({ name: 'order_hash', type: 'varchar' })
+    public orderHash?: string;
```
alexkroeger, line 9:
```suggestion
    public orderHash: string;
```

src/entities/RfqmQuoteEntity.ts
```diff
@@ -0,0 +1,52 @@
...
+
+    @Column({ name: 'metatransaction_hash', type: 'varchar', nullable: true, unique: true })
+    public metaTransactionHash?: string;
```
alexkroeger, line 12: I think nullable fields can be defined but `null` on read, so we should add `| null` to nullable fields

test/rfqm_test.ts
alexkroeger: I also think it's worth testing an entity read and write to make sure values are being stored and read as expected

src/entities/index.ts
phil-ociraptor: yup, just sorted alphabetically and added the `RfqmQuoteEntity`

test/rfqm_test.ts
phil-ociraptor: Added a read from the DB at the end!

Will add a query to Truncate the table that was added to.

migrations/1620879168566-CreateRfqmQuoteTable.ts
phil-ociraptor: good call, just added some logic to allow for saving the `maker_uri` and made it not nullable

src/entities/RfqmQuoteEntity.ts
phil-ociraptor: This one is a bit funny - by some oddity of how our `ormconfig.ts` works, it really wants this to be `orderHash?`. There's probably a way to make it just be `orderHash`, but not sure if it's worth it at this moment. Will look into this if I have some more time

src/entities/RfqmQuoteEntity.ts
phil-ociraptor: nice, I like that - just updated

migrations/1620879168566-CreateRfqmQuoteTable.ts
```diff
@@ -0,0 +1,34 @@
...
+                    { name: 'chain_id', type: 'varchar' },
+                    { name: 'integrator_id', type: 'varchar', isNullable: true },
+                    { name: 'maker_uri', type: 'varchar' },
+                    { name: 'fee', type: 'jsonb', isNullable: true },
+                    { name: 'order', type: 'jsonb', isNullable: true },
+                ],
+            }),
+        );
+
+        const indices = ['created_at'].map(
```
PirosB3, line 23: Are we also going to add an index on `order_hash`?

src/runners/http_rfqm_service_runner.ts
```diff
@@ -113,12 +118,13 @@ export async function runHttpRfqmServiceAsync(
...
     configManager: ConfigManager,
     config: HttpServiceConfig,
+    connection: Connection,
     _app?: core.Express,
 ): Promise<{ app: express.Application; server: Server }> {
     const app = _app || express();
     app.use(addressNormalizer);
     const server = createDefaultServer(config, app, logger, async () => {
-        /* TODO - clean up DB connection when present */
+        await connection.close();
```
PirosB3, line 127: Just to confirm, the `connection.close()` is run once the server terminates? is that right?

src/services/rfqm_service.ts
```diff
@@ -60,6 +62,12 @@ export interface MetaTransactionRfqmQuoteResponse extends BaseRfqmQuoteResponse
 
 export type FetchFirmQuoteResponse = MetaTransactionRfqmQuoteResponse;
 
+export interface RfqmFee {
```
PirosB3, line 65: Can you import this from `quote-server` package?

src/services/rfqm_service.ts
```diff
@@ -186,8 +195,13 @@ export class RfqmService {
         const marketOperation = isSelling ? MarketOperation.Sell : MarketOperation.Buy;
         const assetFillAmount = isSelling ? sellAmount! : buyAmount!;
 
-        // Prepare gas estimate
+        // Prepare gas estimate and fee
         const gas: BigNumber = await this._protocolFeeUtils.getGasPriceEstimationOrThrowAsync();
+        const fee: RfqmFee = {
+            amount: ZERO,
```
PirosB3, line 201: Just to refresh my memory, we we going to compute the fee in a second PR right?

src/services/rfqm_service.ts
```diff
@@ -242,7 +258,18 @@ export class RfqmService {
         );
         const metaTransactionHash = metaTransaction.getHash();
 
-        // TODO: Save a record of the metatransactionHash (indexed), orderHash, and fee (a json blob)
+        // TODO: Save the integratorId
+        // Save the RfqmQuote
+        await this._connection.getRepository(RfqmQuoteEntity).insert(
```
PirosB3, line 263: Could we add a Prometheus counter for every firm quote request (this could help us get a real-time sense on how many records are being inserted)?

PirosB3 (approved):
FENOMENAL PR! Incredible work, simple and concise. Thank you for being a star.

Just a few minor comments and suggestions, but I feel we are 100% on the same page and this code looks spot on.

src/services/rfqm_service.ts
phil-ociraptor: correct! Wanted to make sure we were able to focus on that one separately

src/services/rfqm_service.ts
phil-ociraptor: ah, yes will do

src/services/rfqm_service.ts
phil-ociraptor: 👍   I had a reason for doing this, but after thinking about it, that reason isn't very good. Will do

src/runners/http_rfqm_service_runner.ts
phil-ociraptor: correct!

migrations/1620879168566-CreateRfqmQuoteTable.ts
phil-ociraptor: so both `order_hash` and `metatransaction_hash` will have indices on them by virtue of being the primary key and being a unique key, respectively. I believe adding an index on top is possible, but redundant

src/services/rfqm_service.ts
```diff
@@ -228,6 +238,12 @@ export class RfqmService {
             return null;
         }
 
+        // Get the makerUri
```
PirosB3, line 241: Apologies I can't remember what we agreed upon, are we going to run the RFQ balance checker process here or did we decide to punt?

src/services/rfqm_service.ts
phil-ociraptor: ended up going with a numeric representation. Keeping the convention that ChainId is a number will reduce confusion. Non EVM chains will likely need much more thought across the stack, so will punt on that

src/services/rfqm_service.ts
phil-ociraptor: I vaguely recall us putting a starter task on our backlog for something similar - doing an eth call on the order to check balances - but not sure if that's the same thing

phil-ociraptor:
deploy staging

PR #580, opened 2021-05-18
feat: add Rfqm SQS consumer

phil-ociraptor:
# Description

Changes for adding a Rfqm SQS Consumer. 

Lots of config related changes and would like to get some feedback on that!

# Testing Instructions

The following has been tested:

- Producing and consuming from a queue
- Retries after failures
- Max retries hit (as specified in SQS) and sent to a Dead Letter Queue
- Multiple instances running at the same time (preventing workers from picking up the same job)

```
yarn dev:service:rfqm_sqs
```

Will need the following configs filled out in `.env`:
```
AWS_REGION='us-east-1'
AWS_ACCESS_KEY='XXX'
AWS_SECRET_ACCESS_KEY='YYY'
RFQM_META_TX_SQS_URL='https://sqs.us-east-1.amazonaws.com/<YOUR_QUEUE>'
```

And will need a way to produce into the queue. (I used the `sqs-producer` library)

# Checklist

    -   [ ] SRA/Limit orders
    -   [ ] Swap endpoints
    -   [ ] Depth charts


xianny (commented):
Hmm this feels a little weird to me. Was it an intentional decision to *not* start the consumer in `http_rfqm_service_runner.ts`? Presumably you'd never want to run the http service/endpoints without also running the consumer...

If `RfqmConsumer` is going to be a class, I think it should hold an instance of the SQS consumer and define the event listeners on instantiation - so we can keep all the logic related to the consumer in one module.

src/config.ts
```diff
@@ -327,6 +327,23 @@ export const META_TX_WORKER_REGISTRY: string | undefined = _.isEmpty(process.env
     ? undefined
     : assertEnvVarType('META_TX_WORKER_REGISTRY', process.env.META_TX_WORKER_REGISTRY, EnvVarType.ETHAddressHex);
 
+export const RFQM_META_TX_SQS_URL: string | undefined = _.isEmpty(process.env.RFQM_META_TX_SQS_URL)
+    ? undefined
+    : assertEnvVarType('RFQM_META_TX_SQS_URL', process.env.RFQM_META_TX_SQS_URL, EnvVarType.Url);
+
+// AWS Configs
+export const AWS_REGION: string | undefined = _.isEmpty(process.env.AWS_REGION)
```
PirosB3, line 335: We don't need these, because they will come automagically from role-based authentication, right @opaolini ?

PirosB3 (changes requested):

src/config.ts
phil-ociraptor: would be great to remove these! Let's see what Oskar says :D 

Although, one consideration I had was on how do we run it locally without env variables - right now I put these here to make it easier to run locally, but there might be a better way

phil-ociraptor:
> If `RfqmConsumer` is going to be a class, I think it should hold an instance of the SQS consumer and define the event listeners on instantiation - so we can keep all the logic related to the consumer in one module.

Great point @xianny! will look into doing this tomorrow

> Was it an intentional decision to not start the consumer in `http_rfqm_service_runner.ts`?

In this situation, yes. It has to do with the fact that each instance of a consumer is going to be stateful (one wallet per consumer), so we'll want to have fairly different scaling strategies for the http side and the consumer side.

src/consumers/rfqm_consumers.ts
```diff
@@ -0,0 +1,14 @@
...
+
+export class RfqmConsumers {
+    /**
+     * processJobAsync takes in a raw message from SQS and processes it as an RfqmJob
+     */
+    // TODO: remove tslint:disables once implemented
+    // tslint:disable: no-empty
+    // tslint:disable: prefer-function-over-method
+    public async processJobAsync(orderHash: string): Promise<void> {
+        logger.info(`DONE  ${orderHash}`);
```
opaolini, line 11:
```suggestion
        logger.info(`finished processing job`, {orderHash});
```

Something like this would be better for structured logs and easier search.

src/config.ts
opaolini: The SDKs do some automatic region determination when running on AWS EC2 and/or Lambda [0], also the SDK will check the `AWS_REGION` env variable on its own, so no need to specify it here. That's also true for all the AWS client configuration parameters actually.

IAM Role will solve the authentication so `ACCESS_KEY` and `SECRET_ACCESS_KEY`won't need to be passed in.

[0] - https://docs.amazonaws.cn/sdk-for-java/latest/developer-guide/region-selection.html

src/runners/rfqm_meta_transaction_consumer_runner.ts
```diff
@@ -0,0 +1,134 @@
...
+if (require.main === module) {
+    (async () => {
+        // Start the Metrics Server
+        startMetricsServer();
+
+        // Build dependencies
+        const provider = providerUtils.createWeb3Provider(defaultHttpServiceWithRateLimiterConfig.ethereumRpcUrl);
+        const contractAddresses = await getContractAddressesForNetworkOrThrowAsync(provider, CHAIN_ID);
+        const quoteRequestor = new QuoteRequestor(
+            RFQT_MAKER_ASSET_OFFERINGS,
```
PirosB3, line 66: Can we just drop RFQT makers here (just for safety purpose)

src/runners/rfqm_meta_transaction_consumer_runner.ts
```diff
@@ -0,0 +1,134 @@
...
+            RFQM_MAKER_ASSET_OFFERINGS,
+            Axios.create(getAxiosRequestConfig()),
+            undefined, // No Alt RFQM offerings at the moment
+            logger.warn.bind(logger),
+            logger.info.bind(logger),
+            SWAP_QUOTER_OPTS.expiryBufferMs,
+        );
+        const protocolFeeUtils = ProtocolFeeUtils.getInstance(
```
PirosB3, line 74: Why do we need this?

src/runners/rfqm_meta_transaction_consumer_runner.ts
```diff
@@ -0,0 +1,134 @@
...
+            PROTOCOL_FEE_UTILS_POLLING_INTERVAL_IN_MS,
+            ETH_GAS_STATION_API_URL,
+        );
+        const metaTxWorkerRegistry = META_TX_WORKER_REGISTRY || NULL_ADDRESS;
```
PirosB3, line 78: We should never use `NULL`, can we raise an error and exit here, if an address is not provided?

src/runners/rfqm_meta_transaction_consumer_runner.ts
```diff
@@ -0,0 +1,134 @@
...
+        await runRfqmMetaTransactionConsumerAsync(rfqmService);
+    })().catch((error) => logger.error(error.stack));
+}
+
+/**
+ * Runs the Rfqm Consumer
+ */
+export async function runRfqmMetaTransactionConsumerAsync(rfqmService: RfqmService): Promise<SqsConsumer> {
+    const consumer = new SqsConsumer({
+        sqs: new SQS({ apiVersion: '2012-11-05' }),
```
PirosB3, line 101: What is this `apiVersion` about?

src/utils/sqs_consumer.ts
```diff
@@ -0,0 +1,102 @@
...
+            .receiveMessage({
+                MaxNumberOfMessages: 1,
+                WaitTimeSeconds: 20, // long polling
+                QueueUrl: this._queueUrl,
+            })
+            .promise();
+
+        // No message
+        if (response === undefined || response.Messages?.length !== 1) {
+            return;
```
PirosB3, line 67: Can we add some logging here justto be verbose? `Fetched xxx messages..`

src/utils/sqs_consumer.ts
```diff
@@ -0,0 +1,102 @@
...
+            logger.error({ err, message }, 'Encountered error while handling message');
+            // Retry message
+            await this._sqs
+                .changeMessageVisibility({
+                    QueueUrl: this._queueUrl,
+                    ReceiptHandle: message.ReceiptHandle!,
+                    VisibilityTimeout: 0,
+                })
+                .promise();
+            return;
```
PirosB3, line 86: Should we sleep here? there could be a risk that this "damaged" worker picks up the same task?

src/runners/rfqm_meta_transaction_consumer_runner.ts
phil-ociraptor: this is used by the `rfqmService` to get the gas price.

src/runners/rfqm_meta_transaction_consumer_runner.ts
phil-ociraptor: In this commit, we agreed to raise an error in the constructor of the `rfqmService` itself: https://github.com/0xProject/0x-api/pull/550/files#r621677486

Technically, an error will get raised, but I think it's better to move the error out of the constructor and into the bootstrap process. Will try doing that

src/runners/rfqm_meta_transaction_consumer_runner.ts
phil-ociraptor: just an AWS thing - to lock in the version of the API https://docs.aws.amazon.com/AWSJavaScriptSDK/latest/AWS/SQS.html

I don't think the API is going to change at all, but just in case it does, we'll be fine

src/utils/sqs_consumer.ts
phil-ociraptor: I thought about it, but we'll be logging quite a bit - every 20 seconds, each worker will emit a log that says "Fetched 0 messages...". Thoughts on the value of that?

src/utils/sqs_consumer.ts
phil-ociraptor: Yup, we can add a short sleep of a few seconds

src/runners/rfqm_meta_transaction_consumer_runner.ts
phil-ociraptor: done

src/utils/sqs_client.ts
```diff
@@ -0,0 +1,47 @@
...
+    public async receiveMessageAsync(): Promise<SQS.Message | null> {
+        const response = await this._sqs
+            .receiveMessage({
+                MaxNumberOfMessages: 1,
+                WaitTimeSeconds: 20, // long polling
+                QueueUrl: this._queueUrl,
+            })
+            .promise();
+
+        if (response === undefined || response.Messages?.length !== 1) {
```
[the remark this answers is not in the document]
PirosB3: So it looks like the `client` is the Lower level component, and the `consumer` is the higher level wrapper that is responsible for the lifecycle of the queue. Is that right?

src/utils/sqs_consumer.ts
```diff
@@ -0,0 +1,89 @@
...
+    private readonly _sqsClient: SqsClient;
+    private readonly _beforeHandle?: () => Promise<boolean>;
+    private readonly _handleMessage: (message: SQS.Types.Message) => Promise<any>;
+    private readonly _afterHandle?: () => Promise<void>;
+    private _isConsuming: boolean;
+
+    constructor(params: {
+        sqsClient: SqsClient;
+        beforeHandle?: () => Promise<boolean>;
+        handleMessage: (message: SQS.Types.Message) => Promise<any>;
```
PirosB3, line 18: Can we save this function callback as a separate type?

src/utils/sqs_consumer.ts
```diff
@@ -0,0 +1,89 @@
...
+        this._afterHandle = params.afterHandle;
+        this._isConsuming = false;
+    }
+
+    public stop(): void {
+        this._isConsuming = false;
+    }
+
+    public async consumeAsync(): Promise<void> {
+        this._isConsuming = true;
```
PirosB3, line 33: What happens if this function gets called twice and `isConsuming` is already true? Can we throw an error?

src/utils/sqs_consumer.ts
PirosB3: Good question, maybe it's not needed? I was thinking it could give us a sense of if the worker got stuck somehow? What do you think?

src/utils/sqs_consumer.ts
```diff
@@ -0,0 +1,89 @@
...
+            let beforeCheck;
+            try {
+                beforeCheck = await this._beforeHandle();
+            } catch (e) {
+                logger.warn(e, 'Error encountered in the preHandle check');
+                return;
+            }
+
+            if (!beforeCheck) {
+                logger.warn('before validation failed');
```
PirosB3, line 51: Do we want to sleep if the validation function failed (or returned `false`)? We know for a fact that this happens when the bot has no ETH or there is a pending transaction, so would it make sense to sleep for a reasonable amount of time like 1s?

src/utils/sqs_consumer.ts
```diff
@@ -0,0 +1,89 @@
...
+        await this._sqsClient.deleteMessageAsync(message.ReceiptHandle!);
+
+        // Run the after hook
+        if (this._afterHandle) {
+            await this._afterHandle();
+        }
+    }
+}
+
+async function sleepAsync(ms: number): Promise<void> {
```
PirosB3, line 85: We have the `delay` package that we often use. Can we use that?

PirosB3 (changes requested):
Incredible work, thanks and sorry for the delay!

src/config.ts
```diff
@@ -331,10 +331,6 @@ export const RFQM_META_TX_SQS_URL: string | undefined = _.isEmpty(process.env.RF
     ? undefined
     : assertEnvVarType('RFQM_META_TX_SQS_URL', process.env.RFQM_META_TX_SQS_URL, EnvVarType.Url);
 
-export const RFQM_META_TX_SQS_REGION: string = _.isEmpty(process.env.SQS_REGION)
```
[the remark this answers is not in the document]
phil-ociraptor: AWS already looks for the `AWS_REGION` env variable so no need for us to add it

src/runners/http_rfqm_service_runner.ts
```diff
@@ -54,55 +52,61 @@ process.on('unhandledRejection', (err) => {
...
+        {}, // No RFQT offerings
+        RFQM_MAKER_ASSET_OFFERINGS,
+        Axios.create(getAxiosRequestConfig()),
+        undefined, // No Alt RFQM offerings at the moment
+        logger.warn.bind(logger),
+        logger.info.bind(logger),
+        SWAP_QUOTER_OPTS.expiryBufferMs,
+    );
+
+    const protocolFeeUtils = ProtocolFeeUtils.getInstance(
```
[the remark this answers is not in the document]
phil-ociraptor: this is needed just to build the rfqmService, in the long run, we might want to have a `gasEstimator` component, but reusing the protocol fee utils is fairly pragmatic

src/runners/http_rfqm_service_runner.ts
```diff
@@ -54,55 +52,61 @@ process.on('unhandledRejection', (err) => {
...
+        PROTOCOL_FEE_UTILS_POLLING_INTERVAL_IN_MS,
+        ETH_GAS_STATION_API_URL,
+    );
+    const metaTxWorkerRegistry = META_TX_WORKER_REGISTRY || NULL_ADDRESS;
```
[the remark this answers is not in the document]
phil-ociraptor: I believe I commented on this previously, and this comment is just coming back, but basically, we made the decision to throw an error in the constructor of the `rfqmService` here: https://github.com/0xProject/0x-api/pull/550/files#r621677486

src/runners/rfqm_worker_runner.ts
```diff
@@ -0,0 +1,96 @@
...
+    const sqsClient = new SqsClient({
+        sqs: new SQS({ apiVersion: '2012-11-05' }),
+        queueUrl: RFQM_META_TX_SQS_URL!,
+    });
+    const consumer = new SqsConsumer({
+        sqsClient,
+        handleMessage: async (message) => {
+            RFQM_JOB_DEQUEUED.inc();
+            const orderHash = message.Body!;
+            return rfqmService.processRfqmJobAsync(orderHash);
```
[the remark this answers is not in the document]
phil-ociraptor: Not necessary, in fact a lint error if added

src/runners/rfqm_worker_runner.ts
```diff
@@ -0,0 +1,96 @@
...
+        },
+        afterHandle: async () => {
+            RFQM_JOB_SUCCEEDED.inc();
```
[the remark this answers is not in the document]
phil-ociraptor: Can do!

src/services/rfqm_service.ts
```diff
@@ -424,6 +425,15 @@ export class RfqmService {
         };
     }
 
+    /**
+     * Process an orderHash as an RfqmJob. Throws an error if job must be retried
+     */
+    // tslint:disable: prefer-function-over-method
+    public async processRfqmJobAsync(orderHash: string): Promise<void> {
+        logger.info({ orderHash }, 'finished processing job');
+        return Promise.reject('failing');
```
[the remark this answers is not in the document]
phil-ociraptor: yup, easy enough

src/utils/sqs_client.ts
```diff
@@ -0,0 +1,47 @@
+import { SQS } from 'aws-sdk';
+
+/**
+ * SqsClient wraps SQS, making it far easier to unit test SQS and ignore SQS details
+ */
+export class SqsClient {
+    private readonly _sqs: SQS;
+    private readonly _queueUrl: string;
+    constructor(params: { sqs: SQS; queueUrl: string }) {
```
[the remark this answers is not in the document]
phil-ociraptor: ya, good idea, this was some copy pasta! can clean it up

src/utils/sqs_client.ts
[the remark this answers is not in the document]
phil-ociraptor: Correct - the `client` exists because it was extremely frustrating to write unit tests mocking out `AWS` itself, so I didn't bother and created a wrapper around it. This allows for the `consumer` logic to be well unit tested

src/utils/sqs_client.ts
```diff
@@ -0,0 +1,47 @@
...
+                MaxNumberOfMessages: 1,
+                WaitTimeSeconds: 20, // long polling
+                QueueUrl: this._queueUrl,
+            })
+            .promise();
+
+        if (response === undefined || response.Messages?.length !== 1) {
+            return null;
+        }
+        return response.Messages[0];
```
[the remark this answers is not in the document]
phil-ociraptor: My preference is to do the schema validation, parsing, etc in the `worker`/`consumer` level. This is analogous to doing `parsing` and validation in the Handler level. 

And then the `service` level is simply business logic, and doesn't matter if it started from the `worker` entrypoint or the `handler` entrypoint.

src/utils/sqs_consumer.ts
phil-ociraptor: sure!

src/utils/sqs_consumer.ts
phil-ociraptor: Good call! I'll exit early. An error is aggressive

src/utils/sqs_consumer.ts
phil-ociraptor: yeah, that's a good idea

src/utils/sqs_consumer.ts
phil-ociraptor: Hmm... if we want to get a sense for if the worker is trying to process something, we could add a request at the top of the workflow. I could see that being useful as a sanity check.

That said, I think we can add some logging in our worker liveness check and worker startup, so I imagine we'll have logging for that when we integrate that code.

src/utils/sqs_consumer.ts
phil-ociraptor: ah, much better - hate having to write this function every time

src/utils/sqs_consumer.ts
```diff
@@ -0,0 +1,91 @@
...
+            return;
+        }
+
+        this._isConsuming = true;
+        while (this._isConsuming) {
+            await this.consumeOnceAsync();
+        }
+    }
+
+    public async consumeOnceAsync(): Promise<void> {
```
PirosB3, line 46: should this be private?

src/utils/sqs_consumer.ts
```diff
@@ -0,0 +1,91 @@
...
+    }) {
+        this._sqsClient = params.sqsClient;
+        this._beforeHandle = params.beforeHandle;
+        this._handleMessage = params.handleMessage;
+        this._afterHandle = params.afterHandle;
+        this._isConsuming = false;
+    }
+
+    public stop(): void {
+        this._isConsuming = false;
```
PirosB3, line 32: throw error if not running?

src/utils/sqs_consumer.ts
phil-ociraptor: I made it public so I could test it more easily. Also figured it might be useful to have a consumer where someone manually handles the consumption as well (maybe they want to consume a predefined number of messages)

src/utils/sqs_consumer.ts
phil-ociraptor: I think it's ok not to. If the consumer is not consuming, then setting it to false is a no-op.

src/utils/sqs_consumer.ts
PirosB3: I see.. so it's public for testing purposes right?

src/utils/sqs_consumer.ts
```diff
@@ -0,0 +1,91 @@
...
+import delay from 'delay';
+
+import { ONE_SECOND_MS } from '../constants';
+import { logger } from '../logger';
+
+import { SqsClient } from './sqs_client';
+
+export type MessageHandler = (message: SQS.Types.Message) => Promise<any>;
+
+export class SqsConsumer {
```
PirosB3, line 11: One small thing: can the SQS Consumer accept a `uid` parameter? `uid` can be a unique indentifier of this wallet (so this can be set to the account address)? Ideally, this parameter is passed in to every log command, so that we can diagnose issues relevant to a particular account

src/utils/sqs_consumer.ts
```diff
@@ -0,0 +1,91 @@
...
+
+        // No message
+        if (message === null) {
+            return;
+        }
+
+        // Handle message
+        try {
+            await this._handleMessage(message);
+        } catch (err) {
```
PirosB3, line 75: Just curious, how can we differentiate here for kinds of errors that are recoverable (errors where we decide to re-try the message in the future or in another worker. Example: network blip) compard to unexpected errors that we do not recover from (order expired, simulation fails). Are the latter not errors in your opinion?

src/runners/rfqm_worker_runner.ts
```diff
@@ -0,0 +1,96 @@
...
+ * Runs the Rfqm Consumer
+ */
+export async function runRfqmWorkerAsync(rfqmService: RfqmService): Promise<SqsConsumer> {
+    // Build the Sqs consumer
+    const sqsClient = new SqsClient(new SQS({ apiVersion: '2012-11-05' }), RFQM_META_TX_SQS_URL!);
+    const consumer = new SqsConsumer({
+        sqsClient,
+        handleMessage: async (message) => {
+            RFQM_JOB_DEQUEUED.inc();
+            const orderHash = message.Body!;
```
PirosB3, line 64: Just to confirm. This `orderHash` is going to be a `string` or `Buffer` right? not an JSON object. Was that the right intention?

src/utils/sqs_client.ts
```diff
@@ -0,0 +1,44 @@
...
+            .changeMessageVisibility({
+                QueueUrl: this._queueUrl,
+                ReceiptHandle: receiptHandle,
+                VisibilityTimeout: visibilityTimeout,
+            })
+            .promise();
+    }
+
+    public async deleteMessageAsync(receiptHandle: string): Promise<void> {
+        await this._sqs
```
PirosB3, line 37: Will this explode if there is a problem deleting the message? what do we expect the behavour to look like?

src/utils/sqs_consumer.ts
phil-ociraptor: correct! Although I could see it being useful by itself as well

src/utils/sqs_consumer.ts
phil-ociraptor: Good question! In the past, I've done the following:
- Create a list of "retryable errors" and "nonretryable errors"
- Catch error. If the type of the error is "retryable", then retry. If not, then not

But I think in our case, we can add that functionality later if we feel that it provides value. I think starting with this stance is pretty pragmatic:
- All errors are retryable
- Except for a certain class of errors that we're expecting. These will terminate the job and put the job in its correct status
- All retries, after a certain number of attempts, will get placed on the DLQ, where we can manually merge them later

That said, we could also add an `onError` handler to the SqsConsumer, which would allow us to implement this logic fairly easily

src/runners/rfqm_worker_runner.ts
phil-ociraptor: ah yeah, this needs to be `JSON.parse`d 

Will update!

src/runners/rfqm_worker_runner.ts
```diff
@@ -0,0 +1,96 @@
...
+        afterHandle: async (message) => {
+            const orderHash = message.Body!;
+            logger.info({ orderHash }, 'job successfully ran');
+            RFQM_JOB_SUCCEEDED.inc();
+        },
+    });
+
+    // Start the consumer - aka the worker
+    consumer.consumeAsync().catch((e) => logger.error(e));
+    logger.info('Rfqm Consumer running');
```
rhinodavid, line 77:
```suggestion
    consumer.consumeAsync()
        .then(() =>   logger.info('Rfqm Consumer running'))
        .catch((e) => logger.error(e));
```

src/services/rfqm_service.ts
```diff
@@ -424,6 +421,16 @@ export class RfqmService {
...
     }
 
+    /**
+     * Process an orderHash as an RfqmJob. Throws an error if job must be retried
+     */
+    // tslint:disable: prefer-function-over-method
+    public async processRfqmJobAsync(orderHash: string): Promise<void> {
+        logger.info({ orderHash }, 'finished processing job');
+        // tslint:disable-next-line: custom-no-magic-numbers
+        return Math.random() > 0.5 ? Promise.reject('failing') : Promise.resolve();
```
rhinodavid, line 431: What does the random logic do?

src/utils/sqs_client.ts
phil-ociraptor: the behavior would be:
- Job in DB is already placed in either `success`, `failed`, or some terminal status
- We fail to delete the message from SQS (i.e. this line blows up)
- Worker dies and gets restarted by k8s
- Message becomes visible after its visibility timeout expires (maybe to another worker)
- We consume the message again

There's no way to entirely prevent against this from happening, but one thing we an do is do some retries. For example, using the https://www.npmjs.com/package/async-retry package.

I actually tried adding that package, but had some issues. As this diff was getting pretty big, I think it makes sense to beef up our systems a bit better in a separate commit

We might also want to define the behavior for what we should do if SQS throws an error. My plan is to kill the worker, but only after N number of retries - which I think should be sufficient

src/utils/sqs_consumer.ts
```diff
@@ -0,0 +1,91 @@
+import { SQS } from 'aws-sdk';
+import delay from 'delay';
+
+import { ONE_SECOND_MS } from '../constants';
+import { logger } from '../logger';
+
+import { SqsClient } from './sqs_client';
+
+export type MessageHandler = (message: SQS.Types.Message) => Promise<any>;
```
rhinodavid, line 9: Can avoid the `any` here (may or may not be worth it):
```
export type MessageHandler<T> = (message: SQS.Types.Message) => Promise<T>;
```

src/services/rfqm_service.ts
phil-ociraptor: this is temporary logic that Daniel and I have added for testing purposes. Essentially, this returns rejects or resolves this flow with 50/50 probability. It allows us to see both the successful and failure code paths.

Will be replaced with an actual impl in the next diff

src/utils/sqs_consumer.ts
```diff
@@ -0,0 +1,91 @@
...
+            return;
+        }
+
+        // Delete message
+        await this._sqsClient.deleteMessageAsync(message.ReceiptHandle!);
+
+        // Run the after hook
+        if (this._afterHandle) {
+            await this._afterHandle(message);
+        }
```
rhinodavid, line 89: Moar flair.

```suggestion
this._afterHandle && await this._afterHandle(message);
```

src/runners/rfqm_worker_runner.ts
phil-ociraptor: ahh, so because `consumeAsync` won't get resolved until we call `consumer.stop` (and for our purposes, we don't want the consumer to ever stop), the `then` block isn't expected to run. (nor is the `catch` block for that matter, but it's there just in case)

The logger line follows the `consumeAsync` invocation - so we start the infinite consume loop, and then we immediately log that we're starting

src/utils/sqs_consumer.ts
phil-ociraptor: yes - this is better!

src/utils/sqs_consumer.ts
phil-ociraptor: I like this!

phil-ociraptor: ah, but sadly, doing this will force `T` to cascade through the Sqs Consumer, which is a bigger change. 

I will keep this in the back of my head though, if we ever need typesafety here, we should do it

src/utils/sqs_consumer.ts
phil-ociraptor: sadly, tslint isn't liking it 😞 , will go back to original

src/runners/rfqm_worker_runner.ts
```diff
@@ -0,0 +1,97 @@
...
+    }
+});
+
+if (require.main === module) {
+    (async () => {
+        // Start the Metrics Server
+        startMetricsServer();
+
+        // Build dependencies
+        const connection = await getDBConnectionAsync();
```
PirosB3, line 46: if there is a DB connection issue, will this LOC throw an error?

src/runners/rfqm_worker_runner.ts
```diff
@@ -0,0 +1,97 @@
...
+ */
+export async function runRfqmWorkerAsync(rfqmService: RfqmService): Promise<SqsConsumer> {
+    // Build the Sqs consumer
+    const sqsClient = new SqsClient(new SQS({ apiVersion: '2012-11-05' }), RFQM_META_TX_SQS_URL!);
+    const consumer = new SqsConsumer({
+        id: 'TODO-replace-with-address',
+        sqsClient,
+        handleMessage: async (message) => {
+            RFQM_JOB_DEQUEUED.inc();
+            const { orderHash } = JSON.parse(message.Body!);
```
PirosB3, line 65: Can we handle improper message encoding? here is an example:

```
let res: any;
try {
    const res = JSON.parse(message.Body!);
} catch (e) {
    log.error("Found unknown payload: {message.Body}");
    // Still mark job as "complete" (we dont want to retry);
    return;
}

if (res.orderHash === undefined) {
        log.error("Found invalid payload: {message.Body}");
    // Still mark job as "complete" (we dont want to retry);
    return;
}
.. continue as normal
```

PirosB3 (approved):
Just a few comments, overall amazing stuff

src/runners/rfqm_worker_runner.ts
phil-ociraptor: Definitely! Although, I will handle this more explicitly in my next commit (the one that implements the `processJobAsync` function)

src/runners/rfqm_worker_runner.ts
phil-ociraptor: I think this will be ok - this function throws an error only if the configs are wrong - and if the configs are wrong, the app will never work.

PR #606, opened 2021-06-01
feat: Filter RFQM quotes with the wrong chain id

rhinodavid:
# Description

Closes MKR-78.

When a MM posts a quote, compare the chain id of the quote with the environment chain id. If they are not the same, do not include the quote in the best quotes and log a message.

# Testing Instructions

Run new unit test.

# Checklist

    -   [ ] SRA/Limit orders
    -   [ ] Swap endpoints
    -   [ ] Depth charts


src/services/rfqm_service.ts
```diff
@@ -270,9 +271,18 @@ export class RfqmService {
             opts,
         );
 
+        const firmQuotesWithCorrectChainId = firmQuotes.reduce((result, quote) => {
+            if (quote.order.chainId !== CHAIN_ID) {
+                logger.debug(`Received a quote with incorrect chain id: ${quote}`);
```
rhinodavid, line 276: Just want to make sure we want this at the `debug` level

test/rfqm_test.ts
```diff
@@ -551,7 +552,7 @@ describe(SUITE_NAME, () => {
                 takerToken: contractAddresses.etherToken,
                 pool: '0x',
                 salt: '0',
-                chainId: 1,
+                chainId: 1337,
```
rhinodavid, line 555: This is the chain id specified in the test config

src/services/rfqm_service.ts
PirosB3: this should always be a scary `error` imo. It should never happen, and if it does we (or a market maker) should adjust immediately.

src/services/rfqm_service.ts
```diff
@@ -270,9 +271,18 @@ export class RfqmService {
             opts,
         );
 
+        const firmQuotesWithCorrectChainId = firmQuotes.reduce((result, quote) => {
```
PirosB3, line 274: Why not use `firmQuotes.filter` ?

test/rfqm_test.ts
```diff
@@ -641,6 +642,108 @@ describe(SUITE_NAME, () => {
...
+                        },
+                        responseCode: 200,
+                        responseData: {
+                            signedOrder: {
+                                ...BASE_SIGNED_ORDER,
+                                makerAmount: correctChainQuote,
+                                takerAmount: sellAmount,
+                                signature: {
+                                    ...VALID_SIGNATURE,
+                                    r: '0xb1',
```
PirosB3, line 719: I think that `r` and `s` needs to be the same length as a legit signature, otherwise it won't pass the JSON schema check

PirosB3 (approved):

test/rfqm_test.ts
rhinodavid: Hmmm... seems to work for me. it's copied from the test above.

src/services/rfqm_service.ts
rhinodavid: good call

src/services/rfqm_service.ts
rhinodavid: done

PR #613, opened 2021-06-02
feat: add implementation for processJobAsync in Rfqm

phil-ociraptor:
# Description

Add the business logic for our RFQM worker - this lives in `processJobAsync`

# Testing Instructions

Basic integration test added. May do some refactoring to make unit testing easier

# Checklist

    -   [ ] SRA/Limit orders
    -   [ ] Swap endpoints
    -   [ ] Depth charts


src/services/rfqm_service.ts
```diff
@@ -432,11 +436,90 @@ export class RfqmService {
...
+
+        // Get job
+        const job = await this.dbUtils.findJobByOrderHashAsync(orderHash);
+        if (job === undefined) {
+            throw new NotFoundError(`job for orderHash ${orderHash} not found`);
+        }
+
+        // Basic validation
+        const { calldata, makerUri, order, fee } = job;
+        if (calldata === undefined) {
```
alexkroeger, line 450: can we throw these checks in another function, e.g. validateJob()?

src/services/rfqm_service.ts
```diff
@@ -432,11 +436,90 @@ export class RfqmService {
...
+        }
+
+        // Start processing
+        await this.dbUtils.updateRfqmJobAsync(orderHash, {
+            status: RfqmJobStatus.Processing,
+        });
+
+        // Validate w/Eth Call
+        try {
+            await this._blockchainUtils.decodeMetaTransactionCallDataAndValidateAsync(calldata, 'TODO Sender');
```
alexkroeger, line 489: this will be called by the worker, which will be aware of its own address. I think that can be an input to the function

src/services/rfqm_service.ts
phil-ociraptor: yeah, that's much better

src/services/rfqm_service.ts
phil-ociraptor: haha nice, you caught the unfinished portion of this diff - I think you're right, passing in the address as an input to this function is the best way to handle this (as opposed to initializing the rfqmService with an optional workerAddress)

src/services/rfqm_service.ts
```diff
@@ -456,4 +509,39 @@ export class RfqmService {
             deduplicationId: orderHash,
         });
     }
+
+    private async _validateJobAsync(orderHash: string, job: RfqmJobEntity): Promise<void> {
```
alexkroeger, line 513: nice

alexkroeger (approved):

src/services/rfqm_service.ts
```diff
@@ -439,11 +443,60 @@ export class RfqmService {
...
+
+        // Start processing
+        await this.dbUtils.updateRfqmJobAsync(orderHash, {
+            status: RfqmJobStatus.Processing,
+        });
+
+        // Validate w/Eth Call
+        try {
+            await this._blockchainUtils.decodeMetaTransactionCallDataAndValidateAsync(calldata!, workerAddress);
+        } catch (e) {
```
alexkroeger, line 467: we should log the message

src/services/rfqm_service.ts
```diff
@@ -439,11 +443,60 @@ export class RfqmService {
...
+        }
+
+        // Get last look from MM
+        const submitRequest: SubmitRequest = {
+            order: storedOrderToRfqmOrder(order!),
+            orderHash,
+            fee: storedFeeToFee(fee!),
+        };
+
+        const shouldProceed = await this._quoteServerClient.confirmLastLookAsync(makerUri!, submitRequest);
```
alexkroeger, line 483: more logging here would be good too, we'll want to keep track of the ratio of confirmed quotes to rejected

src/services/rfqm_service.ts
```diff
@@ -456,4 +509,39 @@ export class RfqmService {
...
+                status: RfqmJobStatus.Failed,
+                statusReason: 'Missing makerUri',
+            });
+            return;
+        }
+
+        if (order === null) {
+            await this.dbUtils.updateRfqmJobAsync(orderHash, {
+                status: RfqmJobStatus.Failed,
+                statusReason: 'Missing order on fee',
```
rhinodavid, line 534: "order on fee"?

src/services/rfqm_service.ts
phil-ociraptor: good call - definitely

src/services/rfqm_service.ts
phil-ociraptor: very true! Will be able to get this from the DB as well, but being able to alert on this could be very useful too

src/services/rfqm_service.ts
phil-ociraptor: haha, yes, this is not right

PR #616, opened 2021-06-03
feat: RFQM transaction submission

alexkroeger:
# Description

- creates `rfqm_transaction_submissions` table with corresponding typeorm entity
- creates `completeSubmissionLifecycleAsync` function with initial submission, monitoring, and re-submission logic

# Testing Instructions

- [x] test migration and migration revert locally
- [x] unit tests for `completeSubmissionLifecycleAsync`
- [x] database tests for `rfqmTransactionSubmissionEntity`

# Checklist

    -   [ ] SRA/Limit orders
    -   [ ] Swap endpoints
    -   [ ] Depth charts


migrations/1622670193018-CreateRfqmTransactionSubmissionsTable.ts
```diff
@@ -0,0 +1,54 @@
+import { MigrationInterface, QueryRunner, Table, TableIndex, TableForeignKey } from 'typeorm';
+
+const newTableName = 'rfqm_transaction_submissions';
+const transactionSubmissionsTable = new Table({
+    name: newTableName,
+    columns: [
+        { name: 'transaction_hash', type: 'varchar', isPrimary: true },
```
PirosB3, line 7: @rhinodavid this may impact your work on the `status` update?

src/services/rfqm_service.ts
```diff
@@ -439,6 +450,118 @@ export class RfqmService {
...
+     */
+    private async _submitTransactionAsync(
+        orderHash: string,
+        workerAddress: string,
+        callData: string,
+        gasPrice: BigNumber,
+    ): Promise<RfqmTransactionSubmissionEntity> {
+        // get transaction properties
+        const [nonce, gasEstimate] = await Promise.all([
+            this._blockchainUtils.getNonceAsync(workerAddress),
```
PirosB3, line 532: Shouldn't we be fetching the nonce only once? `_submitTransactionAsync` is called by `completeSubmissionLifecycleAsync`. `_submitTransactionAsync`. If for some reason a transaction ends up being mined during `completeSubmissionLifecycleAsync`, after the [receipt check](https://github.com/0xProject/0x-api/pull/616/files#diff-289e53810d2a508fae0fff19710e2e2b9846a98a8935466d7e20a1490f179376R474), wouldn't we be re-submitting a new trade with a new incremented nonce?

src/services/rfqm_service.ts
```diff
@@ -439,6 +450,118 @@ export class RfqmService {
...
+
+        let gasPrice = await this._protocolFeeUtils.getGasPriceEstimationOrThrowAsync();
+
+        const firstSubmission = await this._submitTransactionAsync(orderHash, workerAddress, callData, gasPrice);
+        submissions[firstSubmission.transactionHash!] = firstSubmission;
+
+        let isFinishedMonitoring = false;
+        let shouldTryResubmitting = true;
+        while (!isFinishedMonitoring) {
+            await sleepAsync(TRANSACTION_WATCHER_SLEEP_TIME_MS);
```
PirosB3, line 471: I think we have a library called `delay`.

`await delay(TRANSACTION_WATCHER_SLEEP_TIME_MS`

src/services/rfqm_service.ts
```diff
@@ -439,6 +450,118 @@ export class RfqmService {
...
+                    });
+                    await this.dbUtils.updateRfqmTransactionSubmissionsAsync(Object.values(submissions));
+                    break;
+                }
+            }
+
+            if (shouldTryResubmitting) {
+                const newGasPrice = await this._protocolFeeUtils.getGasPriceEstimationOrThrowAsync();
+
+                if (gasPrice.lt(newGasPrice)) {
```
PirosB3, line 512: are we going to be constantly resubmitting? (I assume this statement will always be `true`)

PirosB3: Should be otherwise only resubmit if gas price changes by a certain threshold?

src/utils/rfq_blockchain_utils.ts
```diff
@@ -114,6 +115,35 @@ export class RfqBlockchainUtils {
...
+    }
+
+    public async estimateGasForExchangeProxyCallAsync(callData: string, workerAddress: string): Promise<number> {
+        const txData: Partial<TxData> = {
+            to: this._exchangeProxy.address,
+            data: callData,
+            from: workerAddress,
+        };
+
+        const gasEstimate = await this._web3Wrapper.estimateGasAsync(txData);
```
PirosB3, line 141: Why do we estimate every single time? can we hard-code an estimate instead? we should know upfront what the upper and lower bounds for this trade is, right? can we hard-code and avoid an RPC call?

src/utils/rfq_blockchain_utils.ts
```diff
@@ -114,6 +115,35 @@ export class RfqBlockchainUtils {
         return this._exchangeProxy.executeMetaTransaction(metaTx, metaTxSig).getABIEncodedTransactionData();
     }
 
+    public async getNonceAsync(workerAddress: string): Promise<number> {
```
PirosB3, line 118: These 4 functions below look like proxies, is there a need for this layer of abstraction?

src/utils/rfqm_db_utils.ts
```diff
@@ -22,6 +22,21 @@ export interface RfqmJobOpts {
     metadata?: object;
 }
 
+export interface RfqmTransactionSubmissionOpts {
```
PirosB3, line 25: why not extend `TxOpts`?

PirosB3 (changes requested):
Did another review! amazing stuff.

I'd like to chat more about the transaction receipt flow since I'm not sure that I fully understand it. Let me know when you have a sec and we can discuss :)

src/services/rfqm_service.ts
```diff
@@ -121,6 +126,12 @@ const RFQM_SIGNED_QUOTE_EXPIRY_TOO_SOON = new Counter({
     help: 'A signed quote was not queued because it would expire too soon',
 });
 
+async function sleepAsync(ms: number): Promise<void> {
```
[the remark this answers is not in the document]
rhinodavid: If it's this simple my inclination is to avoid a dependency

src/services/rfqm_service.ts
```diff
@@ -439,6 +450,118 @@ export class RfqmService {
...
+            for (const receipt of receipts) {
+                // if a tx has been mined, stop trying to resubmit
+                if (receipt.response !== undefined) {
+                    shouldTryResubmitting = false;
+                    const currentBlock = await this._blockchainUtils.getCurrentBlockAsync();
+                    if (receipt.response.blockNumber - currentBlock >= BLOCK_FINALITY_THRESHOLD) {
+                        isFinishedMonitoring = true;
+                    }
+                    // update all entities
+                    receipts.map((r) => {
```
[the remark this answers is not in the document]
rhinodavid: `forEach` here if not consuming the result of `map`

src/services/rfqm_service.ts
```diff
@@ -439,6 +450,118 @@ export class RfqmService {
         return Math.random() > 0.5 ? Promise.reject('failing') : Promise.resolve();
     }
 
+    /**
+     * Submit a transaction, re-submit if necessary, and poll for resolution
+     */
+    public async completeSubmissionLifecycleAsync(
```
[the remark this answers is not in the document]
phil-ociraptor: correct, that is done outside of this function, right before this function gets called

src/services/rfqm_service.ts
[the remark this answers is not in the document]
phil-ociraptor: alas, `delay` is already a dep

src/services/rfqm_service.ts
```diff
@@ -439,6 +450,118 @@ export class RfqmService {
...
+    ): Promise<void> {
+        const submissions: { [transactionHash: string]: RfqmTransactionSubmissionEntity } = {};
+
+        let gasPrice = await this._protocolFeeUtils.getGasPriceEstimationOrThrowAsync();
+
+        const firstSubmission = await this._submitTransactionAsync(orderHash, workerAddress, callData, gasPrice);
+        submissions[firstSubmission.transactionHash!] = firstSubmission;
+
+        let isFinishedMonitoring = false;
+        let shouldTryResubmitting = true;
```
phil-ociraptor, line 469: hmm... having two booleans strikes me as a little complex. Is it possible to get away with just one? I'll noodle on this a bit too

src/services/rfqm_service.ts
```diff
@@ -439,6 +450,118 @@ export class RfqmService {
...
 
+    /**
+     * Submit a transaction, re-submit if necessary, and poll for resolution
+     */
+    public async completeSubmissionLifecycleAsync(
+        orderHash: string,
+        workerAddress: string,
+        callData: string,
+    ): Promise<void> {
+        const submissions: { [transactionHash: string]: RfqmTransactionSubmissionEntity } = {};
```
phil-ociraptor, line 461: nit: can we name this `submissionMap` - I keep on thinking of `submissions` as a list

src/services/rfqm_service.ts
phil-ociraptor: ok, after thinking about it a little, I think I see that this is loop is a two phased process:
- submitting (and resubmitting if the gas prices moves against our highest gas price so far)
- waiting for finality


I wonder if we could simplify the code by breaking these into two distinct processes. Not sure if possible, but striving for code that at a high level looks like:
 ```
await mineTransaction(...);
await achieveTransactionFinality(...);
```

File: src/services/rfqm_service.ts (hunk 4 above)
Comment (phil-ociraptor) [reply src/services/rfqm_service.ts:512]: Hmmm.... in theory, there are cases when we will not resubmit (i.e. gas price stays constant). 

In practice, I think @PirosB3 is correct - we will pretty much only see this if the txn wasn't mined because gas prices have increased. It's possible we send txns a little too aggressively if we don't have a threshold

File: src/services/rfqm_service.ts (hunk 3 above)
Comment (alexkroeger) [reply src/services/rfqm_service.ts:471]: 👍

File: src/services/rfqm_service.ts [hunk 13]
```diff
@@ -439,6 +450,118 @@ export class RfqmService {
...
+                    };
+                }),
+            );
+
+            for (const receipt of receipts) {
+                // if a tx has been mined, stop trying to resubmit
+                if (receipt.response !== undefined) {
+                    shouldTryResubmitting = false;
+                    const currentBlock = await this._blockchainUtils.getCurrentBlockAsync();
+                    if (receipt.response.blockNumber - currentBlock >= BLOCK_FINALITY_THRESHOLD) {
```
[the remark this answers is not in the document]
alexkroeger: yes 👍

src/services/rfqm_service.ts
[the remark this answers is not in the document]
alexkroeger: yea map not necessary 👍

src/services/rfqm_service.ts
alexkroeger: I think we should think of submission as free tho (within reason). This is pretty aggressive resubmission but might as well be aggressive to get a better chance at mining

src/services/rfqm_service.ts
alexkroeger: yes, changed this

src/utils/rfq_blockchain_utils.ts
```diff
@@ -114,6 +115,35 @@ export class RfqBlockchainUtils {
...
+        const txData: Partial<TxData> = {
+            to: this._exchangeProxy.address,
+            data: callData,
+            from: workerAddress,
+        };
+
+        const gasEstimate = await this._web3Wrapper.estimateGasAsync(txData);
+
+        // add a buffer
+        return Math.ceil((GAS_ESTIMATE_BUFFER + 1) * gasEstimate);
```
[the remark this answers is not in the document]
alexkroeger: we could but the number will be small (no more than 250K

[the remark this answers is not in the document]
alexkroeger: I'd also have to re-convert it to a number for consumption by web3Wrapper

src/utils/rfq_blockchain_utils.ts
alexkroeger: alternatively we could expose web3wrapper publicly

src/utils/rfq_blockchain_utils.ts
alexkroeger: it can change based on on-chain conditions, but that's a bit paranoid--we can just rely on the buffer

src/services/rfqm_service.ts
```diff
@@ -439,6 +450,118 @@ export class RfqmService {
...
+     * Submit a transaction, re-submit if necessary, and poll for resolution
+     */
+    public async completeSubmissionLifecycleAsync(
+        orderHash: string,
+        workerAddress: string,
+        callData: string,
+    ): Promise<void> {
+        const submissions: { [transactionHash: string]: RfqmTransactionSubmissionEntity } = {};
+
+        let gasPrice = await this._protocolFeeUtils.getGasPriceEstimationOrThrowAsync();
```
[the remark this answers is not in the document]
alexkroeger: increases submission costs by 1-2% 🤷‍♂️ 

I think we should try without at first

src/services/rfqm_service.ts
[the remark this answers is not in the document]
PirosB3: that's correct, I agree with David but `delay` is already part of our source code

src/services/rfqm_service.ts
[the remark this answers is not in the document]
PirosB3: Sounds good, I was just thinking that many bots out there simply submit trades at `fastest` gas price, and we could easily "front-run" that wall by adding 1 gwei more or so?

src/services/rfqm_service.ts
```diff
@@ -439,6 +450,118 @@ export class RfqmService {
...
+    /**
+     * Determine transaction properties and submit a transaction
+     */
+    private async _submitTransactionAsync(
+        orderHash: string,
+        workerAddress: string,
+        callData: string,
+        gasPrice: BigNumber,
+    ): Promise<RfqmTransactionSubmissionEntity> {
+        // get transaction properties
```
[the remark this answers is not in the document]
phil-ociraptor: This will be done right before the `completeSubmissionLifecycleAsync` will get called, so no need to do it twice

src/utils/rfq_blockchain_utils.ts
PirosB3: this could be a little cleaner, even though less 00? up to you to decide :) - this current impl works fine too!

src/utils/rfq_blockchain_utils.ts
PirosB3: If we hard-coded something like 180k, we should be safe right? that's like a 10-15% buffer.

src/utils/rfqm_db_utils.ts
alexkroeger: this is for the db, not the web3 call

rhinodavid: Maybe put `Storage` or something like that somewhere in the name of the storage types so that's clear

src/utils/rfq_blockchain_utils.ts
alexkroeger: probably not, some tokens can cost much higher

src/services/rfqm_service.ts
```diff
@@ -439,6 +450,118 @@ export class RfqmService {
...
+            callData,
+            workerAddress,
+            txOptions,
+        );
+
+        // save tx submission to DB
+        const entityOpts: RfqmTransactionSubmissionOpts = {
+            transactionHash,
+            orderHash,
+            createdAt: new Date(),
```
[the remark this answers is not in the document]
alexkroeger: I'll default to just using now() from DB

src/utils/rfqm_db_utils.ts
```diff
@@ -22,6 +22,24 @@ export interface RfqmJobOpts {
     metadata?: object;
 }
 
+export interface RfqmTransactionSubmissionEntityOpts {
```
phil-ociraptor, line 25: see comment above about possibly using `Partial< RfqmTransactionSubmissionEntity>`

test/rfq_blockchain_utils_test.ts
```diff
@@ -130,7 +142,7 @@ describe(SUITE_NAME, () => {
         await teardownDependenciesAsync(SUITE_NAME);
     });
 
-    describe('validateMetaTransaction', () => {
+    describe.only('validateMetaTransaction', () => {
```
phil-ociraptor, line 145: nit: remove .only

src/services/rfqm_service.ts
alexkroeger: I chunked up the logic a bit more to make this a bit more clear, tho not quite as suggested

phil-ociraptor (commented):
Amazing! Big props on all of the nitty gritty details here:
- Decoding logs
- Parsing call data
- Handling partial fills
- Testing!

Left some minor comments, but high level looks great

src/services/rfqm_service.ts
```diff
@@ -439,6 +455,189 @@ export class RfqmService {
...
+            submissionsMap = statusCheckResult.submissionsMap;
+
+            if (!isTxMined) {
+                const newGasPrice = await this._protocolFeeUtils.getGasPriceEstimationOrThrowAsync();
+                try {
+                    await this._blockchainUtils.decodeMetaTransactionCallDataAndValidateAsync(callData, workerAddress, {
+                        gasPrice: newGasPrice,
+                    });
+                } catch (err) {
+                    this._protocolFeeUtils.getGasPriceEstimationOrThrowAsync();
```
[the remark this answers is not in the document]
alexkroeger: it's nonsense, I was too tired when I wrote it. It was the start of some logic for giving up on expired tx, but I'm saving that for a follow up!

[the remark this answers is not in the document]
alexkroeger: thanks for catching!

src/services/rfqm_service.ts
```diff
@@ -439,6 +455,189 @@ export class RfqmService {
...
+            Object.keys(submissionsMap).map(async (transactionHash) => {
+                return {
+                    transactionHash,
+                    response: await this._blockchainUtils.getTransactionReceiptIfExistsAsync(transactionHash),
+                };
+            }),
+        );
+
+        for (const receipt of receipts) {
+            if (receipt.response !== undefined) {
```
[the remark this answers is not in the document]
alexkroeger: yes, only one tx should be mined. in theory if we get more than one we'll still record the results in the subloop, but it shouldn't happen

src/services/rfqm_service.ts
```diff
@@ -513,6 +528,184 @@ export class RfqmService {
...
+            orderHash,
+            workerAddress,
+            callData,
+            gasPrice,
+            nonce,
+            gasEstimate,
+            expectedTakerTokenFillAmount,
+        );
+        submissionsMap[firstSubmission.transactionHash!] = firstSubmission;
+
```
PirosB3, line 563: Offering a small suggestion for refactor, unnecessary to do if you don't feel like it.

```
        let latestStatusCheck: SubmissionsMapStatus = {
            isTxFinalized: false,
            isTxMined: false,
            submissionsMap: {
                [firstSubmission.transactionHash!]: firstSubmission,
            }
        };
        while (!latestStatusCheck.isTxFinalized) {
            await delay(TRANSACTION_WATCHER_SLEEP_TIME_MS);

            latestStatusCheck = await this._checkSubmissionMapReceiptsAndUpdateDbAsync(
                latestStatusCheck.submissionsMap,
                expectedTakerTokenFillAmount,
            );
            if (!latestStatusCheck.isTxMined) {
                const newGasPrice = await this._protocolFeeUtils.getGasPriceEstimationOrThrowAsync();

                if (gasPrice.lt(newGasPrice)) {
                    gasPrice = newGasPrice;
                    const submission = await this._submitTransactionAsync(
                        orderHash,
                        workerAddress,
                        callData,
                        gasPrice,
                        nonce,
                        gasEstimate,
                        expectedTakerTokenFillAmount,
                    );
                    latestStatusCheck.submissionsMap[submission.transactionHash!] = submission;
                }
            }
        }
    }
```

PirosB3: This allows you to avoid the
```
isTxFinalized = false
isTxMined = false
```

src/services/rfqm_service.ts
```diff
@@ -513,6 +528,184 @@ export class RfqmService {
...
+                }
+                // update all entities
+                // since the same nonce is being re-used, we expect only 1 defined receipt
+                for (const r of receipts) {
+                    if (r.response !== undefined) {
+                        if (r.response.status === 1) {
+                            const actualTakerTokenFilledAmount = this._blockchainUtils.getRfqOrderTakerTokenFilledAmountFromLogs(
+                                r.response.logs,
+                            );
+                            submissionsMap[r.transactionHash].status = actualTakerTokenFilledAmount.lt(
```
PirosB3, line 633: small suggestion, in JS all objects are references unless explicitly copied, so you can avoid the redundant `submissionsMap[r.transactionHash]` with an interin variable assignment.

```
const successfulTransaction = submissionsMap[r.transactionHash];
successfulTransaction.status = ...
successfulTransaction.actualTakerTokenFillAmount = ....
```

PirosB3 (approved):
A few minor comments!

src/utils/rfq_blockchain_utils.ts
```diff
@@ -129,6 +156,61 @@ export class RfqBlockchainUtils {
...
+    public getDecodedRfqOrderFillEventLogFromLogs(
+        logs: LogEntry[],
+    ): LogWithDecodedArgs<IZeroExRfqOrderFilledEventArgs> {
+        for (const log of logs) {
+            if (log.topics[0] === RFQ_ORDER_FILLED_EVENT_TOPIC0) {
+                return this._abiDecoder.tryToDecodeLogOrNoop(log) as LogWithDecodedArgs<IZeroExRfqOrderFilledEventArgs>;
+            }
+        }
+        throw new Error(
+            `no RfqOrderFilledEvent logs among the logs passed into getRfqOrderTakerTokenFilledAmountFromLogs`,
```
phil-ociraptor, line 210: nit: error message references `getRfqOrderTakerTokenFilledAmountFromLogs` but method is `getDecodedRfqOrderFillEventLogFromLogs`

src/utils/rfq_blockchain_utils.ts
```diff
@@ -129,6 +156,61 @@ export class RfqBlockchainUtils {
...
+            from: workerAddress,
+        };
+
+        const gasEstimate = await this._web3Wrapper.estimateGasAsync(txData);
+
+        // add a buffer
+        return Math.ceil((GAS_ESTIMATE_BUFFER + 1) * gasEstimate);
+    }
+
+    public getRfqOrderTakerTokenFilledAmountFromLogs(logs: LogEntry[]): BigNumber {
```
phil-ociraptor, line 188: Can this method be written in terms of `getDecodedRfqOrderFillEventLogFromLogs`?

src/services/rfqm_service.ts
```diff
@@ -513,6 +528,184 @@ export class RfqmService {
...
+    /**
+     * Check for receipts from the tx's in a SubmissionsMap object
+     * Update database with status of all tx's
+     */
+    private async _checkSubmissionMapReceiptsAndUpdateDbAsync(
+        submissionsMap: SubmissionsMap,
+        expectedTakerTokenFillAmount: BigNumber,
+    ): Promise<SubmissionsMapStatus> {
+        let isTxMined: boolean = false;
+        let isTxFinalized: boolean = false;
```
phil-ociraptor, line 606: super nit: do not need the explicit `boolean` type declaration as type inference will determine that

src/utils/rfq_blockchain_utils.ts
```diff
@@ -16,10 +17,30 @@ import { SubproviderAdapter } from './subprovider_adapter';
 const MIN_GAS_PRICE = new BigNumber(0);
 // 10K Gwei
 const MAX_GAS_PRICE = new BigNumber(1e13);
+const GAS_ESTIMATE_BUFFER = 0.05;
+const RFQ_ORDER_FILLED_EVENT_TOPIC0 = '0x829fa99d94dc4636925b38632e625736a614c154d55006b7ab6bea979c210c32';
```
phil-ociraptor, line 21: is this chain agnostic?

phil-ociraptor (approved):
Things look good on my side, just some small nits!

src/services/rfqm_service.ts
```diff
@@ -92,6 +97,16 @@ export interface MetaTransactionRfqmQuoteResponse extends BaseRfqmQuoteResponse
     orderHash: string;
 }
 
+export interface SubmissionsMapStatus {
+    isTxMined: boolean;
+    isTxFinalized: boolean;
+    submissionsMap: SubmissionsMap;
+}
+
+export interface SubmissionsMap {
```
rhinodavid, line 106: Two code style points to consider:
I would avoid "Map" in the name; a reader can see the type. `TransactionHashtoSubmissionRecord` or something would be more descriptive.
Maybe use a [Map](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Map)

src/utils/rfq_blockchain_utils.ts
alexkroeger: you right

src/utils/rfq_blockchain_utils.ts
alexkroeger: yea 👍

src/utils/rfq_blockchain_utils.ts
alexkroeger: it definitely can be, I'm actually just going to get rid of it. It's not necessary now

PR #622, opened 2021-06-09
fix: Submit tx upgrade

PirosB3:
As discussed, we added [takerTokenFillAmount](https://github.com/0xProject/quote-server/pull/21) in the Submit request and response. I've updated the QuoteServerClient to reflect these changes.

### DIFF
--- a/src/utils/quote_server_client.ts
+++ b/src/utils/quote_server_client.ts
@@ -28,7 +28,11 @@ export class QuoteServerClient {
             const response = await this._axiosInstance.post(`${makerUri}/submit`, payload, {
                 timeout: ONE_SECOND_MS * 2,
             });
-            schemaValidator.validate(response.data, schemas.submitReceiptSchema);
+            const validator = schemaValidator.validate(response.data, schemas.submitReceiptSchema);
+            if (validator.errors.length > 0) {
+                const errorsMsg = validator.errors.map(err => err.toString()).join(',');
+                throw new Error(`Error from validator: ${errorsMsg}`);
+            }
             const responseFee: Fee = {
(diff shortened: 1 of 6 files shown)


src/utils/quote_server_client.ts
```diff
@@ -28,7 +28,11 @@ export class QuoteServerClient {
             const response = await this._axiosInstance.post(`${makerUri}/submit`, payload, {
                 timeout: ONE_SECOND_MS * 2,
             });
-            schemaValidator.validate(response.data, schemas.submitReceiptSchema);
+            const validator = schemaValidator.validate(response.data, schemas.submitReceiptSchema);
+            if (validator.errors.length > 0) {
```
PirosB3, line 32: @phil-ociraptor `validate` does not throw an error by default, so I just force it to throw an error if errors are present.

src/utils/quote_server_client.ts
```diff
@@ -39,6 +43,18 @@ export class QuoteServerClient {
                 throw new Error('Fee in response is not equal to fee in request');
             }
 
+            if (response.data.signedOrderHash !== payload.orderHash) {
```
PirosB3, line 46: @phil-ociraptor I also added this validation, which makes sure that the right order hash is returned.

alexkroeger (approved):

src/utils/quote_server_client.ts
phil-ociraptor: hmmm.... I wonder how some of my tests were passing, lol, but all good!

phil-ociraptor (approved):
Looks good! Tests seem to be failing for a seemingly unrelated test - so I kicked off another build. Once they pass, should be good to go

PR #632, opened 2021-06-10
fix: Align RFQM entity types with their db representations

rhinodavid:
# Description

Maybe??

- Beat at the TypeScript type checker until we have removed the `?` types in the RFQM entities where we known the column is non-nullable.
- Move types the entity depends on to the entity definition file. This removes the import cycle between entities and `rfqmDbUtils` and creates a clearer dependency path.
- Ensure that we are relying on the database's clock for default `createdAt` times

No underlying changes to the database happen due to this PR; everything here is at the TS type level.

# Testing Instructions

- Ensure db comes up and migrates
- Ensure db test passes, paying attention to new time check

# Checklist

    -   [ ] SRA/Limit orders
    -   [ ] Swap endpoints
    -   [ ] Depth charts

### DIFF
--- a/src/entities/RfqmJobEntity.ts
+++ b/src/entities/RfqmJobEntity.ts
@@ -1,5 +1,4 @@
 import { BigNumber } from '@0x/asset-swapper';
-import { RfqOrder } from '@0x/protocol-utils';
 import { Column, Entity, Index, PrimaryColumn } from 'typeorm';
 
 import { BigNumberTransformer } from './transformers';
--- a/src/entities/RfqmJobEntity.ts
+++ b/src/entities/RfqmJobEntity.ts
@@ -33,7 +33,6 @@ export interface V4RfqStoredOrder {
 }
 
 export type StoredOrder = V4RfqStoredOrder;
-export type RfqmOrder = RfqOrder;
 
 export interface V4StringRfqOrderFields {
     txOrigin: string;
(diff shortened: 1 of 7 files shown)


test/rfqm_db_test.ts
```diff
@@ -91,6 +83,29 @@ describe(SUITE_NAME, () => {
     });
 
     describe('rfqm db tests', () => {
+        it("should use the database's timestamp", async () => {
```
alexkroeger, line 86: nice!

src/entities/RfqmJobEntity.ts
```diff
@@ -1,46 +1,94 @@
...
+    order: V4StringRfqOrderFields;
+}
+
+export interface V4RfqStoredOrder {
+    type: RfqmOrderTypes.V4Rfq;
+    order: V4StringRfqOrderFields;
+}
+
+export type StoredOrder = V4RfqStoredOrder;
+export type RfqmOrder = RfqOrder;
```
alexkroeger, line 36: nit, I would move this type back to db_utils since it's related to parsing the stored order into something more useful by the application

alexkroeger (approved):
lgtm! thanks!

test/rfqm_db_test.ts
rhinodavid: 🥰

src/entities/RfqmJobEntity.ts
rhinodavid: dun

PR #635, opened 2021-06-10
feat: Add status logic to rfqm service

rhinodavid:
# Description

Add logic to the rfqm service to handle the future /status endpoint.

TODO: write handler

# Testing Instructions

Unit test suite included.

# Checklist

    -   [ ] SRA/Limit orders
    -   [ ] Swap endpoints
    -   [ ] Depth charts

### DIFF
--- a/src/services/rfqm_service.ts
+++ b/src/services/rfqm_service.ts
@@ -76,6 +76,25 @@ export interface BaseRfqmQuoteResponse {
     sellTokenAddress: string;
 }
 
+interface OrderStatusPending {
+    status: 'pending';
+}
+
+interface OrderStatusSubmitted {
+    status: 'submitted';
+    transactions: { hash: string; timestamp: number /* unix ms */ }[];
+}
+
+interface OrderStatusFailed {
+    status: 'failed';
+    transactions: { hash: string; timestamp: number /* unix ms */ }[];
+}
+
+interface OrderStatusSuccessful {
+    status: 'successful';
+    transaction: { hash: string; timestamp: number /* unix ms */ }; // the filled transaction
+}
+
 export interface MetaTransactionSubmitRfqmSignedQuoteParams {
     type: RfqmTypes.MetaTransaction;
     metaTransaction: MetaTransaction;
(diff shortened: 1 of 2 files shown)


src/services/rfqm_service.ts
```diff
@@ -471,6 +491,70 @@ export class RfqmService {
         return this._blockchainUtils.isWorkerReadyAsync(workerAddress, gasPrice);
     }
 
+    public async getOrderStatusAsync(orderHash: string): Promise<StatusResponse | null> {
+        const transformSubmission = (submission: RfqmTransactionSubmissionEntity) => {
+            const { transactionHash: hash, createdAt } = submission;
```
alexkroeger, line 496: we stored which tx got mined, might be helpful to add an indication to the transaction as to whether we know it's mined

alexkroeger: ah I see for successful jobs we only return 1--the one that succeeded

alexkroeger (approved):
lgtm! consider wrapping both db calls in promise.all()

src/services/rfqm_service.ts
```diff
@@ -471,6 +491,70 @@ export class RfqmService {
...
+            if (!hash) {
```
PirosB3, line 497: `return hash ? { hash, timestamp: createdAt.getTime()} : null;`

^ optional suggestion

src/services/rfqm_service.ts
```diff
@@ -76,6 +76,25 @@ export interface BaseRfqmQuoteResponse {
     sellTokenAddress: string;
 }
 
+interface OrderStatusPending {
+    status: 'pending';
+}
+
+interface OrderStatusSubmitted {
+    status: 'submitted';
+    transactions: { hash: string; timestamp: number /* unix ms */ }[];
```
alexkroeger, line 85: I think this is okay (I get the logic) but mildly worried about a different properties appearing on the response schema (transaction vs. transactions)

rhinodavid: @PirosB3 made it sound like this is an internal only type thing. If that's the case any TS users can just import the type, correct??

src/services/rfqm_service.ts
```diff
@@ -471,6 +491,70 @@ export class RfqmService {
...
+            }
+            return {
+                hash,
+                timestamp: createdAt.getTime(),
+            };
+        };
+        const transformSubmissions = (submissions: RfqmTransactionSubmissionEntity[]) =>
+            submissions.map(transformSubmission).flatMap((s) => (s ? s : []));
+
+        const job = await this._dbUtils.findJobByOrderHashAsync(orderHash);
```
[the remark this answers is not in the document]
rhinodavid: big IF we batched them over the connection maybe this would be worth it for sure

src/services/rfqm_service.ts
rhinodavid: yah

src/services/rfqm_service.ts
rhinodavid: LOVE the swagger

src/services/rfqm_service.ts
rhinodavid: response type has been simplified to just `transactions`

PR #638, opened 2021-06-11
feat: add SqsRetryableError type

phil-ociraptor:
# Description

Previously, all errors from `handleMessage` were retried, leading to some strange behavior. In this change, only errors are that explicitly marked safe for Retry (of type `SqsRetryableError`) will be retried. All others are not retried.

# Testing Instructions

# Checklist

    -   [ ] SRA/Limit orders
    -   [ ] Swap endpoints
    -   [ ] Depth charts


src/utils/sqs_consumer.ts
```diff
@@ -95,10 +97,13 @@ export class SqsConsumer {
...
                 'Encountered error while handling message',
             );
-            // Retry message
-            await this._sqsClient.changeMessageVisibilityAsync(message.ReceiptHandle!, 0);
-            await delay(ONE_SECOND_MS);
-            return;
+
+            if (err instanceof SqsRetryableError) {
+                // Retry message
+                await this._sqsClient.changeMessageVisibilityAsync(message.ReceiptHandle!, 0);
```
PirosB3, line 103: Could we add some logging just to be verbose?

PirosB3 (commented):
Are we going to be throwing `SqsRetryableError` in a separate PR?

src/utils/sqs_consumer.ts
alexkroeger: in this case, this pushes it back to the queue for any worker to grab, right?

src/utils/sqs_consumer.ts
```diff
@@ -1,3 +1,4 @@
+// tslint:disable: max-classes-per-file
```
rhinodavid, line 1: Team,

Start keeping a list (in your head or somewhere) of arbitrary rules like this that aren't serving us well. I've been chatting with @xianny about a migration to eslint and that will be a good time to get rid of things like this we just end up turning off.

-d

src/utils/sqs_consumer.ts
phil-ociraptor: re logging: Yup - can do

re queue behavior: Yup - the message becomes immediately available for any worker to attempt next (i.e. doesn't go to the end of the queue, but stays in the front)

src/utils/sqs_consumer.ts
phil-ociraptor: totally - can also do a search across the codebase for `tsling:disable` to see what types are prevalent. But I think the max classes per file one is a rather annoying restriction

alexkroeger (approved):

src/utils/sqs_consumer.ts
```diff
@@ -87,18 +78,15 @@ export class SqsConsumer {
...
-                },
-                'Encountered error while handling message',
-            );
-            // Retry message
-            await this._sqsClient.changeMessageVisibilityAsync(message.ReceiptHandle!, 0);
-            await delay(ONE_SECOND_MS);
-            return;
+            logger.error({ err, message, id: this._id }, 'Encountered error while handling message');
+
+            if (err instanceof SqsRetryableError) {
```
PirosB3, line 83: Just double checking since I don't know JS. Can you confirm that `instanceOf` is good enough for our use case? I'm not too sure how classes work in JS (my understanding is that they modify the prototype chain, but that may be wrong)

PirosB3 (approved):

src/utils/sqs_consumer.ts
phil-ociraptor: Yeah, that's absolutely correct - the prototype chain of this `err` object may look like:
- `err.__proto__.name === 'SqsRetryableError'`
-`err.__proto__.__proto__.name === 'Error'`

I believe `instanceof` will check the first one, and if it doesn't match, will continue up the `__proto__` chain (i.e. `__proto__.__proto__...`, until the root is hit (i.e. there is no more `__proto__` field), in which case it returns `false`. Of course, returns `true` if a match is found before that.

PR #642, opened 2021-06-14
feat: add Confirmed status to RfqmTransactionSubmission

phil-ociraptor:
# Description

# Testing Instructions

Will add tests.

# Checklist

    -   [ ] SRA/Limit orders
    -   [ ] Swap endpoints
    -   [ ] Depth charts

### DIFF
--- a/src/entities/RfqmTransactionSubmissionEntity.ts
+++ b/src/entities/RfqmTransactionSubmissionEntity.ts
@@ -3,7 +3,7 @@ import { Column, Entity, Index, PrimaryColumn } from 'typeorm';
 
 import { BigIntTransformer, BigNumberTransformer } from './transformers';
 
-export enum RfqmTranasctionSubmissionStatus {
+export enum RfqmTransactionSubmissionStatus {
     Submitted = 'submitted',
     Successful = 'successful',
     Reverted = 'reverted',
@@ -49,7 +49,7 @@ export class RfqmTransactionSubmissionEntity {
 
     @Index()
     @Column({ name: 'status', type: 'varchar' })
-    public status: RfqmTranasctionSubmissionStatus;
+    public status: RfqmTransactionSubmissionStatus;
 
     @Column({ name: 'status_reason', type: 'varchar', nullable: true })
     public statusReason: string | null;
@@ -70,7 +70,7 @@ export class RfqmTransactionSubmissionEntity {
         this.metadata = opts.metadata || null;
         this.nonce = opts.nonce || null;
         this.orderHash = opts.orderHash;
-        this.status = opts.status || RfqmTranasctionSubmissionStatus.Submitted;
+        this.status = opts.status || RfqmTransactionSubmissionStatus.Submitted;
         this.statusReason = opts.statusReason || null;
         this.to = opts.to || null;
         this.transactionHash = opts.transactionHash;
--- a/src/entities/RfqmTransactionSubmissionEntity.ts
+++ b/src/entities/RfqmTransactionSubmissionEntity.ts
@@ -8,6 +8,7 @@ export enum RfqmTranasctionSubmissionStatus {
     Successful = 'successful',
     Reverted = 'reverted',
     DroppedAndReplaced = 'droppedAndReplaced',
+    Confirmed = 'confirmed',
 }
 
 @Entity({ name: 'rfqm_transaction_submissions' })
--- a/src/entities/RfqmTransactionSubmissionEntity.ts
+++ b/src/entities/RfqmTransactionSubmissionEntity.ts
@@ -4,12 +4,12 @@ import { Column, Entity, Index, PrimaryColumn } from 'typeorm';
 import { BigIntTransformer, BigNumberTransformer } from './transformers';
 
 export enum RfqmTransactionSubmissionStatus {
+    DroppedAndReplaced = 'dropped_and_replaced',
+    RevertedConfirmed = 'reverted_confirmed',
+    RevertedUnconfirmed = 'reverted_unconfirmed',
     Submitted = 'submitted',
-    Successful = 'successful',
-    Reverted = 'reverted',
-    DroppedAndReplaced = 'droppedAndReplaced',
-    ConfirmedSuccessful = 'confirmedSuccessful',
-    ConfirmedReverted = 'confirmedReverted',
+    SucceededConfirmed = 'succeeded_confirmed',
+    SucceededUnconfirmed = 'succeeded_unconfirmed',
 }
 
 @Entity({ name: 'rfqm_transaction_submissions' })
--- a/src/entities/RfqmTransactionSubmissionEntity.ts
+++ b/src/entities/RfqmTransactionSubmissionEntity.ts
@@ -8,7 +8,8 @@ export enum RfqmTranasctionSubmissionStatus {
     Successful = 'successful',
     Reverted = 'reverted',
     DroppedAndReplaced = 'droppedAndReplaced',
-    Confirmed = 'confirmed',
+    ConfirmedSuccessful = 'confirmedSuccessful',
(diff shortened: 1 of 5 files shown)


src/entities/RfqmTransactionSubmissionEntity.ts
```diff
@@ -8,6 +8,7 @@ export enum RfqmTranasctionSubmissionStatus {
     Successful = 'successful',
     Reverted = 'reverted',
     DroppedAndReplaced = 'droppedAndReplaced',
+    Confirmed = 'confirmed',
```
alexkroeger, line 11: confirmed refers to number of blocks deep, so it can be confirmed successful or confirmed reverted I think

alexkroeger: this will also have implications for the status function, I would update that in this PR as well, see https://github.com/0xProject/0x-api/blob/0f53d64779bfef344c2f7f66fb5bd69006e50c5f/src/services/rfqm_service.ts#L480

src/entities/RfqmTransactionSubmissionEntity.ts
```diff
@@ -68,7 +70,7 @@ export class RfqmTransactionSubmissionEntity {
         this.metadata = opts.metadata || null;
         this.nonce = opts.nonce || null;
         this.orderHash = opts.orderHash;
-        this.status = opts.status || RfqmTranasctionSubmissionStatus.Submitted;
+        this.status = opts.status || RfqmTransactionSubmissionStatus.Submitted;
```
PirosB3, line 73: Why do we want to add a default here? Can we just be explicit about what status we set?

src/entities/RfqmTransactionSubmissionEntity.ts
```diff
@@ -3,11 +3,13 @@ import { Column, Entity, Index, PrimaryColumn } from 'typeorm';
...
 import { BigIntTransformer, BigNumberTransformer } from './transformers';
 
-export enum RfqmTranasctionSubmissionStatus {
+export enum RfqmTransactionSubmissionStatus {
     Submitted = 'submitted',
     Successful = 'successful',
     Reverted = 'reverted',
     DroppedAndReplaced = 'droppedAndReplaced',
+    ConfirmedSuccessful = 'confirmedSuccessful',
+    ConfirmedReverted = 'confirmedReverted',
```
PirosB3, line 12: Can we document (with a comment maybe?) what is the difference between `Successful` and `ConfirmedSuccessful`?

phil-ociraptor:
@alexkroeger @PirosB3 @rhinodavid 

Given the work being done in https://github.com/0xProject/0x-api/pull/644 - I will be making the following changes to align this PR with that one:

- prefer the term `succeeded` to `successful`
- prefer `"snake_casing"` to `"camelCasing"`
- prioritize the terms `succeeded` and `failed` over `confirmed`

As a result, will have statuses that look like `succeeded_confirmed` as opposed to `confirmedSuccessful`

src/entities/RfqmTransactionSubmissionEntity.ts
phil-ociraptor: That's out of scope for this PR. I only corrected spelling of the enum from `RfqmTranasctionSubmissionStatus` to `RfqmTransactionSubmissionStatus`

src/entities/RfqmTransactionSubmissionEntity.ts
phil-ociraptor: Update coming that should disambiguate

alexkroeger:
@phil-ociraptor I think there needs to be an update to the function `_getJobStatusFromSubmissions` as well.

Also potentially worth writing the interim status of `SucceededUnconfirmed` or `RevertedUnconfirmed` after being mined but before confirmation.

phil-ociraptor:
@alexkroeger - I do believe the `SucceededUnconfirmed`/ `RevertedUnconfirmed` are getting written to the DB when `_checkSubmissionMapReceiptsAndUpdateDbAsync` is called after the first confirmation. After the 3rd, then they move to the confirmed status.

alexkroeger:
> @alexkroeger - I do believe the `SucceededUnconfirmed`/ `RevertedUnconfirmed` are getting written to the DB when `_checkSubmissionMapReceiptsAndUpdateDbAsync` is called after the first confirmation. After the 3rd, then they move to the confirmed status.

I meant on the job status, I believe we just update the transaction submission status right now, tho I could be wrong.

phil-ociraptor:
@alexkroeger - ahh gotcha, I added some changes to add that - what do you think of the logic in this commit: https://github.com/0xProject/0x-api/pull/642/commits/f0e11709a2b08a07a91e84ffce2ea1f687d772cf ?

src/services/rfqm_service.ts
```diff
@@ -205,9 +205,15 @@ export class RfqmService {
     private static _getJobStatusFromSubmissions(submissionsMap: SubmissionsMap): RfqmJobStatus {
         // there should only be one mined transaction, which will either be successful or a revert
```
alexkroeger, line 206: by the time we get here, there should only be one mined tx and it should be confirmed

src/services/rfqm_service.ts
```diff
@@ -205,9 +205,15 @@ export class RfqmService {
...
         for (const submission of Object.values(submissionsMap)) {
-            if (submission.status === RfqmTranasctionSubmissionStatus.Successful) {
+            if (
+                submission.status === RfqmTransactionSubmissionStatus.SucceededUnconfirmed ||
```
alexkroeger, line 209:
```suggestion
```

src/services/rfqm_service.ts
```diff
@@ -205,9 +205,15 @@ export class RfqmService {
...
+                submission.status === RfqmTransactionSubmissionStatus.SucceededConfirmed
+            ) {
                 return RfqmJobStatus.SucceededUnconfirmed;
-            } else if (submission.status === RfqmTranasctionSubmissionStatus.Reverted) {
+            } else if (
+                submission.status === RfqmTransactionSubmissionStatus.RevertedUnconfirmed ||
```
alexkroeger, line 214:
```suggestion
```

src/services/rfqm_service.ts
```diff
@@ -205,9 +205,15 @@ export class RfqmService {
     private static _getJobStatusFromSubmissions(submissionsMap: SubmissionsMap): RfqmJobStatus {
         // there should only be one mined transaction, which will either be successful or a revert
         for (const submission of Object.values(submissionsMap)) {
-            if (submission.status === RfqmTranasctionSubmissionStatus.Successful) {
+            if (
+                submission.status === RfqmTransactionSubmissionStatus.SucceededUnconfirmed ||
+                submission.status === RfqmTransactionSubmissionStatus.SucceededConfirmed
+            ) {
                 return RfqmJobStatus.SucceededUnconfirmed;
```
alexkroeger, line 212:
```suggestion
                return RfqmJobStatus.SucceededConfirmed;
```

src/services/rfqm_service.ts
```diff
@@ -205,9 +205,15 @@ export class RfqmService {
...
-            } else if (submission.status === RfqmTranasctionSubmissionStatus.Reverted) {
+            } else if (
+                submission.status === RfqmTransactionSubmissionStatus.RevertedUnconfirmed ||
+                submission.status === RfqmTransactionSubmissionStatus.RevertedConfirmed
+            ) {
                 return RfqmJobStatus.FailedRevertedUnconfirmed;
```
alexkroeger, line 217:
```suggestion
                return RfqmJobStatus.FailedRevertedConfirmed;
```

alexkroeger (approved):
lgtm! thanks for all the diligence on this one

PR #648, opened 2021-06-16
refactor: rewrite _checkSubmissionMapReceiptsAndUpdateDbAsync without a nested for loop

phil-ociraptor:
# Description

Small refactor to make this function slightly more readable. The goal was to remove the double for loop so the reader doesn't have to think about iteration

# Testing Instructions

# Checklist

    -   [ ] SRA/Limit orders
    -   [ ] Swap endpoints
    -   [ ] Depth charts

### DIFF
--- a/src/services/rfqm_service.ts
+++ b/src/services/rfqm_service.ts
@@ -804,68 +804,76 @@ export class RfqmService {
                 };
             }),
         );
+        const minedReceipt = receipts.find((r) => r.response !== undefined);
 
-        for (const receipt of receipts) {
-            if (receipt.response !== undefined) {
-                isTxMined = true;
-                const currentBlock = await this._blockchainUtils.getCurrentBlockAsync();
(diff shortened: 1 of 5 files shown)


src/services/rfqm_service.ts
```diff
@@ -804,68 +804,80 @@ export class RfqmService {
...
+
+        // A tx was mined
+        isTxMined = true;
+        const currentBlock = await this._blockchainUtils.getCurrentBlockAsync();
+        if (currentBlock - minedReceipt.response!.blockNumber >= BLOCK_FINALITY_THRESHOLD) {
+            isTxConfirmed = true;
+        }
+
+        // Update the SubmissionMap for this mined tx
+        if (minedReceipt.response!.status === 1) {
```
alexkroeger, line 826: does typescript require the declaration that response is defined here?

rhinodavid: Might be able to optional chain here

PR #658, opened 2021-06-22
feat: clear out calldata after last look declined

phil-ociraptor:
# Description

In order to make sure we never submit a job after the last look has been declined, we'll clear out the calldata.

# Testing Instructions

# Checklist

    -   [ ] SRA/Limit orders
    -   [ ] Swap endpoints
    -   [ ] Depth charts

### DIFF
--- a/src/runners/http_rfqm_service_runner.ts
+++ b/src/runners/http_rfqm_service_runner.ts
@@ -100,7 +100,7 @@ export async function buildRfqmServiceAsync(connection: Connection, asWorker: bo
     axiosInstance.defaults.raxConfig = {
         retry: 3, // Retry on 429, 500, etc.
         noResponseRetries: 0, // Do not retry on timeouts
-        instance: axiosInstance
+        instance: axiosInstance,
     };
     rax.attach(axiosInstance);
     const quoteRequestor = new QuoteRequestor(
--- a/src/runners/http_rfqm_service_runner.ts
+++ b/src/runners/http_rfqm_service_runner.ts
@@ -11,9 +11,9 @@ import * as express from 'express';
 import * as core from 'express-serve-static-core';
 import { Agent as HttpAgent, Server } from 'http';
 import { Agent as HttpsAgent } from 'https';
+import * as rax from 'retry-axios';
 import { Producer } from 'sqs-producer';
 import { Connection } from 'typeorm';
-import * as rax from 'retry-axios';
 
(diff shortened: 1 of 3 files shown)


src/runners/http_rfqm_service_runner.ts
```diff
@@ -96,6 +97,12 @@ export async function buildRfqmServiceAsync(connection: Connection, asWorker: bo
 
     const contractAddresses = await getContractAddressesForNetworkOrThrowAsync(provider, CHAIN_ID);
     const axiosInstance = Axios.create(getAxiosRequestConfig());
+    axiosInstance.defaults.raxConfig = {
+        retry: 3, // Retry on 429, 500, etc.
```
PirosB3, line 101: Should we just retry on 5xx? a 4xx is considered to be a reasonable rejection to "last look" right?

src/services/rfqm_service.ts
```diff
@@ -698,6 +698,7 @@ export class RfqmService {
             // Terminate with an error transition
             await this._dbUtils.updateRfqmJobAsync(orderHash, {
                 status: RfqmJobStatus.FailedLastLookDeclined,
+                calldata: '', // clear out calldata so transaction can never be submitted, even by accident
```
PirosB3, line 701: if we are clearing calldata here, do we need to handle some situation where a job is picked up by the worker and the calldata for that order was blanked out?

src/runners/http_rfqm_service_runner.ts
phil-ociraptor: Yup, by default it retries all 5xx and 429 Too Many Requests

src/services/rfqm_service.ts
phil-ociraptor: ah - yes, so you're referring to a strange retry scenario. Because the status and the calldata get updated in the same query, I don't think there's a huge risk to entering some illegal state.

However, that said, I believe Kroeger's "repair" mode will fix even those types of issues.

src/runners/http_rfqm_service_runner.ts
PirosB3: oh I see, thnanks

src/services/rfqm_service.ts
PirosB3: That makes sense, thanks!

PirosB3 (approved):

PR #663, opened 2021-06-24
feat: Scaffold RFQm health check endpoint

rhinodavid:
# Description
Create the route, handler, and service scaffold for the RFQm health check endpoint.

# Testing Instructions
Unit and integration tests are included.

# Checklist

    -   [ ] SRA/Limit orders
    -   [ ] Swap endpoints
    -   [ ] Depth charts


src/services/rfqm_service.ts
```diff
@@ -573,6 +574,29 @@ export class RfqmService {
         }
     }
 
+    /**
+     * Runs checks to determine the health of the RFQm system. The results may be distilled to a format needed by integrators.
+     */
+    // tslint:disable-next-line prefer-function-over-method
+    public async runHealthCheckAsync(): Promise<HealthCheckResult> {
+        const transformPairs = (offerings: RfqMakerAssetOfferings): { [pair: string]: HealthCheckStatus } =>
```
alexkroeger, line 582: can we define this outside the function?

src/utils/rfqm_health_check.ts
```diff
@@ -0,0 +1,50 @@
...
+export interface HealthCheckResult {
+    status: HealthCheckStatus;
+    pairs: {
+        [pair: string]: HealthCheckStatus; // where the pair has the form `${contractA}-${contractB}`
+    };
+    http: {
+        status: HealthCheckStatus;
+        issues: HealthCheckIssue[];
+    };
+    workers: {
```
alexkroeger, line 23: would it make sense to make this an array or object so each worker is represented?

src/utils/rfqm_health_check.ts
```diff
@@ -0,0 +1,50 @@
...
+}
+
+interface HealthCheckIssue {
+    status: HealthCheckStatus;
+    description: string;
+}
+
+export interface HealthCheckResult {
+    status: HealthCheckStatus;
+    pairs: {
```
alexkroeger, line 16: slight preference for array over key, value here to avoid needing to string parse the tokens out of the key

src/utils/rfqm_health_check.ts
```diff
@@ -0,0 +1,50 @@
+export enum HealthCheckStatus {
+    Operational = 'operational',
+    Unknown = 'unknown',
```
alexkroeger, line 3: still unclear to me how to use the 'unknown',  a component that can't report its status I would assume is down

alexkroeger (approved):
left some comments to consider but lgtm!

src/utils/rfqm_health_check.ts
rhinodavid: Mmm...not a bad idea maybe. I'll keep it in mind when I start building that part out.

src/services/rfqm_service.ts
rhinodavid: Do you think that's better?

I think some engineers would see this as odd, but that's because so much of the industry is traumatized by Java (see ["Execution in the Kingdom of Nouns" rant](https://steve-yegge.blogspot.com/2006/03/execution-in-kingdom-of-nouns.html). If you're from that world you'd certainly make this a method, but that's because you have no other choice. Since `runHealthCheckAsync` is the only function concerned with the `transformPairs` logic, why move it to a broader context when no other member of that context cares about it? My inclination is actually the opposite: make it an IIFE and define and invoke it at the same time. But my sense is that this is even more odd to folks.

src/utils/rfqm_health_check.ts
rhinodavid: Ack.

Thought here is that this enables pair-level checks/reporting... i.e. all the transactions on pair X have failed for the last hour. We mark it as `FAILED`, then just don't send it to the front end.

src/utils/rfqm_health_check.ts
rhinodavid: I'll cut it if we end up not needing it (see the discussion in the design about when this would be used).

src/services/rfqm_service.ts
alexkroeger: I find it to be more readable personally. You can more quickly see what `runHealthCheckAsync` is doing (returning operational and applying a transformation to the pairs), and you might not care about digging into the `transformPairs` logic--but if you do then you can check out the function separately.

If we define the logic for every utility `runHealthCheckAsync` needs within the function itself, it could get crowded and hard to follow.

phil-ociraptor: I forgot to press submit, but here's my two cents:

There are other contexts where this could be helpful (to short circuit certain code paths). Additionally, since this is pulling in an environment variable that won't change until a restart, we can pay the compute cost once on startup - as opposed to once per request. My preference would be to have this be done once on startup, and stored in memory as a singleton to be reused across contexts. 

We can do this later thought

rhinodavid: This is one of those great coding debates that we can go back and forth on forever. It is my hope that once I am older and grey I will have my reasonings so well ironed out I can convince anybody in like five minutes.

The one note I'll add here is the env variable already has the behavior you want @phil-ociraptor. That config file grabs it once and that's that.

PR #665, opened 2021-06-25
feat: Create RFQm queue checks

rhinodavid:
# Description

Creates checks which examine the SQS queue for a buildup of messages or for old messages which may be expired or expire soon.

TODO: Add the queue check to the health check

# Testing Instructions

Tests with mocked queue responses are included.

# Checklist

    -   [ ] SRA/Limit orders
    -   [ ] Swap endpoints
    -   [ ] Depth charts

### DIFF
--- a/src/utils/rfqm_health_check.ts
+++ b/src/utils/rfqm_health_check.ts
@@ -2,8 +2,6 @@ import { Producer } from 'sqs-producer';
 
 const SQS_QUEUE_SIZE_DEGRADED_THRESHOLD = 10; // More messages sitting in queue than this will cause a DEGRADED issue
 const SQS_QUEUE_SIZE_FAILED_THRESHOLD = 20; // More messages sitting in queue than this will cause a FAILED issue
-const SQS_MESSAGE_AGE_MINUTES_DEGRADED_THRESHOLD = 3; // A message older than this in the queue will cause a DEGRADED issue
-const SQS_MESSAGE_AGE_MINUTES_FAILED_THRESHOLD = 5; // A message older than this in the queue will cause a FAILED issue
 
 export enum HealthCheckStatus {
     Operational = 'operational',
@@ -39,8 +37,6 @@ export interface RfqmHealthCheckShortResponse {
     pairs: [string, string][];
 }
 
-const MILLISECONDS_IN_MINUTE = 60 * 1000; // tslint:disable-line custom-no-magic-numbers
-
 /**
  * Transform the full health check result into the minimal response the Matcha UI requires.
  */
@@ -59,10 +55,9 @@ export function transformResultToShortResponse(result: HealthCheckResult): RfqmH
 }
 
 /**
- * Runs checks on the SQS queue to detect if there are messages piling up or if there are old messages in the
- * queue which have expired or will expire soon. `nowTime` parameter is provided for testing.
+ * Runs checks on the SQS queue to detect if there are messages piling up.
  */
-export async function checkSqsQueueAsync(producer: Producer, nowTime: Date = new Date()): Promise<HealthCheckIssue[]> {
+export async function checkSqsQueueAsync(producer: Producer): Promise<HealthCheckIssue[]> {
     const results: HealthCheckIssue[] = [];
     const messagesInQueue = await producer.queueSize();
     if (messagesInQueue === 0) {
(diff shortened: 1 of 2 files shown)


src/utils/rfqm_health_check.ts
```diff
@@ -48,3 +57,59 @@ export function transformResultToShortResponse(result: HealthCheckResult): RfqmH
...
+            status: HealthCheckStatus.Failed,
+            description: `SQS queue contains ${messagesInQueue} messages (threshold is ${SQS_QUEUE_SIZE_FAILED_THRESHOLD})`,
+        });
+    } else if (messagesInQueue > SQS_QUEUE_SIZE_DEGRADED_THRESHOLD) {
+        results.push({
+            status: HealthCheckStatus.Degraded,
+            description: `SQS queue contains ${messagesInQueue} messages (threshold is ${SQS_QUEUE_SIZE_DEGRADED_THRESHOLD})`,
+        });
+    }
+    const { Messages: messages } = await producer.sqs
```
PirosB3, line 82: Could we avoid popping off items from the queue? It feels fairly "invasive" to make DB changes for a healthcheck. Can we stick to simply looking at queue backlog?

alexkroeger, line 82: Pulling these off the queue makes me nervous, I would opt for just doing the queue size checks

src/utils/rfqm_health_check.ts
```diff
@@ -1,3 +1,10 @@
+import { Producer } from 'sqs-producer';
+
+const SQS_QUEUE_SIZE_DEGRADED_THRESHOLD = 5; // More messages sitting in queue than this will cause a DEGRADED issue
```
alexkroeger, line 3: I would make these numbers less aggressive to start:

e.g. 10 degraded, 20 failed

PirosB3 (commented):
Thank you for this! Who should be running this function? Should it be run by a worker process or another external process?

alexkroeger (changes requested):
I have concerns about pulling items off the queue, let's sync about it if you think it's important to include them

src/utils/rfqm_health_check.ts
rhinodavid: Ok we figured out how to not make it hide them at all.

rhinodavid:
> Thank you for this! Who should be running this function? Should it be run by a worker process or another external process?

@PirosB3 

This goes in the full health check logic! It will run when someone hits `/healthz`, which will cache the results for some period of time. I was also just chatting with @alexkroeger and I think maybe we can have a cron or something that runs it every minute or something and we can turn the output into a dashboard.

src/utils/rfqm_health_check.ts
rhinodavid: fin

src/utils/rfqm_health_check.ts
```diff
@@ -48,3 +57,59 @@ export function transformResultToShortResponse(result: HealthCheckResult): RfqmH
...
+            status: HealthCheckStatus.Degraded,
+            description: `SQS queue contains ${messagesInQueue} messages (threshold is ${SQS_QUEUE_SIZE_DEGRADED_THRESHOLD})`,
+        });
+    }
+    const { Messages: messages } = await producer.sqs
+        .receiveMessage({
+            MaxNumberOfMessages: 10 /* API maximum */,
+            QueueUrl: producer.queueUrl,
+            AttributeNames: ['SentTimestamp'],
+            VisibilityTimeout: 0 /* sec (don't make the messages invisible to workers) */,
```
alexkroeger, line 87: 👍

alexkroeger (approved):
lgtm with the `visibilityTimeout` change to 0 (so the webserver is not holding workers up from reading tasks)

approved without integration into /healthz, if you add that to this PR get a double check

src/utils/rfqm_health_check.ts
```diff
@@ -48,3 +57,59 @@ export function transformResultToShortResponse(result: HealthCheckResult): RfqmH
...
+            description: `SQS queue contains ${messagesInQueue} messages (threshold is ${SQS_QUEUE_SIZE_FAILED_THRESHOLD})`,
+        });
+    } else if (messagesInQueue > SQS_QUEUE_SIZE_DEGRADED_THRESHOLD) {
+        results.push({
+            status: HealthCheckStatus.Degraded,
+            description: `SQS queue contains ${messagesInQueue} messages (threshold is ${SQS_QUEUE_SIZE_DEGRADED_THRESHOLD})`,
+        });
+    }
+    const { Messages: messages } = await producer.sqs
+        .receiveMessage({
```
PirosB3, line 83: Are we sure that this is doing exactly what we are expecting it to do? Is there any documentation around receiving messages with 0 visibility window? Could we use the Postgres database instead to get this kind of information, instead of peeking the queue?

src/utils/rfqm_health_check.ts
```diff
@@ -48,3 +57,59 @@ export function transformResultToShortResponse(result: HealthCheckResult): RfqmH
...
+            MaxNumberOfMessages: 10 /* API maximum */,
+            QueueUrl: producer.queueUrl,
+            AttributeNames: ['SentTimestamp'],
+            VisibilityTimeout: 0 /* sec (don't make the messages invisible to workers) */,
+        })
+        .promise();
+    if (messages === undefined) {
+        throw new Error("SQS response did not include 'Messages'");
+    }
+    const messageTimestamps = messages.map((m) => parseInt(m.Attributes!.SentTimestamp, 10)).sort();
```
PirosB3, line 93: Can we avoid a crash in case the message does not have `SentTime` attribute? Can we filter out messages that do not have this attribute?

src/utils/rfqm_health_check.ts
```diff
@@ -48,3 +57,59 @@ export function transformResultToShortResponse(result: HealthCheckResult): RfqmH
...
+    const oldestTimestamp = messageTimestamps[0];
```
PirosB3, line 94: If the array is empty, will this return `undefined`?

rhinodavid:
Remove the message age check since ppl were afraid to look at the messages.

PR #673, opened 2021-07-01
feat: add functionality to dynamically block addresses

phil-ociraptor:
# Description

Introduces `RfqBlockedAddressUtils` which allows us to dynamically block addresses 

# Testing Instructions

Integration tests added

# Checklist

    -   [ ] SRA/Limit orders
    -   [ ] Swap endpoints
    -   [ ] Depth charts


src/utils/rfq_blacklist_utils.ts
```diff
@@ -0,0 +1,47 @@
+import { Connection } from 'typeorm/connection/Connection';
+
+import { BlockedAddressEntity } from '../entities/BlockedAddressEntity';
+
+/**
+ * RfqBlacklistUtils helps manage the RFQ blacklist
+ */
+export class RfqBlacklistUtils {
+    private static _instance: RfqBlacklistUtils;
```
PirosB3, line 9: is this a singleton pattern?

PirosB3: do we need it? can we just initialize the class once and pass by reference?

migrations/1625095491619-CreateBlockedAddressesTable.ts
```diff
@@ -0,0 +1,23 @@
+import { MigrationInterface, QueryRunner, Table, TableIndex } from 'typeorm';
+
+const blockedAddressesTable = new Table({
+    name: 'blocked_addresses',
+    columns: [
+        { name: 'address', type: 'varchar', isPrimary: true },
```
alexkroeger, line 6: might be worth forcing to lowercase to avoid duplicate addrs

src/utils/rfq_blacklist_utils.ts
```diff
@@ -0,0 +1,47 @@
...
+    private _blacklist: Set<string>;
+    private readonly _intervalTimeout: NodeJS.Timeout;
+
+    public static getInstance(
```
PirosB3, line 13: Instead of using a timer to continuously refresh items, have we considered storing the last time we fetched the list, and simply reload whenever someone calls `isBlocked` and too much time has passed? would avoid having an other event loop running (we have like billions in the 0x API)

src/utils/rfq_blacklist_utils.ts
```diff
@@ -0,0 +1,47 @@
...
+
+        const intervalTimeout = setInterval(async () => {
+            await this._updateBlacklistAsync();
+        }, intervalMs);
+        this._intervalTimeout = intervalTimeout;
+    }
+
+    private async _updateBlacklistAsync(): Promise<void> {
+        const blockedAddresses = await this._connection.getRepository(BlockedAddressEntity).find();
+        this._blacklist = new Set(blockedAddresses.map((entity) => entity.address));
```
PirosB3, line 45: just for safety, can we `.toLowerCase()`?

test/utils/rfq_blacklist_utils_test.ts
```diff
@@ -0,0 +1,62 @@
...
+import { RfqBlacklistUtils } from '../../src/utils/rfq_blacklist_utils';
+import { setupDependenciesAsync, teardownDependenciesAsync } from '../utils/deployment';
+
+// Force reload of the app avoid variables being polluted between test suites
+delete require.cache[require.resolve('../../src/app')];
+
+const SUITE_NAME = 'rfqBlacklistUtils';
+const refreshIntervalMs = 1000;
+
+describe.only(SUITE_NAME, () => {
```
PirosB3, line 17: `.only()` should go

PirosB3 (changes requested):

migrations/1625095491619-CreateBlockedAddressesTable.ts
phil-ociraptor: nice - good thought, will add a check constraint

src/utils/rfq_blacklist_utils.ts
phil-ociraptor: good point, this is unnecessary because all of this is in the API and we full control of the reference. Will remove. I was just blindly copying a pattern from elsewhere yesterday when I got tired

src/utils/rfq_blacklist_utils.ts
phil-ociraptor: give the new impl a look :D

migrations/1625095491619-CreateBlockedAddressesTable.ts
```diff
@@ -0,0 +1,40 @@
...
+        { name: 'ignore', type: 'boolean', default: false },
+    ],
+});
+const createdAtIndex = new TableIndex({ name: `blocked_addresses_created_at_index`, columnNames: ['created_at'] });
+const ignoreIndex = new TableIndex({ name: `blocked_addresses_ignore_index`, columnNames: ['ignore'] });
+
+export class CreateBlockedAddressesTable1625095491619 implements MigrationInterface {
+    public async up(queryRunner: QueryRunner): Promise<void> {
+        const driver = queryRunner.connection.driver;
+        const addressLowerCaseConstraint = new TableCheck({
```
alexkroeger, line 19: nice 👍 

could be worth adding a test for this

src/utils/rfq_blocked_address_utils.ts
```diff
@@ -0,0 +1,64 @@
...
+
+        // Guaranteed to be a fresh value
+        return this._blocked.has(address.toLowerCase());
+    }
+
+    /**
+     * Updates the blocked set of addresses
+     */
+    private async _updateBlockedSetAsync(): Promise<void> {
+        this._updating = true;
```
alexkroeger, line 58: nice 👍

migrations/1625095491619-CreateBlockedAddressesTable.ts
alexkroeger: should at least check locally after running migrations @phil-ociraptor

src/utils/rfq_blocked_address_utils.ts
```diff
@@ -0,0 +1,64 @@
...
+        this._ttlMs = ttlMs;
+        this._updating = false;
+        this._expiresAt = Date.now().valueOf(); // cache expires immediately
+    }
+
+    /**
+     * isBlocked returns whether an address is blocked from the cache
+     * NOTE: In the background, it also updates the blocked set if the cache is expired
+     */
+    public isBlocked(address: string): boolean {
```
alexkroeger, line 26: I imagine we'll just use the isBlocked function, wdyt?

test/utils/rfq_blocked_address_utils_test.ts
```diff
@@ -0,0 +1,83 @@
...
+    it('should use stale values via isBlocked', async () => {
+        const sampleBadAddress = '0xA10612Ee5432B6395d1F0d6fB2601299a1c64274';
+        expect(rfqBlacklistUtils.isBlocked(sampleBadAddress)).to.be.false();
+
+        // Add it to the blocked list
+        await connection.getRepository(BlockedAddressEntity).save({
+            address: sampleBadAddress.toLowerCase(),
+        });
+
+        expect(rfqBlacklistUtils.isBlocked(sampleBadAddress)).to.be.false();
```
alexkroeger, line 51: could add another check to wait TTL and check `isBlocked` again, should be true

src/entities/BlockedAddressEntity.ts
```diff
@@ -0,0 +1,37 @@
...
+    public address: string;
+
+    @Index()
+    @Column({ name: 'created_at', type: 'timestamptz', default: () => 'now()' })
+    public createdAt!: Date;
+
+    @Column({ name: 'parent', type: 'varchar', nullable: true })
+    public parent: string | null;
+
+    @Column({ name: 'last_seen_nonce', type: 'numeric', nullable: true, transformer: BigNumberTransformer })
```
alexkroeger, line 20: this is set as numeric in the entity definition but bigint in the migration, I'd label this type bigint and just use number since the range won't require BigNumber

alexkroeger (changes requested):
looks good overall, main change is to make migration and entity definition consistent

src/utils/rfq_blocked_address_utils.ts
```diff
@@ -0,0 +1,64 @@
...
+        // Guaranteed to be a fresh value
+        return this._blocked.has(address.toLowerCase());
+    }
+
+    /**
+     * Updates the blocked set of addresses
+     */
+    private async _updateBlockedSetAsync(): Promise<void> {
+        this._updating = true;
+        const blockedAddresses = await this._connection.getRepository(BlockedAddressEntity).find();
```
PirosB3, line 59: can we add a `LIMIT` here just to be safe? just to avoid our system from blowing up and make it more determinisric

PirosB3: For example:
- Limit query to 5000 addresses
- If query returns 5000 addresses, log an error (so that we are informed that the blocked entities have passed a certain threshold)

PirosB3: Do we read from our slave node in this case? or do we use the master node? do we know?

PirosB3 (approved):
Just one strong suggestion about setting a `LIMIT` to some value (just to avoid some risk).

If you feel like it (I know it can be a pain to do more work) would you mind adding some prometheus counters? such as:
- Gauge for # addresses in the blacklist
- Counter for every time the blacklist is updated

src/entities/BlockedAddressEntity.ts
phil-ociraptor: good catch!

src/utils/rfq_blocked_address_utils.ts
phil-ociraptor: correct - the `isBlockedAsync` version is just there to make the test deterministic

test/utils/rfq_blocked_address_utils_test.ts
phil-ociraptor: so there isn't a guarantee that the TTL will be enough for the DB update to happen. What we would need is to `await` on the promise of the DB update (which is currently private).

That said, waiting something like 1 second would pretty much guarantee the DB update finishes - but it in theory is a potential race condition.

Alternatively, could expose the `updatePromise` and await on it. Will noodle on this some more

migrations/1625095491619-CreateBlockedAddressesTable.ts
phil-ociraptor: ah yeah, an automated test would be good

And the local version looks good 😎 

```
api=# insert into blocked_addresses (address) values ('HARRY');
ERROR:  new row for relation "blocked_addresses" violates check constraint "CHK_39e5b70f27376dfac15dd3daf3"
DETAIL:  Failing row contains (HARRY, 2021-07-01 23:33:34.704671+00, null, null, f).
api=# insert into blocked_addresses (address) values ('harry');
INSERT 0 1
```

src/utils/rfq_blocked_address_utils.ts
phil-ociraptor: removed it - got something a little better

test/utils/rfq_blocked_address_utils_test.ts
phil-ociraptor: found something similar to exposing the `updatePromise`, but slightly better

src/utils/rfq_blocked_address_utils.ts
phil-ociraptor: ✅  limit 5000 with logging

regarding the replica, it depends on if the master/slave configuration is enabled: https://typeorm.io/#/multiple-connections/replication

In our case, we do use the master/slave configuration so _all_ `find` queries go to one of the read replicas by default, and all inserts and updates will go to the write master.

test/utils/rfq_blocked_address_utils_test.ts
```diff
@@ -0,0 +1,116 @@
...
+        await connection.getRepository(BlockedAddressEntity).save({
+            address: sampleBadAddress.toLowerCase(),
+        });
+
+        // Initally not blocked
+        const isBlocked_t0 = rfqBlacklistUtils.isBlocked(sampleBadAddress);
+        expect(isBlocked_t0).to.be.false();
+
+        // Await for the update to complete
+        await rfqBlacklistUtils.completeUpdateAsync();
```
alexkroeger, line 90: 👍

alexkroeger (approved):

phil-ociraptor:
deploy staging

PR #683, opened 2021-07-09
fix: check readiness before processing unresolved jobs

phil-ociraptor:
# Description

# Testing Instructions

# Checklist

    -   [ ] SRA/Limit orders
    -   [ ] Swap endpoints
    -   [ ] Depth charts

### DIFF
--- a/src/services/rfqm_service.ts
+++ b/src/services/rfqm_service.ts
@@ -529,18 +529,19 @@ export class RfqmService {
         const balanceUnitAmount = Web3Wrapper.toUnitAmount(balance, ETH_DECIMALS).decimalPlaces(PRICE_DECIMAL_PLACES);
         RFQM_WORKER_BALANCE.labels(workerAddress).set(balanceUnitAmount.toNumber());
 
+        // Check if worker is ready to process jobs
+        const isWorkerReady = await this._blockchainUtils.isWorkerReadyAsync(workerAddress, balance, gasPrice);
+        if (!isWorkerReady) {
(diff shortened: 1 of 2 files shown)


src/services/rfqm_service.ts
```diff
@@ -529,6 +529,13 @@ export class RfqmService {
         const balanceUnitAmount = Web3Wrapper.toUnitAmount(balance, ETH_DECIMALS).decimalPlaces(PRICE_DECIMAL_PLACES);
         RFQM_WORKER_BALANCE.labels(workerAddress).set(balanceUnitAmount.toNumber());
 
+        // Check if worker is ready to process jobs
```
alexkroeger, line 532: I don't agree with putting the readiness check before checking for unresolved jobs.

Having an unresolved job means that the worker previously reached the submission step and then exited for some reason before finishing.

To avoid having multiple workers compete to submit a transaction (and therefore having the guarantee of n-1 reverts for n workers that submit a tx), we lock in a worker at the submission step.

If there are unresolved jobs, the worker should try to reach an end state with the transaction regardless of its readiness.

This means either:
- monitoring existing submissions and updating our DB to their finalized state
- or updating the DB to reflect that it the transaction is failed, because the worker cannot submit it.

The latter shouldn't happen in practice. If the worker has an unresolved job, that means it was ready (had sufficient balance ahead of time), and therefore the only case in which this should happen in theory is the worker had sufficient balance for 1 job, it exited before confirmation, the confirmed tx has decreased its balance below the readiness threshold.

phil-ociraptor:
This change would brick our repair process - closing

PR #714, opened 2021-08-16
fix: Added metrics proxy

PirosB3:
Adds more metrics to 0x API for RFQM and RFQT

### DIFF
--- a/src/runners/rfqm_worker_runner.ts
+++ b/src/runners/rfqm_worker_runner.ts
@@ -24,27 +24,27 @@ import { SqsConsumer } from '../utils/sqs_consumer';
 import { buildRfqmServiceAsync } from './http_rfqm_service_runner';
 
 const RFQM_JOB_DEQUEUED = new Counter({
-    name: "rfqm_job_dequeued",
-    help: "An Rfqm Job was pulled from the queue",
-    labelNames: ["address"],
+    name: 'rfqm_job_dequeued',
+    help: 'An Rfqm Job was pulled from the queue',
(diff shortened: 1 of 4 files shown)


src/runners/rfqm_worker_runner.ts
```diff
@@ -89,21 +98,23 @@ export async function runRfqmWorkerAsync(rfqmService: RfqmService, workerAddress
...
         beforeHandle: async () => rfqmService.workerBeforeLogicAsync(workerAddress),
         handleMessage: async (message) => {
-            RFQM_JOB_DEQUEUED.inc();
+            RFQM_JOB_DEQUEUED.labels(workerAddress).inc();
             const { orderHash } = JSON.parse(message.Body!);
             logger.info({ workerAddress, orderHash }, 'about to process job');
-            return rfqmService.processRfqmJobAsync(orderHash, workerAddress);
+            const stopTimer = RFQM_JOB_PROCESSING_TIME.labels(workerAddress).startTimer();
+            rfqmService.processRfqmJobAsync(orderHash, workerAddress);
+            stopTimer();
```
phil-ociraptor, line 106: can we keep the `return` types the same? Forget if there's something that depends on the return type at all - most likely not, but just want to keep it consistent

phil-ociraptor (approved):

src/runners/rfqm_worker_runner.ts
PirosB3: definitely! good spot!

PR #717, opened 2021-08-19
feat: [rfqm] Get transaction hash and store before submitting

rhinodavid:
~~**Requires 0xProject/protocol#310 to push before tests will pass**~~

**Closes MKR-96**

- Adds Ethers as a direct dependency of 0x API
- Adds function to sign and hash a transaction using ethers
- Adds function to submit a signed transaction still using web3wrapper
- Adds a Presubmit status to the RFQM Transaction Status enum

Now, when a worker is going to submit a transaction, the flow is:
1. Sign and hash the transaction offline
1. Save the transaction to the rfqm_transaction_submission table with a status of Presubmit
1. Submit the transaction to the mempool
1. Ensure the hash computed offline matches the hash from submission (a unit test is added to verify this works as expected, but the implementation also checks and logs an error if they differ)
1. Update the row in rfqm_transaction_submission with a status of Submitted

## Testing Instructions
Unit test added to ensure that the offline hash is the same as the hash from the rpc provider.

Checklist
 Update documentation as needed. Website Documentation PR:

 Prefix PR title with [WIP] if necessary.

 Add tests to cover changes as needed.

 Test changes on the staging environment with Matcha API Staging.

 SRA/Limit orders
 Swap endpoints
 Meta transaction endpoints
 Depth charts
For more information see 0x API Matcha smoke test runbook in Quip.


src/runners/http_rfqm_service_runner.ts
```diff
@@ -95,8 +103,12 @@ export async function buildRfqmServiceAsync(connection: Connection, asWorker: bo
             META_TX_WORKER_MNEMONIC,
             RFQM_WORKER_INDEX,
         );
+
+        // TODO (rhinodavid): Remove once migration to ethers.js is complete
         const privateWalletSubprovider = new PrivateKeyWalletSubprovider(workerPrivateKey);
         provider = RfqBlockchainUtils.createPrivateKeyProvider(rpcProvider, privateWalletSubprovider);
+
+        ethersWallet = Wallet.fromMnemonic(META_TX_WORKER_MNEMONIC, `m/44'/60'/0'/0/${RFQM_WORKER_INDEX!}`);
```
phil-ociraptor, line 111: there's a few functions in `rfq_blockchain_utils` such as `_getPathByIndex` that could reduce duplication of the `m/44'/60'/0'/0/` stuff. That suggests that we could create a new function called `createWallet` on `rfq_blockchain_utils`

src/services/rfqm_service.ts
```diff
@@ -1112,34 +1112,54 @@ export class RfqmService {
...
-        return this._dbUtils.writeRfqmTransactionSubmissionToDbAsync(partialEntity);
+        await this._dbUtils.writeRfqmTransactionSubmissionToDbAsync(partialEntity);
+
+        const transactionHashFromSubmit = await this._blockchainUtils.submitSignedTransactionAsync(signedTransaction);
+
+        if (hash !== transactionHashFromSubmit) {
+            // This should never ever happen
+            logger.error(
+                `Mismatch between transaction hash calculated before submit and after submit:\nBefore submit:\t${hash}\nAfter submit:\t${transactionHashFromSubmit}`,
+            );
```
phil-ociraptor, line 1149: to make this log easier to search upon, can we use the following syntax:
```
logger.error( {
 transactionHashFromSubmit,
 transactionHash: hash
}, "Mismatch between transaction hash calculated before submit and after submit");
```

This will allow us to search on the `transactionHash` and trace it's journey through the log messages

src/services/rfqm_service.ts
```diff
@@ -1112,34 +1112,54 @@ export class RfqmService {
...
+        }
+
+        logger.info(
+            { orderHash, workerAddress, transactionHashFromSubmit },
```
phil-ociraptor: nit: can we rename `transactionHashFromSubmit` as just `transactionHash`, just to make the log tracing easier

src/utils/rfq_blockchain_utils.ts
```diff
@@ -231,4 +256,56 @@ export class RfqBlockchainUtils {
...
+    }
+
+    /**
+     * Uses the Ethers Wallet to sign a transaction. Returns both the signed transaction and its hash.
+     *
+     * If the containing class has been initialized without a wallet, the function throws.
+     */
+    public async signTransactionAsync(
+        transaction: providers.TransactionRequest,
+    ): Promise<{ signedTransaction: string; hash: string }> {
```
phil-ociraptor, line 299: nit: can we return `transactionHash` instead of the term `hash` - will keep it consistent whenever we use the term `transactionHash`

phil-ociraptor:
First pass done - looks really good from a high level - just some nits

src/services/rfqm_service.ts
rhinodavid: Oh good call

src/utils/rfq_blockchain_utils.ts
rhinodavid: sgtm I was going back and forth about this

src/services/rfqm_service.ts
rhinodavid: good call i was going back and forth about this one

src/runners/http_rfqm_service_runner.ts
rhinodavid: Ack. My preference is actually to rip out all those functions once we get rid of web3-wrapper. This line is concise and clear (I think) and we can get rid of a lot of code we have to maintain.

src/runners/http_rfqm_service_runner.ts
```diff
@@ -83,6 +84,13 @@ if (require.main === module) {
 export async function buildRfqmServiceAsync(connection: Connection, asWorker: boolean): Promise<RfqmService> {
     let provider: SupportedProvider;
 
+    // ether.js Provider coexists with web3 provider during migration away from 0x/web3-wrapper.
+    const ethersProvider = new providers.JsonRpcProvider(
```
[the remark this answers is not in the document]
rhinodavid: Yep ethers specifically separates the functionality (which is nice imho). Unless you want to nix web3wrapper we have to have both

src/utils/rfq_blockchain_utils.ts
```diff
@@ -231,4 +256,56 @@ export class RfqBlockchainUtils {
...
+                throw new Error("An Ethers wallet must be provided if no 'from' address is specified");
+            }
+            from = await this._ethersWallet.getAddress();
+        }
+
+        return {
+            to: txOptions.to || this._exchangeProxy.address,
+            from,
+            nonce: txOptions.nonce,
+            gasLimit: EthersBigNumber.from(txOptions.gas),
```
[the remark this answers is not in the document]
rhinodavid: The order format changes -- `gasPrice` gets removed and `max_priority_fee_per_gas, max_fee_per_gas` get added. I think we could hold off a little bit more on this but migrating is something we should look at. (My ledger wallet still doesn't seem to support these orders)

src/runners/http_rfqm_service_runner.ts
rhinodavid: if you really want to.. it's not reused so my preference is to keep it inline

src/utils/rfq_blockchain_utils.ts
```diff
@@ -209,6 +222,19 @@ export class RfqBlockchainUtils {
...
     }
 
+    /**
+     * Broadcasts a raw transaction via the `eth_sendRawTransaction` JSON RPC method.
+     * The transaction must be signed by this point, otherwise submission will fail.
+     *
+     * @returns The transaction hash returned by the RPC provider.
+     */
+    public async submitSignedTransactionAsync(signedTransaction: string): Promise<string> {
+        return this._web3Wrapper.sendRawPayloadAsync<string>({
```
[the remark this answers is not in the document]
rhinodavid: It gets caught and logged by `processRfqmJobAsync` just as it has before (tested this a BUNCH in ropsten lol)

src/services/rfqm_service.ts
```diff
@@ -1112,22 +1112,21 @@ export class RfqmService {
         gasEstimate: number,
     ): Promise<RfqmTransactionSubmissionEntity> {
         const txOptions = {
-            nonce,
+            from: workerAddress,
```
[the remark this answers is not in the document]
rhinodavid: `to` automatically gets set to the exchange proxy

src/utils/rfq_blockchain_utils.ts
```diff
@@ -231,4 +256,56 @@ export class RfqBlockchainUtils {
...
+     * If the `from` address is not present in the `TxData`, the function uses the address
+     * of the Ethers Wallet passed to `rfqBlockchainUtils`
+     *
+     * If the `to` address is not present, the function uses the exchange proxy address.
+     */
+    public async transformTxDataToTransactionRequestAsync(
+        txOptions: Partial<TxData>,
+        callData?: utils.BytesLike,
+    ): Promise<providers.TransactionRequest> {
+        let from = txOptions.from;
```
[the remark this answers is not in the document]
rhinodavid: for the worker we do but added this for flexibility

src/utils/rfq_blockchain_utils.ts
```diff
@@ -231,4 +248,58 @@ export class RfqBlockchainUtils {
...
+                throw new Error("An Ethers wallet must be provided if no 'from' address is specified");
+            }
+            from = await this._ethersWallet.getAddress();
+        }
+
+        return {
+            chainId,
+            data: callData,
+            from,
+            gasLimit: txOptions.gas instanceof BigNumber ? BigInt(txOptions.gas.toString()) : txOptions.gas,
```
phil-ociraptor, line 277: optional nit: can we add a comment calling out the renaming from `gas` to `gasLimit`?

test/rfq_blockchain_utils_test.ts
```diff
@@ -272,6 +279,38 @@ describe(SUITE_NAME, () => {
             expect(txHash).to.match(/^0x[0-9a-fA-F]+/);
         });
     });
```
phil-ociraptor, line 281: Can we add a test for `transformTxDataToTransactionRequestAsync`? Some logic in there that could do well having a test around:
- The fallback to address being the exchangeProxy
- The logic regarding mapping some of the fields
- The ethers fallback (not necessary to test this imo, but nice for completion - also see comment about possibly moving this logic out of this method)

src/utils/rfq_blockchain_utils.ts
```diff
@@ -43,6 +45,8 @@ export class RfqBlockchainUtils {
     private readonly _exchangeProxy: IZeroExContract;
     private readonly _web3Wrapper: Web3Wrapper;
     private readonly _abiDecoder: AbiDecoder;
+    // An ethers.js Wallet. Must be populated for RfqBlockchainUtils instances used by RFQM Workers.
+    private readonly _ethersWallet: Wallet | undefined;
```
phil-ociraptor, line 49: is this optional only because we're in the transition phase from web3wrapper to ethers? If so, can we drop the above comment and replace it with a TODO to make this non-optional once the migration is complete? Or just some message to remind us to remove the `undefined` union type at some point.

src/utils/rfq_blockchain_utils.ts
```diff
@@ -231,4 +248,58 @@ export class RfqBlockchainUtils {
...
+        chainId?: ChainId,
+        callData?: utils.BytesLike,
+    ): Promise<providers.TransactionRequest> {
+        let from = txOptions.from;
+        if (!from) {
+            if (!this._ethersWallet) {
+                throw new Error("An Ethers wallet must be provided if no 'from' address is specified");
+            }
+            from = await this._ethersWallet.getAddress();
+        }
```
phil-ociraptor, line 271: this logic was a little unexpected (was expecting a pure fn). Thoughts on replicating this functionality by offering a `getWalletAddress` fn on `RfqBlockchainUtils`? Then a caller can stitch together this logic as needed

src/runners/http_rfqm_service_runner.ts
phil-ociraptor: I prefer to move it into a constant, but will leave it up to you. (i.e. definitely not a blocker)

phil-ociraptor (approved):
Looks really good! Added some more minor comments but this looks good from my side, especially after tested on Ropsten

src/utils/rfq_blockchain_utils.ts
rhinodavid: Hmmmm.. it's only necessary if you're using `RfqBlockchainUtils` as a Worker (the API version doesn't need to sign/submit transactions). We could make it mandatory, but then we'd have to mock it for tests where it's not used. What do you think?

src/utils/rfq_blockchain_utils.ts
rhinodavid: Yeah Daniel didn't like it either.. maybe we just say it has to be provided

src/utils/rfq_blockchain_utils.ts
rhinodavid: done

src/utils/rfq_blockchain_utils.ts
rhinodavid: ok done

test/rfq_blockchain_utils_test.ts
rhinodavid: good call done

src/utils/rfq_blockchain_utils.ts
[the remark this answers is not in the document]
rhinodavid: removed

PR #724, opened 2021-08-24
feat: Publish Quote Reports to Kafka

Ktl-XV:
# Description

Allow Quote reports to be published to Kafka when including `KAFKA_BROKERS` is set. The Quote Reports will still be logged as usual.

# Testing Instructions

1. Setup https://github.com/0xProject/events_approval-parser
2. Configure `KAFKA_BROKERS` to point to the Kafka instance configured above
3. Querying the API should produce events (only for firm quotes)

# Checklist

    -   [ ] SRA/Limit orders
    -   [ ] Swap endpoints
    -   [ ] Depth charts


Ktl-XV:
deploy staging

Ktl-XV:
Waiting on https://github.com/0xProject/protocol/pull/361

Ktl-XV:
Ensure the Pipeleine orchestration Parser continues working for the current Quote Report

https://github.com/0xProject/0x-pipeline-orchestration/issues/390

src/handlers/swap_handlers.ts
```diff
@@ -233,6 +271,30 @@ export class SwapHandlers {
                 .getPriceComparisonFromQuote(CHAIN_ID, marketSide, quote)
                 ?.map((sc) => priceComparisonUtils.renameNative(sc));
         }
+
+        if (quote.extendedQuoteReport && kafkaProducer) {
+            const bytesPos = quote.data.indexOf('869584cd');
```
phil-ociraptor, line 276: hmm, what is this magic value? Can we add a comment regarding where this comes from?

src/handlers/swap_handlers.ts
```diff
@@ -233,6 +271,30 @@ export class SwapHandlers {
...
+            // tslint:disable-next-line
+            const quoteId = quote.data.slice(bytesPos + 118, bytesPos + 128);
```
phil-ociraptor, line 278: hmm, a few more magic numbers. Can we actually extract this logic into a utility function, maybe called `getQuoteIdFromSwapQuote`?
```
function getQuoteIdFromSwapQuote(quote) {
    const bytesPos = quote.data.indexOf('869584cd'); // some comment about where this comes from
    const quoteIdOffset = 118; // some comment about where this comes from
    const startingIndex = bytePos + quoteIdOffset;
    const quoteId = quote.data.slice(startingIndex, startingIndex + 10);
    return quoteId;
}

File: src/services/rfqm_service.ts [hunk 3]
```diff
@@ -469,18 +501,52 @@ export class RfqmService {
...
+        if (bestQuote) {
+            // Get the makerUri
+            const tmpMakerUri = this._quoteRequestor.getMakerUriForSignature(bestQuote.signature);
+            if (tmpMakerUri === undefined) {
+                throw new Error(`makerUri unknown for maker address ${bestQuote.order.maker}`);
+            } else {
+                makerUri = tmpMakerUri;
+            }
+            wrappedBestQuote = { order: bestQuote, makerUri };
+        }
```
phil-ociraptor, line 515: ahh, ok, logic here will actually change - I think it might be best to wait for the rebase. In the new version, you won't have to attach the makerUri, as it will already done for you. I can work with you to handle the merge

will make this code as simple as the indicative quote version where you just send the `firmQuotes` as `allQuotes` because they will already have the MakerUri attached. (Will also eliminate the need for `WrappedSignedNativeOrderMM`)

src/utils/quote_report_utils.ts
```diff
@@ -82,4 +149,164 @@ export const quoteReportUtils = {
             });
         });
     },
+    publishQuoteReport(logOpts: ExtendedQuoteReportLogOptions, isFirmQuote: boolean, kafkaProducer: Producer): void {
+        const qr: ExtendedQuoteReport = logOpts.quoteReport;
+        let logBase: { [key: string]: string | boolean | number | undefined } = {
```
phil-ociraptor, line 154: I think we can drop the types here and allow typescript to do type inference, i.e.:
```
const qr = logOpts.quoteReprt;
let logBase = {
    ....
}
```

phil-ociraptor (commented):
Looks good, some minor comments, but I would like to do this change on top of my changes that are currently in progress. If we can wait a week or so for deployment, it will make the code a lot easier to maintain!

src/services/rfqm_service.ts
Ktl-XV: Awesome, I've reverted changes related to RFQm and send them to https://github.com/0xProject/0x-rfq-api/pull/33.
It is much cleaner now

src/utils/quote_report_utils.ts
Ktl-XV: Sure

phil-ociraptor (approved):
Nice and simple - great stuff!

PR #733, opened 2021-09-08
fix: Added new API keys for custom integrations

PirosB3:
Adds new support for API keys that target 1 or more specific maker URIs. This allows market makers to test their implementations easily

### DIFF
--- a/src/config.ts
+++ b/src/config.ts
@@ -590,7 +590,13 @@ export const defaultHttpServiceWithRateLimiterConfig: HttpServiceConfig = {
     metaTxnRateLimiters: META_TXN_RATE_LIMITER_CONFIG,
 };
 
-export const INTEGRATOR_KEYED_BY_ID = transformIntegratorsAcl(INTEGRATORS_ACL, 'integratorId');
+export const getWhitelistedIntegratorUrlsForIntegratorId = (
+    (integratorsMap: Map<string, Integrator>) =>
+    (integratorId: string): string[] | undefined => {
+        const integrator = integratorsMap.get(integratorId);
+        return integrator?.whitelistIntegratorUrls;
+    }
+)(transformIntegratorsAcl(INTEGRATORS_ACL, 'integratorId'));
 
 /**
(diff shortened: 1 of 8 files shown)


src/config.ts
```diff
@@ -589,27 +590,45 @@ export const defaultHttpServiceWithRateLimiterConfig: HttpServiceConfig = {
     metaTxnRateLimiters: META_TXN_RATE_LIMITER_CONFIG,
 };
 
+export const INTEGRATOR_KEYED_BY_ID = transformIntegratorsAcl(INTEGRATORS_ACL, 'integratorId');
+
 /**
  * Gets the integrator ID for a given API key. If the API key is not in the configuration, returns `undefined`.
- * `integratorsMap` is closed in so it only needs to be evaluated once. Much efficiency!
  */
 export const getIntegratorIdForApiKey = (
```
rhinodavid, line 598: Ok I looked through this thing for a longgg time trying to make sense of it. I think the things that added the complexity for me were 1) the `byKey` addition to `transformIntegratorsAcl` and 2) the way ApiKey -> Integrator is exported as a function but IntegratorId -> Integrator is a map constant.

I think the `byKey` thing is worth keeping, but for the maps we should stick with 1 pattern. So my humble suggestion is to either:

1) (my preference) export a `getIntegratorForIntegratorId` function similar to `getIntegratorForApiKey`.
2) export a `INTEGRATOR_KEYED_BY_API_KEY` constant `Map`. 

#1 is my slight preference because for the caller of the function gets a nice signature and knows they just need to put in a string and get out `string | undefined`; the implementation using `Map` is hidden. With the constant the caller must know the constant is a `Map` and act accordingly. Ultimately either one works for me.

rhinodavid: Either way we should probably add a comment if the function/value is being exported.

rhinodavid (changes requested):

src/config.ts
PirosB3: Thanks for your feedback! I'm going to bias towards your suggestion! changes coming up right now!

rhinodavid (approved):

PR #738, opened 2021-09-17
fix: Add better market metrics

PirosB3:
Reduces the combinatorial complexity of Prometheus tags (that will reduce the number of tags created, therefore will also reduce the load)

### DIFF
--- a/src/utils/metrics_service.ts
+++ b/src/utils/metrics_service.ts
@@ -28,47 +28,30 @@ const ORDER_FILL_RATIO_WARNING_RANGE = new Counter({
 const RFQ_MAKER_NETWORK_INTERACTION_COUNTER = new Counter({
     name: 'rfq_maker_network_interaction_counter',
     help: 'Provides stats around market maker network interactions',
-    labelNames: [
-        'isLastLook',
-        'integratorLabel',
-        'url',
-        'quoteType',
-        'included',
-        'statusCode',
-        'market',
-    ],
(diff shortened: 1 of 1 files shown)


rhinodavid (approved):

phil-ociraptor (approved):

src/utils/metrics_service.ts
```diff
@@ -28,33 +28,46 @@ const ORDER_FILL_RATIO_WARNING_RANGE = new Counter({
...
 
+// NOTE: Do not use this map for anything sensitive. This is only used for
+// metrics/datavis purposes.
+const KNOWN_TOKENS: { [key: string]: string } = {
+    '0x6b175474e89094c44da98b954eedeac495271d0f': 'DAI',
+    '0xdac17f958d2ee523a2206206994597c13d831ec7': 'USDT',
+    '0xa0b86991c6218b36c1d19d4a2e9eb0ce3606eb48': 'USDC',
+    '0x2260fac5e5542a773aa44fbcfedf7c193bc2c599': 'WBTC',
+    '0xc02aaa39b223fe8d0a0e5c4f27ead9083c756cc2': 'WETH',
+};
```
kimpers, line 48: A bit late to the party but the `@0x/token-metadata` has token information for the most popular tokens on all the chains we support so you don't need to keep a separate list here :) See [token_metadata.ts](https://github.com/0xProject/token-metadata/blob/main/src/token_metadata.ts)

PR #749, opened 2021-10-01
feat: use eip1559 transactions for rfqm

z2trillion:
# Description

# Testing Instructions

# Checklist

    -   [ ] SRA/Limit orders
    -   [ ] Swap endpoints
    -   [ ] Depth charts

### DIFF
--- a/src/services/rfqm_service.ts
+++ b/src/services/rfqm_service.ts
@@ -121,10 +121,10 @@ interface SubmissionsMap {
     [transactionHash: string]: Partial<RfqmTransactionSubmissionEntity>;
 }
 
-type GasFees = {
+interface GasFees {
     maxFeePerGas: BigNumber;
     maxPriorityFeePerGas: BigNumber;
-};
+}
 
 interface SubmissionContext {
     submissionsMap: SubmissionsMap;
--- a/src/services/rfqm_service.ts
+++ b/src/services/rfqm_service.ts
@@ -1165,7 +1165,8 @@ export class RfqmService {
         const txOptions = {
             from: workerAddress,
             gas: gasEstimate,
-            gasPrice,
+            maxFeePerGas: gasPrice,
+            maxPriorityFeePerGas: new BigNumber(4),
             nonce,
             value: 0,
         };
@@ -1185,7 +1186,8 @@ export class RfqmService {
             createdAt: new Date(),
             from: workerAddress,
             to: this._blockchainUtils.getExchangeProxyAddress(),
-            gasPrice,
+            maxFeePerGas: gasPrice,
+            maxPriorityFeePerGas: new BigNumber(4),
             nonce,
             status: RfqmTransactionSubmissionStatus.Presubmit,
         };
--- a/src/services/rfqm_service.ts
+++ b/src/services/rfqm_service.ts
@@ -121,11 +121,16 @@ interface SubmissionsMap {
     [transactionHash: string]: Partial<RfqmTransactionSubmissionEntity>;
 }
 
+type GasFees = {
+    maxFeePerGas: BigNumber;
+    maxPriorityFeePerGas: BigNumber;
+};
+
 interface SubmissionContext {
     submissionsMap: SubmissionsMap;
     nonce: number;
     gasEstimate: number;
-    gasPrice: BigNumber;
+    gasFees: GasFees;
 }
 
 export type FetchFirmQuoteResponse = MetaTransactionRfqmQuoteResponse;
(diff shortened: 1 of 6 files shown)


src/services/rfqm_service.ts
```diff
@@ -1075,6 +1076,8 @@ export class RfqmService {
                             actualTakerFillAmount: decodedLog.args.takerTokenFilledAmount.toString(),
                             decodedFillLog: JSON.stringify(decodedLog),
                         };
+                        // submissionsMap[r.transactionHash].priorityFee = r.response.priorityFee;
```
z2trillion, line 1079: this is commented about because the receipt doesn't include the effective gas price. 

you should just delete these.

src/services/rfqm_service.ts
```diff
@@ -1103,6 +1106,7 @@ export class RfqmService {
                     };
                 }
             }
+            // here is where we need to do it....
```
z2trillion, line 1109: delete this.

phil-ociraptor:
Notes:
 - might want to make maxPriorityFeePerGas configurable (like a constant)
 - transaction receipt - doesn't have the effective gas price (but can look at the block to get the base fee, and do the calculation based off all of this)
 - bump `maxFeePerGas` and `maxPriorityFeePerGas` by 10% on every retry

phil-ociraptor:
look into the `any` vs `all` issue

migrations/1632854586018-AddEip1559Columns.ts
```diff
@@ -0,0 +1,20 @@
+import { MigrationInterface, QueryRunner, TableColumn } from 'typeorm';
+
+const columns = ['priority_fee', 'max_priority_fee', 'max_gas_price'].map(
```
phil-ociraptor, line 3: note - can remove `priority_fee`

src/services/rfqm_service.ts
```diff
@@ -932,27 +940,33 @@ export class RfqmService {
 
             switch (jobStatus) {
                 case RfqmJobStatus.PendingSubmitted:
-                    const newGasPrice = await this._protocolFeeUtils.getGasPriceEstimationOrThrowAsync();
+                    const gasPriceEstimate = await this._protocolFeeUtils.getGasPriceEstimationOrThrowAsync();
```
rhinodavid, line 943: I'm pretty confused by this and by the overall strategy we're using here. Isn't this returning the Type 0 "fast gas" (i.e. the Base Fee + a buffer)?

I'm thinking we should use a strategy similar to the one outlined [here](https://www.blocknative.com/blog/eip-1559-fees): Submit the tx with a `maxFeePerGas` equal to 1 or 1.5 gwei `maxPriortyFeePerGas` + 2 * the base fee, then increase the `maxPriorityFeePerGas` if necessary. Alternatively, I thought we had a new library which can give us real-time estimates for the gas price.

src/services/rfqm_service.ts
```diff
@@ -209,14 +214,17 @@ const RFQM_PROCESS_JOB_LATENCY = new Summary({
 });
 const PRICE_DECIMAL_PLACES = 6;
 
+const INITIAL_MAX_PRIORITY_FEE_PER_GAS = new BigNumber(4);
```
rhinodavid, line 217: It seems like this might be in gwei but ends up being used as if it's wei.

rhinodavid: Also seems really high (this screenshot is from a fairly high gas environment)
Screen Shot 2021-11-03 at 4 16 23 PM

src/services/rfqm_service.ts
phil-ociraptor: Ah, this is another thing that Mason and I discussed. I think this strategy will be "no worse" than what we currently have, with the upside that it caps our payment to the fast gas price

The issue if we give a large `maxFeePerGas` is that it will result in fairly different submission logic (i.e. on EIP-1559 chains, we can submit and wait and only update the `maxPriorityFeePerGas`, but other chains we do the replacing strategy as before)

If I'm not mistaken, I think this strategy allows us to have very similar submission logic for both types of chains

src/services/rfqm_service.ts
```diff
@@ -1180,12 +1194,12 @@ export class RfqmService {
         );
 
         const partialEntity: Partial<RfqmTransactionSubmissionEntity> = {
+            ...gasFees,
             transactionHash,
             orderHash,
             createdAt: new Date(),
             from: workerAddress,
             to: this._blockchainUtils.getExchangeProxyAddress(),
-            gasPrice,
```
phil-ociraptor, line 1188: need to support chains that don't support EIP-1559 as well

test/rfq_blockchain_utils_test.ts
```diff
@@ -312,7 +314,8 @@ describe(SUITE_NAME, () => {
             expect(result.to).to.equal(zeroEx.address);
         });
     });
-    describe('signTransactionRequestAsync', () => {
+    // Unskip this test if/when ganache 7.0.0, with EIP1559 support, is released.
+    describe.skip('signTransactionRequestAsync', () => {
```
phil-ociraptor, line 318: this might actually be a good test to see if our code can handle both EIP-1559 chains and legacy chains. In other words, we treat ganache as a legacy chain (until we update to 7.0.0)

phil-ociraptor (commented):
Primary comment is that we should make support of EIP-1559 optional, so depending on some ChainID mapping, we either use EIP-1559 style gas logic, or legacy logic

PR #754, opened 2021-10-06
fix: Fix logging for rfqm worker exit

rhinodavid:
# Description
My informed suspicion is that the RFQM worker processes are encountering unhandled errors and exiting. Currently, there is a log statement before the exit, but no logs exists for this callsite. There is a special log function provided by the `pino` library which blocks the exit until the log has been written.

Update the RFQM worker routine to use the blocking log function before exiting the process.

# Testing Instructions

# Checklist

    -   [ ] SRA/Limit orders
    -   [ ] Swap endpoints
    -   [ ] Depth charts

### DIFF
--- a/src/runners/rfqm_worker_runner.ts
+++ b/src/runners/rfqm_worker_runner.ts
@@ -1,7 +1,7 @@
 /**
  * Runs the RFQM MetaTransaction Consumer
  */
-import { createMetricsRouter, MetricsService } from '@0x/api-utils';
+import { createMetricsRouter, MetricsService, pino } from '@0x/api-utils';
 import { SQS } from 'aws-sdk';
 import * as express from 'express';
 import { Counter, Summary } from 'prom-client';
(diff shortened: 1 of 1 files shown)


src/runners/rfqm_worker_runner.ts
```diff
@@ -47,17 +47,24 @@ const RFQM_JOB_COMPLETED_WITH_ERROR = new Counter({
...
+        process.exit(1);
+    }),
+);
+
+process.on('unhandledRejection', (reason, promise) => {
+    const finalLogger = pino.final(logger);
+    finalLogger.error(
+        { reason, promise, workerIndex: RFQM_WORKER_INDEX },
+        'RFQM worker exiting due to unhandled rejection',
+    );
```
phil-ociraptor, line 64: hmm, why does the syntax differ between this callback and the one above?

phil-ociraptor (approved):
One small comment, but overall lgtm

src/runners/rfqm_worker_runner.ts
rhinodavid: OOOOooo are you ready to learn some node?

So these two events actually have different signatures. For `unhandledRejection` we were labeling the first parameter `err`but it's NOT actually error, it's `reason`. This case also passes in the promise for which the rejection was not handled. See https://nodejs.org/api/process.html#process_event_unhandledrejection.

For the `uncaughtException` it's more like what you'd expect: https://nodejs.org/api/process.html#process_event_uncaughtexception

PR #755, opened 2021-10-07
feat: Add prometheus gauge for total rfqm trade capacity

rhinodavid:
# Description

Lil' PR to create a prom gauge which shows how many trades the RFQM system can do as a whole.

# Testing Instructions

# Checklist

    -   [ ] SRA/Limit orders
    -   [ ] Swap endpoints
    -   [ ] Depth charts

### DIFF
--- a/src/utils/rfqm_health_check.ts
+++ b/src/utils/rfqm_health_check.ts
@@ -1,6 +1,6 @@
 import { RfqMakerAssetOfferings } from '@0x/asset-swapper';
 import { BigNumber } from '@0x/utils';
-import { Counter } from 'prom-client';
+import { Counter, Gauge } from 'prom-client';
 import { Producer } from 'sqs-producer';
 
 import { ETH_DECIMALS } from '../constants';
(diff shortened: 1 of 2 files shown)


src/utils/rfqm_health_check.ts
```diff
@@ -97,6 +103,12 @@ export async function computeHealthCheckAsync(
         RFQM_HEALTH_CHECK_ISSUE_COUNTER.labels(issue.status, issue.label).inc(),
     );
 
+    if (gasPrice) {
+        const totalWorkerBalance = heartbeats.reduce((total, { balance }) => total.plus(balance), new BigNumber(0));
+        const totalSystemTradeCapacity = totalWorkerBalance.div(gasPrice.times(RFQM_TX_GAS_ESTIMATE));
```
phil-ociraptor, line 108: so technically, we need to:

- take the floor of the amount of tx a worker can do
- sum the floors

Otherwise, we might get into a scenario where each worker has like the ability to do 0.99 trades, and we say "this is enough for N - 1 trades", when in fact we have 0

phil-ociraptor (approved):
Looks good, added a small comment when this metric might be unreliable

PR #779, opened 2021-11-01
fix: simplify logic for resubmitting tx's with higher gas prices

z2trillion:


rhinodavid (approved):
Thanks for tackling this -- we def needed it.

PR #780, opened 2021-11-03
feat: add migration and update entities for EIP1559

z2trillion:


migrations/1632854586018-AddEip1559Columns.ts
```diff
@@ -0,0 +1,20 @@
+import { MigrationInterface, QueryRunner, TableColumn } from 'typeorm';
+
+const columns = ['max_fee_per_gas', 'max_priority_fee_per_gas'].map(
+    (name) =>
+        new TableColumn({
+            name,
+            type: 'bigint',
```
rhinodavid, line 7: I think this is 'numeric'

rhinodavid (approved):

PR #807, opened 2022-01-12
fix: remove unnecessary logs

tobernguyen:
# Description

Remove the line of code that logs out info about an InternalServerError. This log is not necessary as we already re-throw the error and information about the error (and stack trace) will be printed out by the errorHandler middleware here: https://github.com/0xProject/api-utils/blob/69dd016bc3e1a97b27b2a19936603334c71562c3/src/middleware/error_handler.ts#L107

# Testing Instructions

- Checkout `master`. Provide an invalid RPC URL to `.env`. Build and run locally with `NODE_ENV=production yarn start:service:swap_http`
- Go to http://localhost:3000/swap/v1/quote?buyToken=DAI&sellToken=ETH&sellAmount=300000000000000000000
- The app will print out 3 lines of logs with 2 duplicated logs about the error (one at info level which was printed by line I'm removing, one at error level which is from api-utils).

- Now, checkout this branch and build and run the above command again
- Go to the above URL again
- The app will print only 3 line logs (the duplicated log line is removed)

Note: You might see even 5 lines of log (3 of them are for the error). This PR https://github.com/0xProject/api-utils/pull/4 will fix that but it is separated.

# Checklist

    -   [ ] SRA/Limit orders
    -   [ ] Swap endpoints
    -   [ ] Depth charts

### DIFF
--- a/src/handlers/swap_handlers.ts
+++ b/src/handlers/swap_handlers.ts
@@ -377,7 +377,6 @@ export class SwapHandlers {
                     },
                 ]);
             }
-            req.log.info('Uncaught error', e.message, e.stack);
             throw new InternalServerError(e.message);
         }
     }


src/handlers/swap_handlers.ts
```diff
@@ -377,7 +377,6 @@ export class SwapHandlers {
                     },
                 ]);
             }
-            req.log.info('Uncaught error', e.message, e.stack);
```
tobernguyen, line 380: This is not necessary as we already re-throw the error and information about the error (and stack trace) will be printed out by the errorHandler middleware here: https://github.com/0xProject/api-utils/blob/69dd016bc3e1a97b27b2a19936603334c71562c3/src/middleware/error_handler.ts#L107

dekz (approved):

PR #818, opened 2022-01-25
chore: Estimate the price impact from fee source data

dekz:
# Description

Uses the fee sources (note: we sell and buy 0.1 ETH to UniswapV2|V3|SushiSwap and take the median) data to estimate if there is a large price impact from the quoted price.

This data is not perfect as it is not exhaustive (we don't check EVERY liquidity source) and does have some problems (median value could be worse than the routed value). It **may** prevent someone from absolutely getting nuked if we can detect it.

# Testing Instructions

# Checklist

    -   [ ] SRA/Limit orders
    -   [ ] Swap endpoints
    -   [ ] Depth charts

### DIFF
--- a/src/services/swap_service.ts
+++ b/src/services/swap_service.ts
@@ -18,7 +18,10 @@ import {
     SwapQuoteRequestOpts,
     SwapQuoterOpts,
 } from '@0x/asset-swapper';
-import { NATIVE_FEE_TOKEN_BY_CHAIN_ID } from '@0x/asset-swapper/lib/src/utils/market_operation_utils/constants';
+import {
+    NATIVE_FEE_TOKEN_BY_CHAIN_ID,
+    ZERO_AMOUNT,
+} from '@0x/asset-swapper/lib/src/utils/market_operation_utils/constants';
 import { ChainId } from '@0x/contract-addresses';
 import { WETH9Contract } from '@0x/contract-wrappers';
 import { ETH_TOKEN_ADDRESS, RevertError } from '@0x/protocol-utils';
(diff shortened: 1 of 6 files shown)


dekz:
deploy fantom production

dekz:
deploy fantom production

src/services/swap_service.ts
```diff
@@ -90,6 +93,53 @@ export class SwapService {
...
+        sellTokenToEthRate: BigNumber,
+        buyTokenToEthRate: BigNumber,
+        marketSide: MarketOperation,
+    ): BigNumber {
+        // ETH to USDC
+        // price: "2418.92"
+        // sellTokenToEthRate: "1"
+        // buyTokenToEthRate: "2438.74"
+
+        // If sell than price is in taker token, if buy price is inverted
```
kimpers, line 117: NIT:
```suggestion
        // If sell then price is in taker token, if buy price is inverted
```

src/services/swap_service.ts
```diff
@@ -90,6 +93,53 @@ export class SwapService {
...
+        const normalizedPrice = marketSide === MarketOperation.Sell ? price : new BigNumber(1).dividedBy(price);
+        // 2418.92 / (2438.74/1) = 0.99187 or 99.187%
+        const actualPriceToEstimatedPriceRatio = normalizedPrice.dividedBy(
+            buyTokenToEthRate.dividedBy(sellTokenToEthRate),
+        );
+        // 0.99187 -> 0.812%
+        const estimatedPriceImpactPercentage = new BigNumber(1)
+            .minus(actualPriceToEstimatedPriceRatio)
+            // tslint:disable:custom-no-magic-numbers
+            .times(100)
```
kimpers, line 127: NIT: While I personally prefer this percentage representation as it's hard to get it wrong. With this, we would have params like `slippagePercentage`, `buyTokenPercentageFee` that are 0 - 1, and this percentage response will be 0 - 100. Maybe we want to be consistent or name this slightly differently?

kimpers (approved):
This is great! 🙏  We wanted something like this for Matcha forever ago 😄

src/types.ts
```diff
@@ -208,6 +208,9 @@ interface QuoteBase {
     protocolFee: BigNumber;
     minimumProtocolFee: BigNumber;
     allowanceTarget?: string;
+    // Our calculated price impact or undefined if we were unable to
+    // to calculate any price impact
+    estimatedPriceImpactPercentage: BigNumber | undefined;
```
kimpers, line 213: NIT(just opinion): I know we already do this, but using `undefined` to represent no value in API responses is kind of a pet peeve o mine. The key will disappear when the JSON is serialized and it can be confusing to the consumer of the API, I personally prefer using `null` values.

dekz: oh yeah I agree that it disappearing is actually extremely annoying

dekz:
deploy fantom production

dekz:
deploy ethereum production

dekz:
deploy production

PR #804, opened 2022-01-07
fix: get better gas estimates on /price

phil-ociraptor:
# Description

Presently, we hardcode `skipValidation: true`  in every `/price` request. This means we do not have accurate gas estimates on /price. As we will be using the gasEstimate to compare the all-in price of gasless swaps vs traditional swaps, it's important we get the gas estimate as close to correct as possible.

After speaking w/ @dekz - the reason we do not produce high fidelity gas estimates on `/price` is that we cannot call `eth_estimateGas` when RFQ indicative quotes are present. Thus, a large percentage of `/price` requests would fail when we estimate gas.

However, when RFQ is not present, we _do_ have the ability to estimate.

This change removes the hardcoding of skipValidation on price, and uses the sources to determine if we can estimate gas. It should do the following:
- Firm quotes should always be able to estimate gas
- Indicative quotes without Native liquidity should always be able to estimate gas
- Indicative quotes with Native liquidity should NOT be able to estimate gas

Of course, it will also skip validation if `skipValidation=true` is passed in from the API.

# Testing Instructions

We can test this on staging, and try to trigger the above scenarios:
- Firm quote
- Indicative quote without RFQ
- Indicative quote with RFQ (by disabling all other sources)

# Checklist

-   [ ] Update [documentation](https://github.com/0xProject/website/blob/development/mdx/api/index.mdx) as needed. **Website Documentation PR:**
-   [ ] Prefix PR title with `[WIP]` if necessary.
-   [ ] Add tests to cover changes as needed.
-   [ ] Test changes on the staging environment with [Matcha API Staging](https://api-staging.matcha.xyz).

    -   [ ] SRA/Limit orders
    -   [ ] Swap endpoints
    -   [ ] Meta transaction endpoints
    -   [ ] Depth charts

    For more information see `0x API Matcha smoke test runbook` in Quip.

PR #820, opened 2022-01-26
fix: /price endpoint does not fail on eth_gasEstimate error

phil-ociraptor:
# Description

Previously, in #804 - a change was introduced to call `eth_estimateGas` when we had the appropriate calldata to do so. However, this change did not take into account that sometimes this rpc call could fail for balance and allowance reasons. 

If the taker did not have enough balance, they would get an error that looks like this:

```
{
  "code": 105,
  "reason": "Error",
  "values": {
    "message": "ERC20: transfer amount exceeds allowance"
  }
}
```

This effectively broke any integrations that were relying on the `/price` to always work (even if the taker could not fill the order)

This change provides error handling around such a scenario.

However, we want to preserve this behavior on the `/quote` endpoint, and hence, bubble the error up if seen on a firm quote

# Testing Instructions

Added a new test for `/price` endpoint in `swap_test.ts` to automate the checking of this case. Previous `/quote` endpoints also give as some coverage.

# Checklist

    -   [ ] SRA/Limit orders
    -   [ ] Swap endpoints
    -   [ ] Depth charts

### DIFF
--- a/src/services/swap_service.ts
+++ b/src/services/swap_service.ts
@@ -332,21 +332,28 @@ export class SwapService {
         // using eth_gasEstimate
         // If an error occurs we attempt to provide a better message then "Transaction Reverted"
         if (takerAddress && !skipValidation && canEstimateGas) {
-            let estimateGasCallResult = await this._estimateGasOrThrowRevertErrorAsync({
-                to,
-                data,
-                from: takerAddress,
-                value,
-                gasPrice,
-            });
-            // Add any underterministic gas overhead the encoded transaction has detected
-            estimateGasCallResult = estimateGasCallResult.plus(gasOverhead);
-            // Take the max of the faux estimate or the real estimate
-            conservativeBestCaseGasEstimate = BigNumber.max(
-                // Add a little buffer to eth_estimateGas as it is not always correct
-                estimateGasCallResult.times(GAS_LIMIT_BUFFER_MULTIPLIER).integerValue(),
-                conservativeBestCaseGasEstimate,
-            );
+            try {
+                let estimateGasCallResult = await this._estimateGasOrThrowRevertErrorAsync({
+                    to,
+                    data,
+                    from: takerAddress,
+                    value,
+                    gasPrice,
+                });
+                // Add any underterministic gas overhead the encoded transaction has detected
+                estimateGasCallResult = estimateGasCallResult.plus(gasOverhead);
+                // Take the max of the faux estimate or the real estimate
+                conservativeBestCaseGasEstimate = BigNumber.max(
+                    // Add a little buffer to eth_estimateGas as it is not always correct
+                    estimateGasCallResult.times(GAS_LIMIT_BUFFER_MULTIPLIER).integerValue(),
+                    conservativeBestCaseGasEstimate,
+                );
+            } catch (error) {
+                logger.warn(
+                    { takerAddress, data, value, gasPrice, error: error?.message },
+                    'Unable to use eth_estimateGas. Falling back to faux estimate',
+                );
+            }
         }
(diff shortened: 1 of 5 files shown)


src/services/swap_service.ts
```diff
@@ -332,21 +332,32 @@ export class SwapService {
...
+                );
+            } catch (error) {
+                if (isFirmQuote) {
+                    // On firm quotes, when skipValidation=false, we want to raise an error
+                    throw error;
+                }
+                logger.warn(
+                    { takerAddress, data, value, gasPrice, error: error?.message },
+                    'Unable to use eth_estimateGas. Falling back to faux estimate',
+                );
```
kimpers, line 359: This is great, but I'm wondering if we are getting some different unintended consequences. For `/price` it's great that we can fall back to the gas schedule if we can't use `eth_estimateGas`. 

But I'm fairly sure this would also affect `/quote`, which means that we may give inaccurate gas estimations even tho the taker address was supplied, or worse we give out quotes that are guaranteed to revert due to missing balance/allowance etc.

phil-ociraptor: Ahh - for quote, we will rethrow the error, so the behavior should be the same as before. Our tests actually caught this and got me to add this above:

```
                if (isFirmQuote) {
                    // On firm quotes, when skipValidation=false, we want to raise an error
                    throw error;
                }
```

phil-ociraptor: Currently adding a test around `/price` so we'll have some certainty that both `/price` and `/quote` will work as intended

kimpers: Correct me if I'm wrong, but I think with the current logic it would not throw for `/quote` requests that do not use RFQ, or am I missing something?

phil-ociraptor: Ahh, I think you were right

I did just update the code to fix some tests and in doing so, I think it also happens to fix the case you were asking about (validation on `/quote` when no api key is present)

I think the confusion here was around what the boolean `isFirmQuote` actually represented. I've made it much simpler by just using the `endpoint` to distinguish if we're looking at a quote or a price. Let me know if the new version checks out to you!

kimpers: I think it's correctly throwing for quotes now but with `const canEstimateGas = isQuote || !isNativeIncluded;` aren't you back to skipping gas estimations for all `/price` requests? 😅

kimpers (approved):
I think it's all good but we should probably manually test this a bit on staging before 🚢 !

phil-ociraptor:
deploy staging

phil-ociraptor:
Following cases tested on staging:
- `/price` without skipValidation=false returns faux gas estimate
- `/price` with skipValidation=true and takerAddress present returns true gas estimate
- `/price` with takerAddress that does not have sufficient approval/token falls back to faux gas estimate (does not fail to 500). Triggers this log:
```
{
  "level": "warn",
  "error": "StringRevertError({ message: 'ERC20: transfer amount exceeds allowance' })",
  "msg": "Unable to use eth_estimateGas. Falling back to faux estimate"
}
```

- `/quote` is inaffected

phil-ociraptor:
deploy staging

phil-ociraptor:
Tested again on staging after changes - looks good. Will merge and deploy

phil-ociraptor:
deploy polygon production

phil-ociraptor:
deploy polygon production

phil-ociraptor:
deploy bsc production

phil-ociraptor:
deploy fantom production

phil-ociraptor:
deploy optimism production

phil-ociraptor:
deploy celo production

phil-ociraptor:
deploy production

PR #835, opened 2022-02-17
fix: Fixes To Enable Project To Run With Defaults

crypto-date:
# Description
At present, the project does not run if you follow the instructions in the readme, as reported by users in a number of issues. This is because 1) the url http://ganache is used instead of http://localhost and 2) there is an error in the null checking of the apiKey in config.ts.

This PR fixes these problems.

# Testing Instructions

With this branch, you can follow the instructions in the readme and you can actually run **yarn dev** successfully. This is not the case with the master branch.


samisbakedham:
I get "error role:api does not exist"

samisbakedham:
Image 2-16-22 at 10 53 PM

kimpers:
Thanks for fixing this! I'll make sure it gets merged

kimpers:
> I get "error role:api does not exist"

Looks like your DB does not have a user named `api` if you have any other postgres instances running, make sure you close them down before running the tests, also make sure you have properly migrated your local DB inside the docker compose

Walodja1987:
Same issue here. It throws on `yarn db:migrate` when following the README. I implemented the changes proposed in this PR but it looks like I need to specify an API key in the `.env` file. Where do I get it from?

image

Walodja1987:
Looks like running `@0x/neon-router@0.3.1` before `yarn build` fixes the issue. Solution found here:

https://ethereum.stackexchange.com/questions/109359/api-key-isnt-uuid-compliant-error

kimpers:
@crypto-date would you mind updating the fix to filter out empty `apiKeys` instead of bypassing the validity check?

Suggested change
```patch
diff --git a/src/config.ts b/src/config.ts
index 4b55c16..1ce2113 100644
--- a/src/config.ts
+++ b/src/config.ts
@@ -821,9 +821,8 @@ function assertEnvVarType(name: string, value: any, expectedType: EnvVarType): a
             return parseUtils.parseJsonStringForMetaTransactionRateLimitConfigOrThrow(value);
         case EnvVarType.APIKeys:
             assert.isString(name, value);
-            const apiKeys = (value as string).split(',');
+            const apiKeys = (value as string).split(',').filter((key) => !!key.trim());
             apiKeys.forEach((apiKey) => {
-                if (!apiKey) return;
                 const isValidUUID = validateUUID(apiKey);
                 if (!isValidUUID) {
                     throw new Error(`API Key ${apiKey} isn't UUID compliant`);

```

crypto-date:
@kimpers updated as suggested

src/config.ts
```diff
@@ -821,8 +821,9 @@ function assertEnvVarType(name: string, value: any, expectedType: EnvVarType): a
             return parseUtils.parseJsonStringForMetaTransactionRateLimitConfigOrThrow(value);
         case EnvVarType.APIKeys:
             assert.isString(name, value);
-            const apiKeys = (value as string).split(',');
+            const apiKeys = (value as string).split(',').filter((key) => !!key.trim());
             apiKeys.forEach((apiKey) => {
+                if (!apiKey) return;
```
kimpers, line 826: This line can be removed when you filter out the empty keys

kimpers (changes requested):
Thanks for the update 🙏 , just a small thing then we're good to go!

src/config.ts
crypto-date: done!

PR #858, opened 2022-05-11
feat: add decodedUniqueId to quote report

byeongminP:
# Description

This PR adds `decodedUniqueId` field to the Kafka quote report, making it easier to correlate two quote reports (e.g. /swap quote report and RFQm quote report). The `timestamp` field of the quote report is also modified to be consistent with how we generate `decodedUniqueId`.

# Testing Instructions

# Checklist

    -   [ ] SRA/Limit orders
    -   [ ] Swap endpoints
    -   [ ] Depth charts

### DIFF
--- a/src/utils/quote_report_utils.ts
+++ b/src/utils/quote_report_utils.ts
@@ -76,7 +76,7 @@ const omitFillData = (source: QuoteReportEntry) => {
  * @returns timestamp used when the unique identifier is generated
  */
 const getTimestampFromUniqueId = (decodedUniqueId: string): number => {
-    return parseInt(decodedUniqueId.slice(11), 10);
+    return parseInt(decodedUniqueId.slice(decodedUniqueId.indexOf('-') + 1), 10);
 };
 
 export const quoteReportUtils = {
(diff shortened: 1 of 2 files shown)


src/utils/quote_report_utils.ts
```diff
@@ -123,7 +132,8 @@ export const quoteReportUtils = {
             const extendedQuoteReport: ExtendedQuoteReport = {
                 quoteId: logOpts.quoteId,
                 taker: logOpts.taker,
-                timestamp: Date.now(),
+                timestamp:
+                    logOpts.submissionBy === 'taker' ? getTimestampFromUniqueId(logOpts.decodedUniqueId) : Date.now(),
```
novaliu86, line 136: If `submissionBy` is not `taker`, a single `decodedUniqueId` will not be enough to refer to a Swap quote report, right?

novaliu86: Seems `submissionBy` is always `taker` when this function is called. However in type definition it could be `metaTxn`. @Ktl-XV If interface `ExtendedQuoteReportForMetaTxn` is never used, could we delete it?

Ktl-XV: Yeah, we can remove it

novaliu86: Oh only call to `/meta_transaction/v1` will result in MetaTxn type of quote report. As long as we are calling `/swap/v1` we should always have a decodedUniqueId.

novaliu86 (approved):

PR #868, opened 2022-06-01
fix: slippage model issues

novaliu86:
# Description

Fix a few issues of `SlippageModelManager` according to recent discussion:
* `expectedSlippage` shall be `null` for trades below $10k in volume
* `expectedSlippage` shall not exceed the max slippage threshold set by the end-user
* `expectedSlippage` shall be `null` for unsupported pairs
* `expectedSlippage` shall be `null` for unsupported sources
* for trades spanning across multiple sources, the overall `expectedSlippage` shall be `null` if `expectedSlippage` of any individual source is `null`.

Also according to Jacob's comment, remove derived fields `expectedBuyAmount` and `expectedSellAmount`.
# Testing Instructions

Unit test added to cover `SlippageModelManager`.

# Checklist

    -   [ ] SRA/Limit orders
    -   [ ] Swap endpoints
    -   [ ] Depth charts

### DIFF
--- a/src/services/swap_service.ts
+++ b/src/services/swap_service.ts
@@ -523,7 +523,7 @@ export class SwapService {
                 } else {
                     apiSwapQuote.expectedBuyAmount = apiSwapQuote.buyAmount.times(
                         apiSwapQuote.expectedSlippage.plus(1),
-                    );
+                    ).integerValue();
                 }
             } else {
                 if (apiSwapQuote.expectedSlippage === null) {
@@ -531,7 +531,7 @@ export class SwapService {
                 } else {
                     apiSwapQuote.expectedSellAmount = apiSwapQuote.sellAmount.times(
                         apiSwapQuote.expectedSlippage.times(-1).plus(1),
-                    );
+                    ).integerValue();
                 }
             }
         }
--- a/src/services/swap_service.ts
+++ b/src/services/swap_service.ts
@@ -521,17 +521,17 @@ export class SwapService {
                 if (apiSwapQuote.expectedSlippage === null) {
                     apiSwapQuote.expectedBuyAmount = null;
                 } else {
-                    apiSwapQuote.expectedBuyAmount = apiSwapQuote.buyAmount.times(
-                        apiSwapQuote.expectedSlippage.plus(1),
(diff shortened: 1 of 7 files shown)


src/services/swap_service.ts
```diff
@@ -82,6 +82,8 @@ import { serviceUtils } from '../utils/service_utils';
 import { SlippageModelManager } from '../utils/slippage_model_manager';
 import { utils } from '../utils/utils';
 
+const DEFAULT_SLIPPAGE_PERCENTAGE = 0.0005; // Default slippage rate if no `slippagePercentage` is specified by users
```
dekz, line 85: If no API param is passed in the default is [here](https://github.com/0xProject/0x-api/blob/master/src/constants.ts#L29). Where 1 represents 100% slippage.

It may already be defaulted to this value in the `handler`, can't recall precisely.

src/services/swap_service.ts
```diff
@@ -508,19 +510,29 @@ export class SwapService {
...
                 );
             } else {
-                apiSwapQuote.expectedSlippage = new BigNumber(0);
+                apiSwapQuote.expectedSlippage = null;
             }
 
             if (marketSide === MarketOperation.Sell) {
-                apiSwapQuote.expectedBuyAmount = apiSwapQuote.buyAmount.times(apiSwapQuote.expectedSlippage.plus(1));
+                if (apiSwapQuote.expectedSlippage === null) {
+                    apiSwapQuote.expectedBuyAmount = null;
```
dekz, line 522: TBH I think we want to discuss this response field. I'm not entire sure Matcha is going to use it, and if they are, they could do this calculation themselves.

If we decide to keep this I think let's break it out to a function as it's gotten long in the tooth.

dekz (approved):
I think let's get 100% buy in on those additional fields. If they're truly needed than LGTM.

src/services/swap_service.ts
novaliu86: I am fine to remove these fields. Matcha probably will use new fields introduced by SlippageProtect project. If they are not, I agree a single `expectedSlippage` is enough for them.

PR #872, opened 2022-06-08
Chore: add estimatedpriceimpact to quote report [TKR-440]

idokleinman:
# Description

add optional estimatedPriceImpact to quote report for data team improved slippage model

# Testing Instructions

# Checklist

    -   [ ] SRA/Limit orders
    -   [ ] Swap endpoints
    -   [ ] Depth charts

### DIFF
--- a/package.json
+++ b/package.json
@@ -136,7 +136,7 @@
     "dependencies": {
         "@0x/api-utils": "^0.0.2",
         "@0x/assert": "^3.0.34",
-        "@0x/asset-swapper": "^16.61.0",
+        "@0x/asset-swapper": "^16.61.1",
         "@0x/contract-addresses": "^6.15.0",
         "@0x/contract-wrappers": "^13.20.3",
         "@0x/contracts-zero-ex": "^0.34.0",
--- a/src/handlers/swap_handlers.ts
+++ b/src/handlers/swap_handlers.ts
@@ -181,6 +181,7 @@ export class SwapHandlers {
                         blockNumber: quote.blockNumber,
                         slippage: undefined,
                         estimatedGas: quote.estimatedGas,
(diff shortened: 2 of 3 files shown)


robpal1990:
Will this return the price impact of **each source** in the quote? It seems like the API calculates it in _calculateEstimatedPriceImpact ( swap_service.ts ) which does not go source by source, only aggregated number.

Overall, we'd like to get a number per each quote report entry, not just per header.
This probably requires more digging.

robpal1990:
> > 
> 
> Indeed this is just the aggregated number. The whole QuoteReport entry seems to be referencing to the entire quote vs specific sources. The only point where there's a per-source reference is in `quoteReportSources` but including priceImpact there is currently not how the code/data is structured. I see fillData for each source is being passed on - maybe that calculation can be done downstream on your side?

My assumption is you'd need to pull it from the internals of `asset-swapper` (which would source it from the sampler).
`fillData` is not enough I think, it's only the token  paths/amounts, not liquidity -- correct me if I'm wrong.

Tbh, I think the datapoint we need is currently not available and requires a refactor by the codebase owner.
However if we ever want to use it in production in the future as an input to a model, we will need to be able to serve this datapoint during the quoting process, so it will need a restructuring anyway.

zhehao-0x:
> > 
> 
> Indeed this is just the aggregated number. The whole QuoteReport entry seems to be referencing to the entire quote vs specific sources. The only point where there's a per-source reference is in `quoteReportSources` but including priceImpact there is currently not how the code/data is structured. I see fillData for each source is being passed on - maybe that calculation can be done downstream on your side?

Ido, do you know who the best person is to understand how long /hard it takes to get this information in? I guess it is Jacob and I am planning to create a more formal request, given slippage estimate is a high priority project to make more money

joshrrose:
I'm going to close this PR. The changes have been included into https://github.com/0xProject/0x-api/pull/1141 per request from product

PR #875, opened 2022-06-14
fix: Incomplete data for response time buckets

eitu5ami:
# Description

Swap requests use RPC gateway to access data from blockchain. We use non-default value for RPC timeout, at the moment of writing it's value is 15000 (15 seconds). Unfortunately, a call to `linearBucket` generates buckets up to 6 seconds and drops anything above 6 seconds. With incomplete data sets, Grafana does not present a full picture of service behavior and makes a user to think we hit upper limit. 

This change generates a list bucket by use of defined rpc request timeout. This way, we should always have correct data.

Screenshot 2022-06-14 at 10 50 42

# Testing Instructions

When shipped, we should be able to set data for buckets above 6 seconds on Grafana.


dekz (approved):
Ah yeah nice, we had that limit because we assumed it would entirely timeout above 5s in the past.

PR #877, opened 2022-06-15
bump AS version

idokleinman:


idokleinman:
deploy production polygon

PR #912, opened 2022-08-06
feat: bump packages dont exclude goerli mumbai

idokleinman:


kyu-c:
There is a compatibility issue due to node version used in CI (v12) vs what's required by `asset-swapper` (>=v14), but increasing node version caused another issue with `fast-abi`... 

`fast-abi` has been re-published and `asset-swapper` has been also re-published (with the updated `fast-abi) .

This is being addressed in https://github.com/0xProject/0x-api/pull/913

PR #990, opened 2022-09-27
fix: log ammQuoteUniqueId for metatransaction quote report

rhinodavid:
# Description

`decodedUniqueId` doesn't show up in the database. I thought this translated (somewhere) to ammQuoteUniqueId, but the mysteries of quote report are not that easily divined.

Send `ammQuoteUniqueId` instead.

# Testing Instructions

# Checklist

-   [ ] Update [documentation](https://github.com/0xProject/website/blob/development/mdx/api/index.mdx) as needed. **Website Documentation PR:**
-   [ ] Prefix PR title with `[WIP]` if necessary.
-   [ ] Add tests to cover changes as needed.
-   [ ] Test changes on the staging environment with [Matcha API Staging](https://api-staging.matcha.xyz).

    -   [ ] SRA/Limit orders
    -   [ ] Swap endpoints
    -   [ ] Meta transaction endpoints
    -   [ ] Depth charts

    For more information see `0x API Matcha smoke test runbook` in Quip.

PR #1001, opened 2022-10-04
feat: Set `ammQuoteUniqueId` (to `decodeUniqueId` of gasless swap RFQ quote report) for gasless swap AMM quote report

Haozhuo:
# Description

Related to https://github.com/0xProject/0x-rfq-api/pull/471. Similar changes as #990: 
* Set `ammQuoteUniqueId` (to `decodeUniqueId` of gasless swap RFQ quote report) for gasless swap AMM quote report. The `decodeUniqueId` of gasless swap RFQ quote report is sent as a query param when gasless swap service calls `meta-transaction`
* Set `decodeUniqueId` for gasless swap AMM quote report field to be `decodedUniqueId` of the generated quote (format: `${quoteId}-${timestamp}`)

# Testing Instructions

# Checklist

    -   [ ] SRA/Limit orders
    -   [ ] Swap endpoints
    -   [ ] Depth charts


src/utils/quote_report_utils.ts
```diff
@@ -83,6 +84,7 @@ export function publishQuoteReport(
                 logOpts.submissionBy === 'taker' || logOpts.submissionBy === 'gaslessSwapAmm'
                     ? logOpts.decodedUniqueId
                     : undefined,
+            ammQuoteUniqueId: logOpts.submissionBy === 'gaslessSwapAmm' ? logOpts.ammQuoteUniqueId : undefined,
```
phil-ociraptor, line 87: this term `gaslessSwapAmm` is becoming a bit of a magic value - can we pull it into a constant? Do you also understand what `taker` vs `gaslessSwapAmm` is - and if it makes sense to group them, an enum might be better to represent them

src/services/meta_transaction_service.ts
```diff
@@ -137,7 +137,8 @@ export class MetaTransactionService {
                     taker: params.takerAddress,
                     quoteReportSources: quote.extendedQuoteReportSources,
                     submissionBy: 'gaslessSwapAmm',
-                    decodedUniqueId: params.quoteUniqueId ? params.quoteUniqueId : quote.decodedUniqueId,
+                    decodedUniqueId: quote.decodedUniqueId,
+                    ammQuoteUniqueId: params.quoteUniqueId,
```
phil-ociraptor, line 141: don't love that there's a name change here. Can we just call this field `quoteUniqueId`?

Haozhuo: The name of the field can't be changed since this object would be produced to quote report Kafka topic.

Maybe we can change the query param name in `params` but I do think the current name in query param `quoteUniqueId` is accurate since it's the `decodedUniqueId` for corresponding gasless swap RFQ quote report

src/utils/quote_report_utils.ts
Haozhuo: The current way `submissionBy` field is defined: `submissionBy: 'taker' | 'gaslessSwapAmm' | 'rfqm'`

`gaslessSwapAmm` means this quote report is for gasless swap AMM quote. `taker` means `/swap`. For now, I personally think either way (the current way or enum) works fine. If we were to make changes on this, thinking about doing it in a separate PR as otherwise this PR would contain different set of changes

src/services/meta_transaction_service.ts
phil-ociraptor: Gotcha - understood. Can we leave a comment on why there's a name change here?

phil-ociraptor (approved):
looks good, just a request for a small comment

PR #1012, opened 2022-10-11
feat: Updated SDK package. Added new entities to sg query for new pool type.

johngrantuk:
## Description

* Updated Balancer SDK package:
    * New path algorithm for Boosted stable pools.
    * Adds support for new Balancer ComposableStable pools.
    * Adds support for [Gyroscope](https://www.gyro.finance) pools. 
    * Various bug fixes & improvements.
* Added new entities to Subgraph query. These are required for Gyroscrope pools. Also Polygon Subgraph supports linear pools so added these.

## Testing instructions

* There should be no breaking change, rather we're just adding support for new Balancer pools.
* There is a large matic/stMatic ComposableStable pool on [Polygon](https://polygon.balancer.fi/#/pool/0x8159462d255c1d24915cb51ec361f700174cd99400000000000000000000075d) that can be tested against. (Other aggregators aren't picking up this yet so should provide good volume)

## Types of changes

* New feature (non-breaking change which adds functionality)

## Checklist:

-   [ ] Update documentation as needed.
-   [ ] Add new entries to the relevant CHANGELOG.jsons.


kyu-c:
@johngrantuk looks like in the new version `BalancerSdkConfig.rpcUrl` is not optional and it results in a runtime error when `BalancerSDK` is initialized. 

Is there a workaround?

johngrantuk:
> 

It should still be possible to pass just pass an empty string as the rpcUrl. I don't see any issue when I run it like this myself. Is there anyway I could see more details about the error? Thanks.

kyu-c:
```
        const config: BalancerSdkConfig = {
            network: 137,
            rpcUrl: '',
        };
        const balancerSdk = new BalancerSDK(config);
```

```
     Error: invalid signer or provider (arg="signerOrProvider", value={"_isProvider":true,"_events":[],"_emitted":{"block":-2},"formatter":{"formats":{"transaction":{},"transactionRequest":{},"receiptLog":{},"receipt":{},"block":{},"blockWithTransactions":{},"filter":{},"filterLog":{}}},"anyNetwork":false,"_network":{"name":"matic","chainId":137,"ensAddress":null,"_defaultProvider":null},"_maxInternalBlockNumber":-1024,"_lastBlockNumber":-2,"_pollingInterval":4000,"_fastQueryDate":0,"connection":{"url":"https://polygon-rpc.com"},"_nextId":42}, version=4.0.49)
      at Object.throwError (node_modules/ethers/errors.js:76:17)
      at new Contract (node_modules/ethers/contract.js:281:20)
      at Function.connect (node_modules/@balancer-labs/typechain/dist/factories/Vault__factory.js:1192:16)
      at new fn (node_modules/@balancer-labs/sdk/src/modules/swaps/swaps.module.ts:50:41)
      at new exports.BalancerSDK (node_modules/@balancer-labs/sdk/src/modules/sdk.module.ts:44:18)
```

I guess in `swap.modules.ts`, it tires to `connect`. 

https://github.com/balancer-labs/balancer-sdk/blob/develop/balancer-js/src/modules/swaps/swaps.module.ts#L50-L54

johngrantuk:
Thanks @kyu-c. The issue was coming from a ethers version issue. There seems to be an existing dep for ethers 4.0.4 but the SDK uses a later version and this caused a problem with the provider. I tried updating to a newer version of ethers but this caused issue with your generated-wrappers. I wasn't sure of the best approach to fix the version issue so I updated the code to workaround by using the SOR directly instead of via the SDK. This seems to build and run ok when I run locally. Let me know how things look for you? Thanks!

kyu-c:
Hey @johngrantuk ! Thanks for looking into the issue 🙏 
I tested Mainnet, Arbitrum, and Polygon and they look good!

I was getting errors on Optimism however (I didn't get a chance to debug unfortunately). 
On Fantom, the new format (v2 batch?) won't work as the the current Fantom contract doesn't support it yet (we will need to deploy a new contract). 

To move forward, since we will need further debugging on Optimism and there is a contract dependency on Fantom, I suggest that we revert Beethoven X changes for now and merge Balancer changes first (if you prefer, we can also wait and address both Balancer and Beethoven X changes in one PR -- just let me know your preference!) 

Also would you rebase to `master` ? (`ChainId.Rinkeby` and `ChainId.Ropsten` aren't there any more at head).

johngrantuk:
> 

Thanks for testing @kyu-c! I've reverted the Beethoven changes for now, we'll debug Optimism asap and add that later. I also rebased as requested. Just let me know if you need anything else.

kyu-c:
Thanks @johngrantuk!

I just applied `prettier` to make CI happy.

PR #1027, opened 2022-10-18
chore: add slippage protection flag to quote reports [GRO-28]

patrick-dowell:
# Description

Data uses Quote Reports to compare settled trades versus the quotes.

We need to add a field to the quote reports to indicate if a quote was slippage protected. Quotes are slippage protected by default unless the enableSlippageProtection parameter is set to false. Being able to distinguish slippage-protected trades is important for analyzing the impact of the feature.

This change affects quote generated in swap_handlers.ts and meta_transaction_service.ts. Note that the meta transactions are not slippage protected, so the flag will always be false.

# Testing Instructions

Setup kafka locally and configure 0xAPI to point to mainnet and your local kafka. Run some quote requests against your local endpoint and verify the quote reports sent to kafka.

More information on kafka can be found at:
https://www.conduktor.io/kafka/starting-kafka

# Checklist

-   [x] Test changes on the staging environment with [Matcha API Staging](https://api-staging.matcha.xyz).

    -   [ ] SRA/Limit orders
    -   [ ] Swap endpoints
    -   [ ] Depth charts


src/asset-swapper/utils/quote_report_generator.ts
```diff
@@ -107,6 +107,7 @@ export interface ExtendedQuoteReport {
     sourcesDelivered: ExtendedQuoteReportIndexedEntryOutbound[] | undefined;
     blockNumber: number | undefined;
     estimatedGas: string;
+    enableSlippageProtection?: boolean;
```
kyu-c, line 110: [bit out of scope for this PR / optional] 
I think we can make `GetSwapQuoteParams.enableSlippageProtection` not optional (I think we only need to change `meta_transaction_service.ts`  in order to do that. 

Then we can make `ExtendedQuoteReport. enableSlippageProtection` not optional as well.

kyu-c (approved):

src/asset-swapper/utils/quote_report_generator.ts
patrick-dowell: I can create a separate task / PR for this change. It's small but it'll make the code a little cleaner.

PR #1072, opened 2022-11-16
chore: change InsufficientAssetLiquidity error message [TKR-662]

eobbad:
# Description

# Testing Instructions

# Checklist

    -   [ ] SRA/Limit orders
    -   [ ] Swap endpoints
    -   [ ] Depth charts


src/asset-swapper/types.ts
```diff
@@ -370,7 +370,7 @@ export enum SwapQuoteConsumerError {
 export enum SwapQuoterError {
     NoEtherTokenContractFound = 'NO_ETHER_TOKEN_CONTRACT_FOUND',
     StandardRelayerApiError = 'STANDARD_RELAYER_API_ERROR',
-    InsufficientAssetLiquidity = 'INSUFFICIENT_ASSET_LIQUIDITY',
+    InsufficientAssetLiquidity = 'We are not able to fulfill an order for this token pair at the requested amount due to a lack of liquidity',
```
phil-ociraptor, line 373: We should keep the enum value as is and potentially add a `reason` field on the response. 

Ultimately, what is shown to the end user is a decision that is best made by the application. We can add a `reason` to supplement the response, but is not ideal

phil-ociraptor (changes requested):

src/utils/schema_utils.ts
```diff
@@ -45,34 +55,38 @@ function schemaValidationErrorToValidationErrorItem(schemaValidationErrorObject:
...
+        code = ValidationErrorCodes.UnsupportedOption;
     } else {
         throw new Error(`Unknown schema validation error name: ${schemaValidationErrorObject.keyword}`);
     }
+
+    if (schemaValidationErrorObject.description === undefined) {
+        return {
+            field: field,
+            code: code,
+            reason: reason,
```
phil-ociraptor, line 81: thought prettier might do this automatically for you, but can use this shorthand syntax: 
```suggestion
            field,
            code,
            reason,
```

src/utils/schema_utils.ts
```diff
@@ -45,34 +55,38 @@ function schemaValidationErrorToValidationErrorItem(schemaValidationErrorObject:
...
+        };
+    }
+
+    return {
+        field: field,
+        code: code,
+        reason: reason,
```
phil-ociraptor, line 88: same comment
```suggestion
        field,
        code,
        reason,
```

src/utils/schema_utils.ts
```diff
@@ -45,34 +55,38 @@ function schemaValidationErrorToValidationErrorItem(schemaValidationErrorObject:
             'dependencies',
         ].includes(schemaValidationErrorObject.keyword)
     ) {
-        return {
-            field: schemaValidationErrorObject.dataPath.replace('.', ''),
-            code: ValidationErrorCodes.IncorrectFormat,
-            reason: schemaValidationErrorObject.message || '',
-        };
+        field = schemaValidationErrorObject.dataPath.replace('.', '');
+        code =  ValidationErrorCodes.IncorrectFormat;
```
phil-ociraptor, line 59: small nit, but I actually prefer each `if` block returning its own object. You can accomplish a very similar goal (where all of them share the same reason) by expressing it as such:
```
return {
    field: schemaValidationErrorObject.dataPath.replace('.', ''),
    code: ValidationErrorCodes.IncorrectFormat,
    reason,
}
```

The rationale (both extremely minor)
- No mutation (when you have `let` this makes it easier to do mutation in the future, which can result in hard to reason about code over time)
- All of the logic is localized in the `if` block. Right now, you have to scroll to the bottom of the fn to understand what the return value is. So you need to read the `if` block and scroll down

SavDont: We decided to write it this way because we need a check for whether the `description` field was present. If the field is not present we do not want to populate it in the returned object. If we want to add a return statement in each `if` block we'd need to add the check for `if(schemaValidationErrorObject.description === undefined)` in each `if` block.

Is there a way we can have both the return statement in each `if` block without having a repetitive check for error description in every `if` block?

phil-ociraptor: ah yeah, I didn't see that down there. To write it without an `if` statement, you can do the following:
```
//somewhere above
const description = schemaValidationErrorObject.description;

// later
return {
    field: schemaValidationErrorObject.dataPath.replace('.', ''),
    code: ValidationErrorCodes.IncorrectFormat,
    reason,
    ...(description && { description }) 
}
```

`description` can either be `undefined` or a string.

 - If undefined:  `...` operator on `undefined`, it will do nothing - which is what we want (i.e. we don't want `{description: undefined}`)
 - If a string, we are using the `...` operator on an object that looks like `{description: "my error description"}` which is what we want

phil-ociraptor (approved):
lifting my request for changes! the nits are just minor stylistic things - not blocking

PR #1075, opened 2022-11-18
feat: Price Impact Protection [LIT-642]

joshrrose:
# Description

added price impact protection param. defaulted to 20% but can be passed as parameter. if price impact would be worse then priceImpactProtection then we will return error from the API

# Testing Instructions

# Checklist

    -   [ ] SRA/Limit orders
    -   [ ] Swap endpoints
    -   [ ] Depth charts


src/constants.ts
```diff
@@ -20,6 +20,7 @@ export const QUOTE_ORDER_EXPIRATION_BUFFER_MS = ONE_SECOND_MS * 60; // Ignore or
 const GAS_LIMIT_BUFFER_PERCENTAGE = 0.1; // Add 10% to the estimated gas limit
 export const GAS_LIMIT_BUFFER_MULTIPLIER = GAS_LIMIT_BUFFER_PERCENTAGE + 1;
 export const DEFAULT_QUOTE_SLIPPAGE_PERCENTAGE = 0.01; // 1% Slippage
+export const DEFAULT_PRICE_IMPACT_PROTECTION_PERCENTAGE = 0.2; // 20%
```
kyu-c, line 23: [no action required] 
what's your thoughts on the unit here?
I have felt these these `_PERCENTAGE` constants are misleading.

src/handlers/swap_handlers.ts
```diff
@@ -449,6 +450,10 @@ const parseSwapQuoteRequestParams = (req: express.Request, endpoint: 'price' | '
             },
         ]);
     }
+    const priceImpactProtectionPercentage =
+        req.query.priceImpactProtectionPercentage === undefined
+            ? DEFAULT_PRICE_IMPACT_PROTECTION_PERCENTAGE
+            : Number.parseFloat(req.query.priceImpactProtectionPercentage as string);
```
kyu-c, line 456: do we expect percentage or decimal from the api user? 

for 20%, 20? or 0.2?

src/constants.ts
joshrrose: yeah its a little weird to say "_PERCENTAGE" and then expect .2 for 20%. I didn't think too much about it when writing it I just followed the convention. I'm up to change it if we don't like it but i'm not sure if `DEFAULT_PRICE_IMPACT_PRETOECTION` is more clear....

src/handlers/swap_handlers.ts
joshrrose: we are expected `0.2` from the user for 20%

kyu-c: I just realized that `slippagePercentage` also uses decimal 😞. I guess we are sort of stuck with this convention.

I think we should add a validation on the range if we already don't have one..
If a user wanted 20% and provide 20 by mistake then that could disable price impact protection altogether.

src/constants.ts
kyu-c: Looks like we are stuck with this convention because of slippage percentage... 
i guess it's better to be consistently "wrong" than having two different styles. 😆 
Let's leave it as it is.

src/handlers/meta_transaction_handlers.ts
```diff
@@ -199,6 +203,11 @@ function parseRequestParams(req: express.Request): MetaTransactionQuoteRequestPa
         ]);
     }
 
+    const priceImpactProtectionPercentage =
+        req.query.priceImpactProtectionPercentage === undefined
+            ? DEFAULT_PRICE_IMPACT_PROTECTION_PERCENTAGE
+            : Number.parseFloat(req.query.priceImpactProtectionPercentage as string);
```
idokleinman, line 209: With parseFloat here - Will the user need to supply ?p=0.315 or ?p=31.5 for setting prime impact protection to be higher than 31.5% ?

joshrrose: `?p=0.315` would be 31.5%. I think this is consistent with our other "percentage" input params

src/handlers/swap_handlers.ts
```diff
@@ -452,6 +458,22 @@ const parseSwapQuoteRequestParams = (req: express.Request, endpoint: 'price' | '
         ]);
     }
 
+    let priceImpactProtectionPercentage = DEFAULT_PRICE_IMPACT_PROTECTION_PERCENTAGE;
+    if (req.query.priceImpactProtectionPercentage !== undefined) {
+        PRICE_IMPACT_PROTECTION_SPECIFIED.inc();
+        priceImpactProtectionPercentage = Number.parseFloat(req.query.priceImpactProtectionPercentage as string);
+        if (priceImpactProtectionPercentage > 1) {
```
kyu-c, line 465: I guess it's unlikely mistake but should we also check lower bound?

src/handlers/swap_handlers.ts
```diff
@@ -452,6 +458,22 @@ const parseSwapQuoteRequestParams = (req: express.Request, endpoint: 'price' | '
         ]);
     }
 
+    let priceImpactProtectionPercentage = DEFAULT_PRICE_IMPACT_PROTECTION_PERCENTAGE;
+    if (req.query.priceImpactProtectionPercentage !== undefined) {
+        PRICE_IMPACT_PROTECTION_SPECIFIED.inc();
+        priceImpactProtectionPercentage = Number.parseFloat(req.query.priceImpactProtectionPercentage as string);
```
kyu-c, line 464: I think we need to add a check for `NaN`.

kyu-c (changes requested):

src/handlers/swap_handlers.ts
joshrrose: I did consider this. I checked in postman when you submit a negative number some regex check catches it.

src/handlers/swap_handlers.ts
joshrrose: we don't seem to check for NaN in slippagePercentage either...

kyu-c: whether the return value of ` Number.parseFloat` is `NaN`

joshrrose: actually looks like the same regex that catches negative numbers also checks to make sure its some kind of number. i attempted to test the requested changes and it was caught earlier by the regex

kyu-c: oh i see -- makes sense

It would have been great if we have a single validator (vs two step process -- error description isn't very great when it fails regex-based validation)

kyu-c (approved):

idokleinman (approved):

src/services/swap_service.ts
```diff
@@ -71,6 +78,11 @@ import { SlippageModelFillAdjustor } from '../utils/slippage_model_fill_adjustor
 import { SlippageModelManager } from '../utils/slippage_model_manager';
 import { utils } from '../utils/utils';
 
+const PRICE_IMPACT_TOO_HIGH = new Counter({
```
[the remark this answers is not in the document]
joshrrose: Yeah i can make this change. Is this mostly to support if we were going to have different reasons in the future?

PR #1105, opened 2022-11-30
chore: removing duplicate gas estimation logic [LIT-534]

SavDont:
# Description

# Testing & Simbot Run

# Checklist


src/asset-swapper/utils/protocol_fee_utils.ts
```diff
@@ -1,24 +1,24 @@
...
 import { SwapQuoterError } from '../types';
 
 const MAX_ERROR_COUNT = 5;
 
-interface GasOracleResponse {
-    result: {
-        // gas price in wei
-        fast: number;
-    };
+interface GasPrices {
```
kyu-c, line 8: Can you document the unit (`wei`) as a comment? (it's been deleted)

src/asset-swapper/utils/protocol_fee_utils.ts
```diff
@@ -66,28 +67,26 @@ export class ProtocolFeeUtils {
         this._initializeHeartBeat();
     }
 
-    private async _getGasPriceFromGasStationOrThrowAsync(): Promise<BigNumber> {
+    private async _getGasPriceFromGasStationOrThrowAsync(): Promise<void> {
```
kyu-c, line 70: it looks like a more proper name would be something like `updateGasPriceFromOracleOrThrow`

src/asset-swapper/utils/protocol_fee_utils.ts
```diff
@@ -1,24 +1,24 @@
...
-    };
+interface GasPrices {
+    fast: number;
+    l1CalldataPricePerUnit?: number;
+}
+interface GasInfoResponse {
+    result: GasPrices;
 }
 
 export class ProtocolFeeUtils {
```
kyu-c, line 16: optional: perhaps rename this as it's not really about protocol fee.

src/asset-swapper/utils/protocol_fee_utils.ts
```diff
@@ -42,15 +42,16 @@ export class ProtocolFeeUtils {
...
     /** @returns gas price (in wei) */
-    public async getGasPriceEstimationOrThrowAsync(shouldHardRefresh?: boolean): Promise<BigNumber> {
-        if (this._gasPriceEstimation.eq(constants.ZERO_AMOUNT)) {
-            return this._getGasPriceFromGasStationOrThrowAsync();
-        }
-        if (shouldHardRefresh) {
-            return this._getGasPriceFromGasStationOrThrowAsync();
-        } else {
-            return this._gasPriceEstimation;
+    public async getGasPriceEstimationOrThrowAsync(defaultGasPrices: GasPrices): Promise<GasPrices> {
```
kyu-c, line 45: As it is it can fail even when `defaultGasPrices` is provided which seems weird. 

How about providing two different methods:
`getGasPriceEstimationOrThrowAsync()` -- no default
`getGasPriceEstimationOrDefault()` -- fallback to provided default.

src/asset-swapper/utils/protocol_fee_utils.ts
```diff
@@ -42,15 +42,16 @@ export class ProtocolFeeUtils {
...
+        await this._getGasPriceFromGasStationOrThrowAsync();
```
kyu-c, line 46: wouldn't this mean we will always hard refresh? (no caching at all).

src/asset-swapper/swap_quoter.ts
```diff
@@ -146,7 +146,11 @@ export class SwapQuoter {
      * Returns the recommended gas price for a fast transaction
      */
     public async getGasPriceEstimationOrThrowAsync(): Promise<BigNumber> {
-        return this._protocolFeeUtils.getGasPriceEstimationOrThrowAsync();
+        const gasPrices = await this._protocolFeeUtils.getGasPriceEstimationOrThrowAsync({
```
kyu-c, line 149: Here I think we shouldn't provide a default. (0.1 gwei as a sefault only makes sense on Arbitrum).

kyu-c (changes requested):
I think the behavior changed in subtle way. I

src/asset-swapper/utils/gas_price_utils.ts
```diff
@@ -41,16 +42,26 @@ export class ProtocolFeeUtils {
         return instance;
     }
 
-    /** @returns gas price (in wei) */
-    public async getGasPriceEstimationOrThrowAsync(shouldHardRefresh?: boolean): Promise<BigNumber> {
-        if (this._gasPriceEstimation.eq(constants.ZERO_AMOUNT)) {
-            return this._getGasPriceFromGasStationOrThrowAsync();
+    public async getGasPriceEstimationOrDefault(defaultGasPrices: GasPrices): Promise<GasPrices> {
+        if (this._gasPriceEstimation === undefined){
```
kyu-c: nit: if `_updateGasPriceFromOracleOrThrow` fails for some reason, it will throw an error rather than returning the default. 

I think it's fine to return the default if `this._gasPriceEstimation` is undefined. 
If we are calling `_updateGasPriceFromOracleOrThrow` then I think we should swallow the error and return the default.

kyu-c (approved):

PR #1122, opened 2022-12-04
Phil/0x api basic

phil-ociraptor:
# Description

Working on a proof of concept on how we might refactor 0x API for clarity. It should handle requests to `/basic/swap/v1/quote` and return mostly similar results.

There are two major missing features:
- No RFQt
- No QuoteReport

But will add those once we get some feedback.

Main insight is the following. 0x API can essentially structured as
- Sampling
- Routing
- SwapQuote generation (a weird but highly entangled abstraction at this point). The following use it:
  - Calldata generation
  - Estimate gas
  - Calculate Protocol fees
  - Estimate Price Impact
  - Estimate Slippage

And it should be relatively easy to follow when we are doing each action.

# Testing & Simbot Run

In my very initial testing, I see that things are working well, with the exception that gasEstimates don't seem sufficiently padded.

# Checklist


src/services/basic_swap_service.ts
```diff
@@ -0,0 +1,767 @@
...
+
+        this._swapQuoteConsumer = new SwapQuoteConsumer(this._swapQuoterOpts);
+        this._web3Wrapper = new Web3Wrapper(this._provider);
+
+        this._contractAddresses = contractAddresses;
+        this._wethContract = new WETH9Contract(this._contractAddresses.etherToken, this._provider);
+        this._fakeTaker = new FakeTakerContract(NULL_ADDRESS, this._provider);
+    }
+
+    public async getQuoteAsync(params: GetSwapQuoteParams): Promise<GetSwapQuoteResponse> {
```
phil-ociraptor, line 245: This is the function to focus on.

PR #1126, opened 2022-12-07
chore: moving getDefaultAppDeps into runners/utils

joshrrose:
fixed the last circular dep
# Description

# Testing & Simbot Run

# Checklist


test/rfqt_test.ts
```diff
@@ -13,7 +13,8 @@ import * as _ from 'lodash';
 import 'mocha';
 import * as request from 'supertest';
 
-import { getAppAsync, getDefaultAppDependenciesAsync } from '../src/app';
+import { getAppAsync } from '../src/app';
+import { getDefaultAppDependenciesAsync } from '../src/runners/utils';
```
SavDont, line 17: Do we want to also delete the require cache for `rfqt_test.ts` and `rfq_blocked_address_utils_test.ts` as well?

SavDont (changes requested):
Added 1 comment.
Since this is the last of the circular deps, can we include a check within the CI to ensure we don't reintroduce any circular deps.

test/rfqt_test.ts
joshrrose: i don't think we need it. the tests pass with the code the way it is now. so don't want to add weird stuff that isn't really required.

package.json
```diff
@@ -34,6 +34,7 @@
         "fix": "eslint --fix src test && yarn prettier",
         "prettier": "prettier --write ${npm_package_config_prettier_target} --config .prettierrc",
         "prettier:ci": "prettier --list-different ${npm_package_config_prettier_target} --config .prettierrc",
+	"circular": "madge --circular --extensions ts ./",
```
SavDont, line 37: nit: can we align this with rest of the scripts?

SavDont (approved):
LGTM

PR #1132, opened 2022-12-09
chore: Remove default excluded sources [LIT-690]

kyu-c:
# Description

Remove default excluded sources and consolidate bridge source configuration
* For determining whether limit order is enabled, use buy/sell quote sources instead of relying on default excluded list

# Testing & Simbot Run

* Manual testing 
* [Simbot sanity run](https://metabase.spaceship.0x.org/dashboard/186?run_id=2022-12-excluded-sources)

# Checklist

### DIFF
--- a/src/asset-swapper/swap_quoter.ts
+++ b/src/asset-swapper/swap_quoter.ts
@@ -25,7 +25,13 @@ import {
 } from './types';
 import { MarketOperationUtils } from './utils/market_operation_utils';
 import { BancorService } from './utils/market_operation_utils/bancor_service';
-import { DEFAULT_GAS_SCHEDULE, SAMPLER_ADDRESS, SOURCE_FLAGS } from './utils/market_operation_utils/constants';
+import {
+    BUY_SOURCE_FILTER_BY_CHAIN_ID,
+    DEFAULT_GAS_SCHEDULE,
+    SAMPLER_ADDRESS,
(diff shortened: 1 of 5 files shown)


src/asset-swapper/swap_quoter.ts
```diff
@@ -206,16 +194,17 @@ export class SwapQuoter {
...
         };
+
         // pass the rfqClient on if rfqt enabled
         if (calcOpts.rfqt !== undefined) {
             calcOpts.rfqt.quoteRequestor = new QuoteRequestor();
             calcOpts.rfqt.rfqClient = rfqClient;
         }
-
-        const result: OptimizerResultWithReport = await this._marketOperationUtils.getOptimizerResultAsync(
+        const limitOrders = await this.getLimitOrders(marketOperation, makerToken, takerToken, calcOpts);
```
kliao, line 203: how's the test coverage on this? i see we have a unit test for "should respect excludedSources" which should cover this?

kliao (approved):

src/asset-swapper/swap_quoter.ts
kyu-c: "should respect excludedSources" does something else imo so i renamed it... 👀 

I found a compromised way testing this (better than not having any coverage).
Ideally the `swap_test.ts` should support at least one non-`Native` source... 😣

PR #1133, opened 2022-12-09
fix: Add OTC Order as RFQ for Quote Reports

Ktl-XV:
# Description

During Quote Report generation we do special handing of RFQ sources, with RFQt v2 we need to treat OTC Order as RFQ.

# Testing & Simbot Run

# Checklist

-   [x ] All dependent changes (e.g. RFQ server, Gas Price Oracle) have been deployed (if any).


src/asset-swapper/utils/quote_report_generator.ts
```diff
@@ -287,6 +287,16 @@ function _isNativeOrderFromCollapsedFill(cf: Fill): cf is Fill<NativeFillData> {
...
 }
 
+function _isRFQOrderfromType(orderType: FillQuoteTransformerOrderType) {
+    switch (orderType) {
+        case FillQuoteTransformerOrderType.Rfq:
+        case FillQuoteTransformerOrderType.Otc:
+            return true;
+        default:
+            return false;
+    }
```
phil-ociraptor, line 297: One small suggestion.

We often use this pattern in situations like this:
```suggestion
    switch (orderType) {
        case FillQuoteTransformerOrderType.Rfq:
        case FillQuoteTransformerOrderType.Otc:
            return true;
        case FillQuoteTransformerOrderType.Limit:
        case FillQuoteTransformerOrderType.Bridge:
            return false;             
        default:
            ((_: never) => {
                throw new Error('unreachable');
            })(orderType);
    }
```

By having a default case throw an error with type `never` - we actually get type safety that we exhaustively switch over all the potential cases. In the future, if someone adds a new order type, we will get a compile time error and it will force us to handle this then

phil-ociraptor (approved):

PR #1153, opened 2022-12-23
chore: Remove `_size` in `Path` and add some tests for `adjustedRate()` [LIT-690]

kyu-c:
# Description

* Remove `_size` in `Path` as it's redundant 
* Add tests for `adjustedRate()`
* Make `adjustedSize()` and `adjustedCompleteRate()` private 

# Testing & Simbot Run

* Added new unit tests.

# Checklist

### DIFF
--- a/test/asset-swapper/utils/market_operations_utils/path_test.ts
+++ b/test/asset-swapper/utils/market_operations_utils/path_test.ts
@@ -1,0 +1,76 @@
+import * as chai from 'chai';
+import 'mocha';
+import {
+    BigNumber,
+    ERC20BridgeSource,
+    Fill,
+    FillQuoteTransformerOrderType,
+    MarketOperation,
+} from '../../../../src/asset-swapper';
+import { ONE_ETHER } from '../../../../src/asset-swapper/utils/market_operation_utils/constants';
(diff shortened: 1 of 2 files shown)


test/asset-swapper/utils/market_operations_utils/path_test.ts
```diff
@@ -0,0 +1,76 @@
...
+            const path = Path.create(
+                MarketOperation.Sell,
+                [
+                    createFakeFill({
+                        input: ONE_ETHER,
+                        adjustedOutput: ONE_ETHER.times(990),
+                    }),
+                    createFakeFill({
+                        input: ONE_ETHER,
+                        output: ONE_ETHER.times(1000),
```
kliao, line 45: how come sometimes there is output but sometimes only adjustedOutput?

kliao (approved):

test/asset-swapper/utils/market_operations_utils/path_test.ts
kyu-c: `output` is only used when the sum of inputs is greater than the target. 

I suspect that this does not happen in prod though 😅 (I didn't want to change any behavior in this PR since it's a refactoring PR)

PR #1163, opened 2023-01-02
chore: Implement `Path.hasTwoHop` [LIT-690]

kyu-c:
# Description

* Implement `Path.hasTwoHop`

# Testing & Simbot Run

* Added unit tests.

# Checklist

### DIFF
--- a/src/asset-swapper/utils/market_operation_utils/path.ts
+++ b/src/asset-swapper/utils/market_operation_utils/path.ts
@@ -10,7 +10,7 @@ import {
     Fill,
 } from '../../types';
 
-import { ZERO_AMOUNT } from './constants';
+import { SOURCE_FLAGS, ZERO_AMOUNT } from './constants';
 import { ethToOutputAmount } from './fills';
 import {
     createBridgeOrder,
@@ -58,6 +58,10 @@ export class Path {
         public readonly sourceFlags: bigint,
         protected readonly adjustedSize: PathSize,
     ) {}
+
+    public hasTwoHop(): boolean {
+        return (this.sourceFlags & SOURCE_FLAGS[ERC20BridgeSource.MultiHop]) > 0;
+    }
 
     public createOrders(): OptimizedOrder[] {
         const { makerToken, takerToken } = getMakerTakerTokens(this.context);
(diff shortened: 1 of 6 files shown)


src/asset-swapper/utils/market_operation_utils/path.ts
```diff
@@ -59,6 +59,10 @@ export class Path {
         protected readonly adjustedSize: PathSize,
     ) {}
 
+    public hasTwoHop(): boolean {
+        return (this.sourceFlags & SOURCE_FLAGS[ERC20BridgeSource.MultiHop]) > 0;
```
kliao, line 63: is this ok to be different than the previous logic `path.sourceFlags === SOURCE_FLAGS[ERC20BridgeSource.MultiHop]`?

test/asset-swapper/utils/market_operations_utils/path_test.ts
```diff
@@ -93,6 +97,52 @@ describe('Path', () => {
...
+        it('Returns true when the path includes a two hop', () => {
+            const path = Path.create(
+                {
+                    side: MarketOperation.Sell,
+                    inputToken: 'fake-input-token',
+                    outputToken: 'fake-output-token',
+                },
+                [
+                    createFakeFillWithFlags(SOURCE_FLAGS[ERC20BridgeSource.UniswapV3]),
+                    createFakeFillWithFlags(SOURCE_FLAGS[ERC20BridgeSource.MultiHop]),
```
kliao, line 132: does this mean this path has 3 hops? 1. UniswapV3 and 2/3. Multihop? i'm just a little confused by the meaning of the presence of the MultiHop flag.

src/asset-swapper/utils/market_operation_utils/path.ts
kyu-c: yes it's intentional! (I don't like the flag-based approach and want to change it to a set in the future).

kliao (approved):

test/asset-swapper/utils/market_operations_utils/path_test.ts
kyu-c: It means that the path has:
image.

The proportion depends on the fill amount. For this test case, it's only testing flag so I omitted the fill amount intentionally. 

(I think a path is a misnomer -- probably Route makes more sense... I will consider renaming it in the future).

Flag is essentially a set of DEXes used in the path.

PR #1169, opened 2023-01-04
chore: Make `Path.createOrders` readonly [LIT-690]

kyu-c:
# Description

* Make `Path.createOrders` readonly
  * Initialize optimizer orders in `Path.create`
* Rename `Path.createOrders` and `Path.createSlippedOrders` to `get*`

# Testing & Simbot Run

* Passes existing tests.

# Checklist


test/asset-swapper/utils/market_operations_utils/path_test.ts
```diff
@@ -486,21 +484,19 @@ describe('Path', () => {
...
-    const { input, output, adjustedOutput, source } = params;
+function createFakeBridgeFill(params: { input: BigNumber; output?: BigNumber; adjustedOutput?: BigNumber }): Fill {
+    const { input, output, adjustedOutput } = params;
     return {
         input,
         output: output || new BigNumber(0),
         adjustedOutput: adjustedOutput || new BigNumber(0),
         gas: 42,
-        source: source || ERC20BridgeSource.UniswapV3,
+        source: ERC20BridgeSource.UniswapV2,
```
kliao, line 494: how come we switched from v3 to v2?

kliao (approved):

test/asset-swapper/utils/market_operations_utils/path_test.ts
kyu-c: good question!

v3 is a special case (😞) that requires a valid filldata (`createFinalBridgeOrderFillDataFromCollapsedFill`) during order creation.

Using uniswap v3 would make `path_test.ts` very cluttered. Since validating uniswap v3 order isn't the focus of `path_test.ts` (it probably belongs to `order.ts`s unit tests) I decided to keep it simple and use v2.

PR #1174, opened 2023-01-05
chore: Scaffold meta-transaction v2 endpoints [RFQ-768]

Haozhuo:
# Description

Scaffold meta-transaction v2 endpoints in meta-transaction router, handler and service. Most of the changes are a copy of existing meta-transaction endpoints & functions with minor modifications. They will be further modified in the coming weeks but for now, scaffolding is sufficient. 

The new changes are mainly adding fees to request & response types and parsing fees.

# Testing & Simbot Run

# Checklist


src/types.ts
```diff
@@ -256,11 +256,28 @@ export interface MetaTransactionV1QuoteResponse extends BasePriceResponse {
     orders?: any;
 }
 
+/**
+ * Response type for /meta_transaction/v2/quote endpoint
+ */
+export interface MetaTransactionV2QuoteResponse extends BasePriceResponse {
```
patrick-dowell, line 262: This may be a dumb question, but what's the difference between quote and price?

src/types.ts
```diff
@@ -289,6 +313,13 @@ export interface MetaTransactionV1QuoteParams extends SwapQuoteParamsBase {
     takerAddress: string;
 }
 
+/**
+ * Parameters for the V2 Meta Transaction Service price and quote functions.
+ */
+export interface MetaTransactionV2QuoteParams extends MetaTransactionV1QuoteParams {
```
patrick-dowell, line 319: why is there a MetaTransactionV2QuoteParams and MetaTransactionsV2QuoteRequestParams? Seems like you should only need one or the other.

src/services/meta_transaction_service.ts
```diff
@@ -109,6 +166,79 @@ export class MetaTransactionService implements IMetaTransactionService {
...
+            shouldSellEntireBalance: false,
+            skipValidation: true,
+        };
+
+        const quote = await this._swapService.calculateSwapQuoteAsync(quoteParams);
+
+        // Quote Report
+        if (endpoint === 'quote' && quote.extendedQuoteReportSources && kafkaProducer) {
+            const quoteId = getQuoteIdFromSwapQuote(quote);
+            publishQuoteReport(
```
patrick-dowell, line 196: should we publish fee info to kafka in the quote report?

src/handlers/meta_transaction_handlers.ts
```diff
@@ -237,3 +373,129 @@ function parseV1RequestParams(req: express.Request): MetaTransactionV1QuoteReque
...
+
+    if (req.query.feeConfigs) {
+        const feeConfigs = req.query.feeConfigs as ParsedQs;
+        parsedFeeConfigs = {};
+
+        // Parse the integrator fee config
+        if (feeConfigs.integrator) {
+            const integratorFee: ParsedQs = feeConfigs.integrator as ParsedQs;
+
+            if (integratorFee.kind !== 'volume') {
```
patrick-dowell, line 405: isn't volume the only kind for integratorFee?

src/handlers/meta_transaction_handlers.ts
```diff
@@ -237,3 +373,129 @@ function parseV1RequestParams(req: express.Request): MetaTransactionV1QuoteReque
...
+                    {
+                        field: 'feeConfigs',
+                        code: ValidationErrorCodes.IncorrectFormat,
+                        reason: ValidationErrorReasons.InvalidGaslessFeeKind,
+                    },
+                ]);
+            }
+
+            const feePercentage = new BigNumber(zeroexFee.volumePercentage as string);
+            if (feePercentage.gt(1)) {
```
patrick-dowell, line 451: Do 100% fees even make sense?

src/handlers/meta_transaction_handlers.ts
```diff
@@ -237,3 +373,129 @@ function parseV1RequestParams(req: express.Request): MetaTransactionV1QuoteReque
...
+                    integratorSharePercentage: new BigNumber(zeroexFee.integratorSharePercentage as string),
+                };
+            }
+        }
+
+        // Parse the gas fee config
+        if (feeConfigs.gas) {
+            const gasFee: ParsedQs = feeConfigs.gas as ParsedQs;
+
+            if (gasFee.kind !== 'gas') {
```
patrick-dowell, line 482: can gasFee ever not be gas?

src/handlers/meta_transaction_handlers.ts
```diff
@@ -39,6 +47,134 @@ export class MetaTransactionHandlers {
         this._metaTransactionService = metaTransactionService;
     }
 
+    /**
+     * Handler for the /meta_transaction/v2/quote endpoint
+     */
+    public async getV2QuoteAsync(req: express.Request, res: express.Response): Promise<void> {
```
patrick-dowell, line 53: how much shared code is there between this logic and the v1QuoteAsync logic? Same for v1PriceAsync?

patrick-dowell (commented):
Mostly just comments. The code itself looks reasonable to me, but I do have questions!

src/types.ts
Haozhuo: No it's a great question! In general, `/quote` endpoint is intended to be called when there is an intention to trade (there are more processing we do on our end). If not, `/price` endpoint should be called if you're just loosely checking. This is a pattern we use in our other APIs as well.

In the case of meta-transaction, `/price` and `/quote` are similar but no meta-transaction object is returned & no quote report is generated in `/price` (there are some other minor differences but not huge).

src/types.ts
Haozhuo: `MetaTransactionsV2QuoteRequestParams` is the type after parsing the incoming request to `/quote` (same for `MetaTransactionsV1QuoteRequestParams`). 
`MetaTransactionV2QuoteParams` is the type being passed to  `getMetaTransactionV2QuoteAsync` of `MetaTransactionService` (same for `MetaTransactionV1QuoteParams`).

The intention here is to keep it consistent with V1 types.

I think the intention on why V1 is structured this way is to have a type that purely represents the incoming request (`QuoteRequestParams`). 
Once we have the parsed request, we can do some processing and fields to it and thus another type (`QuoteParams`). The new type represents a process / modified object from the incoming request which is not changed, thus the distinction of the 2 types.

src/services/meta_transaction_service.ts
Haozhuo: Yep. We should add it after the fee calculation & fee storage code are completed. For scaffolding we can't add it now as the necessary code has been committed yet

src/handlers/meta_transaction_handlers.ts
Haozhuo: Yes for now. Here trying to add an extra validation here for the case that the caller might provide the wrong `kind` and provide error code.

src/handlers/meta_transaction_handlers.ts
Haozhuo: Good catch! Will change it to `gte`

src/handlers/meta_transaction_handlers.ts
Haozhuo: If no mistake is made, then `gasFee` should always have kind `gas`. Again trying to be extra safe here in the case of mistake and surface the error to the caller

src/handlers/meta_transaction_handlers.ts
Haozhuo: Assuming you're talking about handler level, parsing of incoming request and get meta-transaction v2 quote are different. The majority of the shared code on handler level is error handling

Haozhuo:
https://linear.app/0xproject/issue/RFQ-768/meta-transaction-v2-scaffolding

phil-ociraptor (approved):
Approving the scaffolding for v2 - for the service layer, let's try to keep v1 and v2 unified where we can

src/handlers/meta_transaction_handlers.ts
Haozhuo: Got it. I will do another refactoring after merging this scaffolding PR. Want to limit the scope of the scaffolding

PR #1175, opened 2023-01-06
chore: Implement `Path.getSlippedOrdersByType` [LIT-690]

kyu-c:
# Description

* Implement `Path.getSlippedOrdersByType`
  * which will be used instead of `Path.getSlippedOrders` in certain places (if not all)

# Testing & Simbot Run

* Added unit tests.

# Checklist

### DIFF
--- a/src/asset-swapper/utils/market_operation_utils/path.ts
+++ b/src/asset-swapper/utils/market_operation_utils/path.ts
@@ -81,35 +81,34 @@ export class Path implements IPath {
     }
 
     /**
+     * Returns `OptimizedOrdersByType` with slippage applied (Native orders do not have slippage).
+     * @param maxSlippage maximum slippage. It must be [0, 1].
+     * @returns orders by type by with slippage applied when applicable.
+     */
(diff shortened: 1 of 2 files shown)


src/asset-swapper/utils/market_operation_utils/path.ts
```diff
@@ -80,36 +80,35 @@ export class Path implements IPath {
...
 
+    /**
+     * Returns `OptimizedOrdersByType` with slippage applied (Native orders do not have slippage).
+     * @param maxSlippage maximum slippage. It must be [0, 1].
+     * @returns orders by type by with slippage applied when applicable.
+     */
+    public getSlippedOrdersByType(maxSlippage: number): OptimizedOrdersByType {
+        checkSlippage(maxSlippage);
+
+        const { nativeOrders, twoHopOrders, bridgeOrders } = this.getOrdersByType();
```
kliao, line 91: just something to think about -- is it possible we'll have different order types or orders with > 2 hops in the future? it might make sense to have a slippage applier class that loops through orders and applies slippage

kliao (approved):

kliao (approved):

PR #1180, opened 2023-01-10
feat: Add gasless fee calculator [RFQ-769]

Haozhuo:
# Description

Add gasless fee calculator

# Testing & Simbot Run

Unit test

# Checklist


src/types.ts
```diff
@@ -492,6 +492,9 @@ interface IntegratorShareFeeConfig extends GaslessFeeConfigBase {
     integratorSharePercentage: BigNumber;
 }
 
+export type IntegratorFeeConfig = VolumeBasedFeeConfig;
+export type ZeroExFeeConfig = VolumeBasedFeeConfig | IntegratorShareFeeConfig;
```
phil-ociraptor, line 496: Hmmm, these type aliases feel like they might not be all that helpful. My vote would be to remove them since you can always just declare stuff like:


```
interface MyInterface {
  zeroExFeeConfig: VolumeBasedFeeConfig | IntegratorShareFeeConfig;
  integratorFeeConfig: VolumeBasedFeeConfig;
}
```

instead of 

```
interface MyInterface {
  zeroExFeeConfig: ZeroExFeeConfig;
  integratorFeeConfig: IntegratorFeeConfig;
}
```
I prefer not creating extra types

src/utils/gasles_fee_calculator.ts
```diff
@@ -0,0 +1,200 @@
...
+ *
+ * @param feeConfigs Gasless fee configs parsed from input.
+ * @param opts sellToken: Address of the sell token.
+ *             sellTokenAmount: Amount of the sell token.
+ *             sellTokenAmountPerBaseUnitNativeToken: Amount of sell token per base unit native token.
+ *             gasPrice: Estimated gas price.
+ *             quoteGasEstimate: The gas estimate to fill the quote.
+ * @returns Gasless fees object and the total fee amount.
+ */
+export function calculateGaslessFees(opts: {
```
phil-ociraptor, line 33: I wonder if we need to call this `calculateGaslessFees` or if we could just call this `calculateFees` and if the gas potion of the `feeConfigs` is missing, we just don't calculate gas. 

So in other words, we generalize this class to `fee_calculator` - and it works in both a gasless and regular setting

phil-ociraptor: otherwise, I think this looks great!

phil-ociraptor (approved):

PR #1190, opened 2023-01-12
refactor: Require token adjacency graph to be explicitly defined for each chain [LIT-768]

kyu-c:
# Description

* Do not use `valueByChainId` for token adjacency graph and the intermediate token list.
* This change will force those to be defined when a new chain is added.

# Testing & Simbot Run

# Checklist


dextracker (approved):
🙏

PR #1200, opened 2023-01-19
feat: add rfqt availability check

byeongminP:
# Description

Ensures that if a request is exclusively asking for RFQt liquidity, but RFQt is not enabled for the chain, the server responds with an explicit error message. Also kills RFQt-v2 rollout mechanism for easier config maintenance.

# Testing & Simbot Run

# Checklist

### DIFF
--- a/src/services/swap_service.ts
+++ b/src/services/swap_service.ts
@@ -268,7 +268,7 @@ export class SwapService implements ISwapService {
 
         let _rfqt: GetMarketOrdersRfqOpts | undefined;
 
-        if (rfqt?.nativeExclusivelyRFQ && !this._rfqClient.isRfqtEnabled()) {
+        if (rfqt && !this._rfqClient.isRfqtEnabled()) {
             throw new ServiceDisabledError('RFQt Service is not available');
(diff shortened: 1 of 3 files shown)


src/services/swap_service.ts
```diff
@@ -266,6 +267,11 @@ export class SwapService implements ISwapService {
         } = params;
 
         let _rfqt: GetMarketOrdersRfqOpts | undefined;
+
+        if (rfqt?.nativeExclusivelyRFQ && !this._rfqClient.isRfqtEnabled()) {
+            throw new ServiceDisabledError('RFQt Service is not available');
+        }
```
phil-ociraptor, line 273: I believe `nativeExclusivelyRFQ` will only be true when the included sources has `RFQT` but not `Native`, `Mesh`, or `0x`. 

So, for example, it will fire if `includedSources=UniswapV3,RFQT` but not `includedSources=UniswapV3,0x`


Is the intent to fire when `rfqt` is enabled?

phil-ociraptor (approved):

PR #1202, opened 2023-01-19
feat: implementing sell function for UniV3 MultiQuoter [LIT-763]

SavDont:
# Description
Initial implementation and sell function UniswapV3MultiQuoter. Note that current UniswapV3Sampler doesn't use this yet.

# Testing & Simbot Run

# Checklist

### DIFF
--- a/contracts/src/UniswapV3MultiQuoter.sol
+++ b/contracts/src/UniswapV3MultiQuoter.sol
@@ -125,10 +125,6 @@ contract UniswapV3MultiQuoter {
         }
     }
 
-    /// amounts must be monotonically increasing
-    /// TODO: double check behavior of loop when we cross sqrtPriceLimitX96
-    /// TODO make sure that interfaces are in separate file
-
     /// @notice swap multiple amounts of token0 for token1 or token1 for token1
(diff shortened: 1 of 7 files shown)


kliao (approved):

contracts/src/UniswapV3MultiQuoter.sol
```diff
@@ -0,0 +1,301 @@
...
+                }
+
+                state.amountsIndex += 1;
+                state.amountSpecifiedRemaining = amounts[state.amountsIndex].sub(amounts[state.amountsIndex - 1]);
+            }
+        }
+
+        for (uint256 i = state.amountsIndex; i < amounts.length; ++i) {
+            (result.amounts0[i], result.amounts1[i]) = zeroForOne == exactInput
+                ? (amounts[i] - state.amountSpecifiedRemaining, state.amountCalculated)
```
kliao, line 241: would the router logic get confused by this? i.e. returning prorated input amounts if we run our of liquidity instead of returning the original amountsIn

SavDont: I don't believe so. I believe router would only get confused if the output amounts are non-monotonic. In this case the sampler amounts are still monotonically increasing.

PR #1207, opened 2023-01-23
chore: upgrading sampler to v08 [LIT-783]

SavDont:
# Description

Upgrading sampler solidity contracts to use v08 solidity contracts.

# Testing & Simbot Run

[Mainnet](https://metabase.spaceship.0x.org/dashboard/186?run_id=sampler-solidityV08-upgrade-p3&adjusted_amount_bought_win_tolerance__bps_=1.0)

# Checklist


contracts/src/NativeOrderSampler.sol
```diff
@@ -191,7 +189,7 @@ contract NativeOrderSampler {
         if (
             orderInfo.status != IExchange.OrderStatus.FILLABLE ||
             !isSignatureValid ||
-            order.makerToken == IERC20TokenV06(0)
+            address(order.makerToken) == address(0)
```
dextracker, line 192: might want to change this to IERC20TokenV08 just to keep it consistent

contracts/src/interfaces/IGMX.sol
```diff
@@ -1,5 +1,5 @@
-pragma solidity ^0.6;
-pragma experimental ABIEncoderV2;
+pragma solidity ^0.8;
+pragma experimental ABIEncoderV2; //TODO: remove this
```
dextracker, line 2: i think ABIEncoderV2 can be dropped

dextracker (approved):
LGTM just a few nitpicks

kyu-c (approved):

kyu-c:
There is a quite meaningful difference in the simbot result. Any idea why that is the case?

SavDont:
> There is a quite meaningful difference in the simbot result. Any idea why that is the case?

I'm not too sure but could it potentially be due to routing through 0x liquidity which dev doesn't have access to?

kyu-c:
> > There is a quite meaningful difference in the simbot result. Any idea why that is the case?
> 
> I'm not too sure but could it potentially be due to routing through 0x liquidity which dev doesn't have access to?

Ah, yeah that seems the reason.

Running with `excludedSources=0x` can be useful for testing changes like this (no need to re-run it again as long as you feel confident).

SavDont:
Closing this for now, we're currently researching another solution for conflicting sampler contract version using overrides. Therefore, v08 contracts will likely not be needed as part of sampler.

PR #1213, opened 2023-01-24
feat: indicate rfqt availability for rfqt-only requests

byeongminP:
# Description

When RFQt is specifically asked but the service is unavailable, we currently return a `503` error, which does not accurately describe what went wrong. This PR makes it such that it would return `ServiceDisabled` error to indicate that RFQt is not available for the chain.

# Testing & Simbot Run

# Checklist


src/utils/rfq_client.ts
```diff
@@ -51,14 +51,6 @@ const toOtcOrder = (obj: any): OtcOrder => {
 };
 
 export class RfqClient {
-    private static isRolledOut(request: RfqtV2Request): boolean {
-        return isHashSmallEnough({
```
kyu-c, line 55: nit: clean up unused `import`s?

src/utils/rfq_client.ts
```diff
@@ -203,4 +186,11 @@ export class RfqClient {
             return [];
         }
     }
+
+    /**
```
kyu-c, line 190: nit/optional: seems unnecessary (doc doesn't seem to add much value)

kyu-c (approved):

PR #1219, opened 2023-01-25
feat: UniswapV3Sampler MultiQuoter integration via override [LIT-763]

SavDont:
# Description

This PR aims to implement the UniswapV3 optimized quoter into the current sampler using a geth override.

# Testing & Simbot Run
[Benchmark and Testing Report](https://www.notion.so/zeroex/UniswapV3-MultiQuoter-Benchmarks-be7e2dc0e1de4df780868c75f66b623e)

# Checklist


contracts/src/UniswapV3MultiQuoter.sol
```diff
@@ -280,6 +298,9 @@ contract UniswapV3MultiQuoter {
                     ? (amounts[state.amountsIndex], state.amountCalculated)
                     : (state.amountCalculated, amounts[state.amountsIndex]);
 
+                // TODO: find a better way to estimate gas. This is not an ideal way to measure gas
```
kliao, line 301: Maybe we can track the last used tickIndex and if it's the same, then carry forward the previous gas usage number.

contracts/src/UniswapV3MultiQuoter.sol
```diff
@@ -166,22 +167,39 @@ contract UniswapV3MultiQuoter {
...
             );
 
-            for (uint256 i = 0; i < amountsOut.length; ++i) {
-                amountsOut[i] = zeroForOne ? uint256(result.amounts0[i]) : uint256(result.amounts1[i]);
-                gasEstimate[i] += result.gasEstimates[i];
+            for (uint256 i = 0; i < nextAmountsLength; ++i) {
+                uint256 amountReceived = zeroForOne ? uint256(-result.amounts1[i]) : uint256(-result.amounts0[i]);
+                if (amountReceived != amountsOut[i]) {
+                    // for exact output swaps we need to check whether we would receive the full amount due to
+                    // multiswap behavior when hitting the limit
```
kliao, line 174: is this logic something we can remove if we fix the behavior in multiswap? maybe that would be a better place to fix it instead of here.

contracts/src/UniswapV3Sampler.sol
```diff
@@ -346,10 +256,10 @@ contract UniswapV3Sampler {
             }
         }
         // Must have a balance of both tokens.
-        if (pool.token0().balanceOf(address(pool)) == 0) {
+        if (IERC20TokenV06(pool.token0()).balanceOf(address(pool)) == 0) {
```
kliao, line 259: just curious -- how come this changed?

contracts/src/UniswapV3MultiQuoter.sol
SavDont: I was considering returning number of ticks crossed per sample amount and having MultiQuoter return some linear formula for determining the gas usage. I think gas estimate is linearly related to the number of ticks crossed so we can do some sort of linear regression across a bunch of samples to determine what the function parameters are.

contracts/src/UniswapV3MultiQuoter.sol
SavDont: I wanted to separate out the logic for multiswap since the logic there is already very complex. This special handling only needs to be done for exact output swaps so I thought this was a better area for this logic. I think multiswap should simply return deltas of of token0 and token1. Happy to discuss further offline.

contracts/src/UniswapV3Sampler.sol
SavDont: I removed the import for `IERC20TokenV06` in `IUniswapV3.sol`. This is the only line where we use `IERC20TokenV06` functions so didn't think it was needed as an import for the entire set of interfaces.

src/asset-swapper/swap_quoter.ts
```diff
@@ -87,9 +88,16 @@ export class SwapQuoter {
...
+        // Allow the UniV3 MultiQuoter bytecode to be written to a specic address
+        const uniV3MultiQuoterBytecode = _.get(
+            artifacts.UniswapV3MultiQuoter,
+            'compilerOutput.evm.deployedBytecode.object',
+        );
+
         const defaultCodeOverrides = samplerBytecode
             ? {
                   [samplerAddress]: { code: samplerBytecode },
+                  [UNISWAP_V3_MULTIQUOTER_ADDRESS]: { code: uniV3MultiQuoterBytecode },
```
kyu-c, line 100: nit/optional (maybe a future optimization): attach this on chains where uniswap v3 is deployed (it would a waste on other chains and bloat the request size).

contracts/src/UniswapV3Sampler.sol
```diff
@@ -165,39 +106,25 @@ contract UniswapV3Sampler {
...
+
+            bytes memory uniswapPath = toUniswapPath(reversedPath, poolPaths[i]);
+            (uint256[] memory amountsIn, uint256[] memory gasEstimate) = multiQuoter.quoteExactMultiOutput(
+                quoter.factory(),
+                uniswapPath,
+                makerTokenAmounts
+            );
+
+            for (uint256 j = 0; j < amountsIn.length; ++j) {
+                if (amountsIn[j] > 0 && (takerTokenAmounts[j] == 0 || takerTokenAmounts[j] > amountsIn[j])) {
```
kyu-c, line 122: nit/optional

Is this condition equivalent to `takerTokenAmounts[j] < amountsIn[j]` ?

kyu-c (approved):
The code LGTM.

I don't have a deep understanding of uni v3 multi quoter so wasn't able to give meaningful feedback on the solidity part (I will let @kliao to further comment on it).

As mentioned on Slack, I think we should look into price regression shown in the simbot result. It may be a good idea to also run simbot on another change as it's a pretty complex change.

contracts/src/UniswapV3MultiQuoter.sol
kliao: understood.
just to clarify what this logic is doing -- I see in line 193 it's zeroing out the amountsIn[i] if Multiswap wasn't able to fullfill the whole amountOut[i]. Is that to prevent the consumer (router?) of these amountsIn from breaking?

contracts/src/UniswapV3MultiQuoter.sol
```diff
@@ -295,9 +325,17 @@ contract UniswapV3MultiQuoter {
...
         }
+    }
 
-        result.gasEstimates[state.amountsIndex] = gasBefore - gasleft();
+    /// @notice Returns the multiswap gas estimate scaled to UniswapV3Pool:swap estimates
+    /// @dev These parameters have been determined by running a linear regression between UniswapV3QuoterV2
+    /// gas estimates and the unscaled gas estimates from multiswap
+    /// @param multiSwapEstimate the gas estimate from multiswap
+    /// @return swapEstimate the gas estimate equivalent for UniswapV3Pool:swap
+    function scaleMultiswapGasEstimate(uint256 multiSwapEstimate) private view returns (uint256 swapEstimate) {
```
kliao, line 337: can we include the test data and how the linear regression parameters were estimated somewhere in the repo? we'll probably have to do something similar for the new AMMs

kliao (approved):

contracts/src/UniswapV3MultiQuoter.sol
SavDont: Ah actually that is not needed since `amountsIn` will be set to 0 when initialized. 
Just a rough overview of this logic:
1. Check whether `multiswap` result returns amount to be received matches the amount we **want** out
2. If result has the full amount to be received, populate the `amountsOut` for the next hop based on amount payable 
3. If result doesn't have the full amount to be received, set a variable to track how many valid output amounts we have that we can send into the next hop (`nextAmountsLength`)
4. Repeat multiswap for each hop using a new amounts array with valid output amounts from previous hop

contracts/src/UniswapV3Sampler.sol
SavDont: Not quite, we want the best buy sample, so the pool that results in the smallest `amountsIn[j]` to be passed in. However, `amountsIn[j]` can be 0 so we have to filter out the 0s.

kliao (approved):

PR #1226, opened 2023-01-26
feat: Add a partial `Multiplex+MultiHop` support in `SamplerOperations` and `PathOptimizer` [LIT-637]

kyu-c:
# Description

Add a partial `Multiplex+MultiHop` support in `SamplerOperations` and `PathOptimizer`
* NOTE: this only adds `Multiplex+MultiHop` supports to some of the components. It is a no-op change until `MarketOperationUtils` actually starts sampling multiple amounts for two-hop.

# Testing & Simbot Run

* Added a unit test for `PathOptimizer`
* [simbot run](https://metabase.spaceship.0x.org/dashboard/186?run_id=2023-jan-mult-two-hop-samples&adjusted_amount_bought_win_tolerance__bps_=1.0)

# Checklist

### DIFF
--- a/src/asset-swapper/utils/market_operation_utils/path_optimizer.ts
+++ b/src/asset-swapper/utils/market_operation_utils/path_optimizer.ts
@@ -78,7 +78,7 @@ export class PathOptimizer {
 
     public findOptimalPathFromSamples(
         samples: DexSample[][],
-        twoHopQuotes: DexSample<MultiHopFillData>[],
+        twoHopQuotes: DexSample<MultiHopFillData>[][],
         nativeOrders: NativeOrderWithFillableAmounts[],
     ): Path | undefined {
         const beforeTimeMs = performance.now();
@@ -109,7 +109,7 @@ export class PathOptimizer {
 
     private findRoutesAndCreateOptimalPath(
         samples: DexSample[][],
-        twoHopSamples: DexSample<MultiHopFillData>[],
+        twoHopSamples: DexSample<MultiHopFillData>[][],
         nativeOrders: NativeOrderWithFillableAmounts[],
     ): { allSourcesPath: Path | undefined; vipSourcesPath: Path | undefined } | undefined {
         // Currently the rust router is unable to handle 1 base unit sized quotes and will error out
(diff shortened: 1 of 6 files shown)


src/asset-swapper/utils/market_operation_utils/path_optimizer.ts
```diff
@@ -78,7 +78,7 @@ export class PathOptimizer {
 
     public findOptimalPathFromSamples(
         samples: DexSample[][],
-        twoHopQuotes: DexSample<MultiHopFillData>[],
+        twoHopQuotes: DexSample<MultiHopFillData>[][],
```
kliao, line 81: just for my understanding -- the reason we're returning a 2D array instead of 1D array now is because we want to return the raw sample data to the router instead of simply picking the "best" first hop source and the "best" second hop source in TwoHopSampler.sol? so the goal is to allow multiplexing to happen in the router...have we done some benchmark around latency/price improvement or will we see that later when this code is ready?

kliao (approved):

src/asset-swapper/utils/market_operation_utils/path_optimizer.ts
kyu-c: >  the reason we're returning a 2D array instead of 1D array now is because we want to return the raw sample data to the router instead of simply picking the "best" first hop source and the "best" second hop source in TwoHopSampler.sol?

Hmm, it's about increasing the sampling granularity (we are still picking the best first hop source and the best second hop source in the same way).

We have been only sampling the whole amount for two hops so `twoHopQuotes` was an array of samples for the whole amount where each sample is for a different path.  Once we start sampling at a finer granularity for two-hops, we will have N samples per a two-hop path.

For example, for two hop quotes for sampling for the whole amount, the output would look like this
* path 0 (X -> A -> Y): 100
* path 1 (X -> B -> Y):  98
* path 2 (X -> C -> Y): 102

`twoHopQuotes` ~= [100, 98, 102] 

Suppose, we are now sampling 50%, and 100% amount (N  = 2)
* path 0 (X -> A -> Y): 50, 100
* path 1 (X -> B -> Y):  49, 98
* path 2 (X -> C -> Y): 51, 102

`twoHopQuotes` ~=  [ [50,100], [49, 98], [51, 102] ] 


> have we done some benchmark around latency/price improvement or will we see that later when this code is ready?
* Latency will definitely get worse (higher the N the worse latency). For lower N (~3), it was an acceptable increase.
* We have an intuition that it can improve pricing significantly for a very large swaps. We won't know the precise improvements until we run simbot.

PR #1227, opened 2023-01-26
feat: do not add positive slippage for non transformERC20 pathways

phil-ociraptor:
# Description

Avoid adding positive slippage for VIP pathways. This methodology is hacky - it uses two passes of calldata generation. In the future, with the `Rule` system we are building, it will be much easier to answer the question "will this SwapQuote be routed to a VIP" without having to generate the calldata.

For the sake of expediency, doing this minimal change. Make it work, then make it it right. Will refactor this all shortly.

# Testing & Simbot Run

# Checklist


phil-ociraptor:
Closing this since there's a better way

PR #1230, opened 2023-01-26
feat: add txOrigin field to MetaTransaction params

byeongminP:
# Description

# Testing & Simbot Run

# Checklist


src/handlers/meta_transaction_handlers.ts
```diff
@@ -345,10 +360,40 @@ function parseV1RequestParams(req: express.Request): MetaTransactionV1QuoteReque
...
-        // TODO: implement RFQt parsing logic
-        (excludedSources = []), (includedSources = []);
+        const parsedSourcesResult = parseUtils.parseRequestForExcludedSources(
+            {
+                excludedSources: req.query.excludedSources as string | undefined,
+                includedSources: req.query.includedSources as string | undefined,
+                intentOnFilling: intentOnFilling ? 'true' : 'false',
+                takerAddress,
+                txOrigin, // Gasless RFQt requires txOrigin in addition to takerAddress
+                apiKey,
```
phil-ociraptor, line 374: what is the apiKey here? Will it be a 0x internal one?  Since we do a lookup by integratorId, and that's currently passed in via a query parameter, would this apiKey just be used for securing down this endpoint (and not for integratorId lookup)?

phil-ociraptor (approved):

src/handlers/meta_transaction_handlers.ts
byeongminP: You're right in that the `apiKey` supposedly passed in here is not related to `integratorId`. If we were to make `apiKey` required for RFQt like vanilla `/swap`, then we can simply add extra validation to confirm that `apiKey` corresponds to the integrator, but the caller would need to supply both `apiKey` and `integratorId`. If we want to just use the `integratorId`, we would need to modify validation logic to bypass all `apiKey` checks.

phil-ociraptor: I think this is ok, but can we add a comment to highlight this behavior? Might be a little confusing since this logic drifts from `/swap` (and rightfully so)

PR #1246, opened 2023-02-02
fix: Use fee sources only when a swap size is extremely small (Optimism) [LIT-811]

kyu-c:
# Description
Use fee sources only when a swap size is extremely small (Optimism)

* Extremely small size (sub $1) is often over-optimized and routed to a niche liquidity (e.g. $50 TVL pool) and causes high revert rate.
* Introduce a concept of micro swap (defined as 1% of native fee token amount (from `NATIVE_FEE_TOKEN_AMOUNT_BY_CHAIN_ID`) and only route to fee sources and disable multi-hop.
  * On Optimism, that would translate to a swap size less than 1/1000 ETH.
* As an experimental feature, it's only enabled on Optimism to begin with.

# Testing & Simbot Run

* [Simbot run](https://metabase.spaceship.0x.org/dashboard/186?run_id=micro-swap-10c-100c--fix-v2&adjusted_amount_bought_win_tolerance__bps_=1.0) on $0.01-$1 (to verify the fix).
* [Simbot run](https://metabase.spaceship.0x.org/dashboard/186?run_id=micro-swap-fix-v2-all&adjusted_amount_bought_win_tolerance__bps_=1.0) on various sizes (for regression testing).

# Checklist

### DIFF
--- a/src/asset-swapper/utils/market_operation_utils/index.ts
+++ b/src/asset-swapper/utils/market_operation_utils/index.ts
@@ -1,3 +1,4 @@
+import { ChainId } from '@0x/contract-addresses';
 import { FillQuoteTransformerOrderType, RfqOrder } from '@0x/protocol-utils';
 import { BigNumber, NULL_ADDRESS } from '@0x/utils';
 import * as _ from 'lodash';
@@ -17,6 +18,7 @@ import {
     SignedLimitOrder,
     FeeEstimate,
     ExchangeProxyOverhead,
+    FillData,
 } from '../../types';
 import { getAltMarketInfo } from '../alt_mm_implementation_utils';
 import { QuoteRequestor, V4RFQIndicativeQuoteMM } from '../quote_requestor';
@@ -204,6 +206,7 @@ export class MarketOperationUtils {
             gasBefore,
             tokenDecimals,
             orderFillableTakerAmounts,
+            // TODO: rename inputAmountPerEth, outputAmountPerEth (the unit is wei and also it's per native token)
             outputAmountPerEth,
             inputAmountPerEth,
             dexQuotes,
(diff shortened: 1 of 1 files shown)


src/asset-swapper/utils/market_operation_utils/index.ts
```diff
@@ -778,19 +782,57 @@ export class MarketOperationUtils {
...
+    private isMicroSwap(inputAmount: BigNumber, inputAmountPerNativeWei: BigNumber): boolean {
+        // Only enable it on Optimism as it's experimental.
+        if (this.sampler.chainId !== ChainId.Optimism) {
+            return false;
+        }
+        const inputTokenValueInNativeWei = inputAmount.div(inputAmountPerNativeWei);
+
+        // If the value of input token amount is less than 1/100 of `this._nativeFeeTokenAmount`
+        // then it is considered as a micro swap.
+        // NOTE: gt is used because `takerAmountPerNativeWei` can be 0 when it's not available.
```
kliao, line 800: nit: do you mean inputAmountPerNativeWei

kyu-c: yes! fixed!

kliao (approved):

src/asset-swapper/utils/market_operation_utils/index.ts
```diff
@@ -359,6 +362,7 @@ export class MarketOperationUtils {
         SAMPLER_METRICS.logBlockNumber(blockNumber);
 
         const [makerTokenDecimals, takerTokenDecimals] = tokenDecimals;
+        const isMicroSwap = this.isMicroSwap(makerAmount, ethToMakerAssetRate);
```
kliao, line 365: would ethToMakeAssetRate  be really small for big trades? so we'd want the inverse?

kyu-c: discussed offline (trade size doesn't matter for this value)

PR #1252, opened 2023-02-07
chore: add prometheus counters to orderbook

nicob87:
# Description

Fixes issue RFQ-807

# Testing & Simbot Run

# Checklist


novaliu86 (commented):
I would recommend to use `express-prom-bundle`, which is a middleware to write metrics for http traffics. See example and comments here: https://github.com/0xProject/0x-rfq-api/blob/6f9595ae97036b22946a260abf97e53243aabec2/src/runners/http_rfqm_service_runner.ts#L175-L198

novaliu86 (approved):

PR #1258, opened 2023-02-10
chore: Refactor meta-transaction v2 to handle different meta-transaction types [RFQ-828]

Haozhuo:
# Description

Initially, meta-transaction v2 endpoints are designed to only handle meta-transaction v2 types. However, there is a new use case that needs v2 endpoints to support existing meta-transaction. This PR makes the corresponding changes.

# Testing & Simbot Run

# Checklist

### DIFF
--- a/src/types.ts
+++ b/src/types.ts
@@ -245,16 +245,16 @@ export interface MetaTransactionV1QuoteResponse extends BasePriceResponse {
  * Response type for /meta_transaction/v2/quote endpoint
  */
 export interface MetaTransactionV2QuoteResponse extends BasePriceResponse {
-    metaTransaction: MetaTransactionField;
+    trade: TradeContext;
     fees?: Fees;
 }
 
-type MetaTransactionField = MetaTransactionV1Field /* add meta-transaction v2 when it's ready */;
+type TradeContext = MetaTransactionV1TradeContext /* add MetaTransactionV2TradeContext when it's ready */;
 
-interface MetaTransactionV1Field {
-    kind: 'v1';
+interface MetaTransactionV1TradeContext {
+    kind: 'metatransaction';
     hash: string;
-    object: MetaTransaction;
+    metaTransaction: MetaTransaction;
 }
 
 /**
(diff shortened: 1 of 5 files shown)


src/types.ts
```diff
@@ -245,12 +245,18 @@ export interface MetaTransactionV1QuoteResponse extends BasePriceResponse {
...
  */
 export interface MetaTransactionV2QuoteResponse extends BasePriceResponse {
-    metaTransactionHash: string;
-    // TODO(vic): Update to MetaTransactionV2 when it's ready
-    metaTransaction: ExchangeProxyMetaTransaction;
+    metaTransaction: MetaTransactionField;
     fees?: Fees;
 }
 
+type MetaTransactionField = MetaTransactionV1Field /* add meta-transaction v2 when it's ready */;
```
Haozhuo, line 252: Need some suggestions on the name of the type @phil-ociraptor @byeongminP

src/types.ts
```diff
@@ -245,12 +245,18 @@ export interface MetaTransactionV1QuoteResponse extends BasePriceResponse {
...
+
+interface MetaTransactionV1Field {
+    kind: 'v1';
+    hash: string;
+    object: MetaTransaction;
```
Haozhuo, line 257: Same for this field

src/types.ts
byeongminP: If we were to go for the same nomenclature as `rfq-api`, I'd suggest `MetaTransactionV1Response`

src/types.ts
```diff
@@ -245,12 +245,18 @@ export interface MetaTransactionV1QuoteResponse extends BasePriceResponse {
  * Response type for /meta_transaction/v2/quote endpoint
  */
 export interface MetaTransactionV2QuoteResponse extends BasePriceResponse {
-    metaTransactionHash: string;
-    // TODO(vic): Update to MetaTransactionV2 when it's ready
-    metaTransaction: ExchangeProxyMetaTransaction;
+    metaTransaction: MetaTransactionField;
```
phil-ociraptor, line 248: can we call this
```suggestion
    trade: TradeContext;
```

phil-ociraptor (approved):

PR #1260, opened 2023-02-11
feat: integrate kyberswap elastic sampler with API

kliao:
# Description

Add KyberElastic sampler. Enable for Mainnet and Polygon. 

NOTE: Requires Protocol deployment (add KyberElastic to BridgeProtocol enum)

# Testing & Simbot Run

# Checklist


src/asset-swapper/utils/market_operation_utils/constants.ts
```diff
@@ -1431,6 +1436,15 @@ export const KYBER_DMM_ROUTER_BY_CHAIN_ID = valueByChainId<string>(
     NULL_ADDRESS,
 );
 
+export const KYBER_ELASTIC_CONFIG_BY_CHAIN_ID = valueByChainId(
+    {},
```
SavDont, line 1440: This map will default to a valid quoter, factory, and router addresses for every chain. I think this will cause sampler to fail for chains other than Mainnet and Polygon.

src/asset-swapper/utils/market_operation_utils/sampler_operations.ts
```diff
@@ -2044,6 +2062,20 @@ export class SamplerOperations {
                             ...intermediateTokens.map((t) => [takerToken, t, makerToken]),
                         ].map((path) => this.uniswapV3Sampler.createSampleBuysOperation(path, makerFillAmounts));
                     }
+                    case ERC20BridgeSource.KyberElastic: {
+                        // Rebasing tokens lead to a high revert rate.
+                        if (REBASING_TOKENS.has(takerToken) || REBASING_TOKENS.has(makerToken)) {
+                            return [];
+                        }
+                        const { factory, router } = KYBER_ELASTIC_CONFIG_BY_CHAIN_ID[this.chainId];
+                        if (!isValidAddress(router) || !isValidAddress(factory)) {
```
SavDont, line 2071: This path will never be hit, see comment on constants.ts.

SavDont (commented):
Left some nits and comments. In addition to addressing my comments, could youalso run a run local vs. local comparison of the API with KyberElastic vs. without it so we can see latency metrics for both mainnet and polygon.

contracts/src/KyberElasticMultiQuoter.sol
```diff
@@ -16,22 +16,21 @@
...
-import "./interfaces/IKyberSwapElastic.sol";
 import "./libraries/Multiswap.sol";
-import "@kyberelastic/interfaces/IFactory.sol";
-import "@kyberelastic/libraries/TickMath.sol";
-import "@kyberelastic/libraries/SwapMath.sol";
-import "@kyberelastic/libraries/Linkedlist.sol";
-import "@kyberelastic/libraries/LiqDeltaMath.sol";
-import "@kyberelastic/periphery/libraries/PathHelper.sol";
+import "./interfaces/IMultiQuoter.sol";
 import "@uniswap/v3-core/contracts/libraries/LowGasSafeMath.sol";
```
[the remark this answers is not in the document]
kliao: No, multiquoter uses it to do subtraction on the int256 amounts

src/asset-swapper/utils/market_operation_utils/constants.ts
kliao: This is intended to mean the addresses are the same for every chain. The source shouldn't be sampled if it is not listed in the constant SELL_SOURCE_FILTER_BY_CHAIN_ID. I'll give it a test.

src/asset-swapper/utils/market_operation_utils/sampler_operations.ts
kliao: I will remove this validation since the constructor of KyberElasticSampler already does a similar check.

src/asset-swapper/utils/market_operation_utils/constants.ts
kliao: I ran simbot for CELO chain, and it runs fine. The sources not listed in the source filters shouldn't be sampled, so we can control which samplers are "active" for each chain.

src/asset-swapper/utils/market_operation_utils/sampler_operations.ts
SavDont: Can you point me to where this validation is done? I don't see a constructor in KyberElasticSampler.

src/asset-swapper/utils/market_operation_utils/constants.ts
SavDont: Got it, sounds good to me.

src/asset-swapper/utils/market_operation_utils/sampler_operations.ts
kliao: https://github.com/0xProject/0x-api/blob/19876e1694fabe9dff5a6e6751de45eedc7827f7/src/samplers/kyber_elastic_sampler.ts#L32

your point stands that as is, the constructor validation also won't get hit. i think it's still good to validate the inputs in the constructor in case something is broken in the future

SavDont (approved):

PR #1275, opened 2023-02-16
feat: Add meta-transaction sell token fees

Haozhuo:
# Description

Add sell token fees for meta-transaction. The reason we'd like to introduce sell token fee is meta-transaction v2 (currently under development) might not have access to `buyToken`. It's currently being applied to meta-transaction v1 as the initial verion.

Main changes:
1. `fee_calculator` is where fee computation happens
2. If it's a `sell` operation, we need to subtract the fee before passing it down to sampler & router:
    - We will need to know gas cost ahead of time which requires a gas estimate according to historic data & sell token to native token rate (a `eth_call`)
3. Perform some adjustments for `price` and `sellAmount` similar to the existing `buyToken` fee adjustments

# Testing & Simbot Run
- Polygon meta-txn v1 endpoint regression test [simbot run](https://metabase.spaceship.0x.org/dashboard/186?run_id=regression-test-matcha-vs-dev-polygon-meta-txn-v1-endpoint-pt3&adjusted_amount_bought_win_tolerance__bps_=1.0)
- Polygon swap regression test [simbot run](https://metabase.spaceship.0x.org/dashboard/186?run_id=regression-test-matcha-vs-dev-polygon-swap&adjusted_amount_bought_win_tolerance__bps_=1.0)
- e2e meta-txn trades
- Unit tests

# Checklist


src/services/swap_service.ts
```diff
@@ -798,4 +889,63 @@ export class SwapService implements ISwapService {
...
+    /**
+     * Get sell token fee and corresponding on-chain transfers.
+     *
+     * @param sellToken Sell token address.
+     * @param sellAmount Sell amount.
+     * @param sellTokenAmountPerWei Sell token token amount per 1 wei native token.
+     * @param gasPrice Gas price.
+     * @param quoteGasEstimate Gas estimate for swap quote.
+     * @param feeConfigs Fee configs object.
+     * @param metaTransactionVersion Version of meta-transaction.
```
phil-ociraptor, line 902: this comment is incorrect

src/services/swap_service.ts
```diff
@@ -798,4 +889,63 @@ export class SwapService implements ISwapService {
...
+     *          used for on-chain transfers.
+     */
+    private _getSellTokenFees(
+        sellToken: string,
+        sellAmount: BigNumber,
+        sellTokenAmountPerWei: BigNumber | undefined,
+        gasPrice: BigNumber,
+        quoteGasEstimate: BigNumber,
+        feeConfigs: FeeConfigs | undefined,
+        onChainFeeTransferType: 'affiliateFee' | 'transferFrom',
```
phil-ociraptor, line 913: hmm, it's unclear what `affiliateFee` and `transferFrom` mean in this context.

src/utils/fee_calculator.ts
```diff
@@ -12,32 +13,62 @@ import {
...
  */
 const TRANSFER_FROM_GAS = new BigNumber(72e3);
 
+/**
+ * `transformerTransfer` estimated gas (used by `AffiliateFeeTransformer`):
+ * - Decrease balance of the owner (SLOAD + SSTORE): 24,000
+ * - Increase balance of the spender (SLOAD + SSTORE): 24,000
+ */
+const TRANSFER_GAS = new BigNumber(48e3);
+
```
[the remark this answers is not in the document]
phil-ociraptor: Seems high. I believe the following is true:
- If you SSTORE to something you've already done an SLOAD from, you save gas, so worst case I think is 22,100
- One of the storage slots should be non-zero, so it's either going from non-zero to non-zero, or non-zero to zero (which actually creates a gas refund), which should mean worst case is 5,000 gas (2900 for dirty update, 2100 for the original SLOAD)

So I'd say 27k gas is solid for Transfer

For TransferFrom, same logic on non-zero to non-zero, which should get us to 32k gas I think

Great website to play around with these values: https://evm.codes

src/utils/fee_calculator.ts
```diff
@@ -69,31 +101,112 @@ export function calculateFees(opts: {
...
+            break;
+        case 'affiliateFee':
+            onChainTransfersGas = AFFILIATE_FEE_TRANSFORMER_GAS.plus(TRANSFER_FROM_GAS).times(
+                feeRecipientToOnChainTransfer.size,
+            );
+            break;
+        default:
+            ((_: never) => {
+                throw new Error('unreachable');
+            })(onChainFeeTransferType);
```
phil-ociraptor, line 198: I feel like the `onChainFeeTransferType` abstraction is a lot of complexity to add the `AFFILIATE_FEE_TRANSFORMER_GAS` overhead. I would recommend we just ignore this overhead and just underquote gas (especially since it will be short lived)

phil-ociraptor (commented):
First pass - looking at the calculator and suggesting to remove the `onChainFeeTransferType` abstraction since it adds a lot of complexity, for a short term benefit. The affiliate fee transformer gas overhead can be recouped via other mechanisms

src/services/swap_service.ts
Haozhuo: Yeah. `onChainFeeTransferType` here means which way we're going to transfer the fee on-chain. Currently we only have affiliate fee transformer but later in meta-transaction v2 we will use `transferFrom` to directly transfer token instead of using affiliate fee transformer

src/utils/fee_calculator.ts
Haozhuo: Agree. Another use for `onChainFeeTransferType` (other than determining we're going to use affiliate fee transformer or use `transferFrom` directly to transfer fee) is to decide if we'd like to use `TRANSFER_GAS` (used by affiliate fee transfer) or `TRANSFER_FROM_GAS` (used by `transferFrom`). Here I made a typo. It should be:
```        case 'transferFrom':
            onChainTransfersGas = TRANSFER_FROM_GAS.times(...);
            break;
        case 'affiliateFee':
            onChainTransfersGas = AFFILIATE_FEE_TRANSFORMER_GAS.plus(TRANSFER_GAS).times(
               ...
            );
```

Haozhuo: Maybe the caller should just pass the gas cost for an on-chain transfer in as a parameter? That would simply the logic a lot

src/utils/fee_calculator.ts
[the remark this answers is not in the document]
Haozhuo: Thanks Phil! Yeah I played around with the values on https://evm.codes/ and your computation is correct. Worst case `27.1k` for `transfer` and `32.1K` for `transferFrom`. I have also included the detailed calculation in the comments above the constants

src/services/swap_service.ts
Haozhuo: Let me modify the function such the caller would just pass the gas cost for an on-chain transfer in as a parameter. That could simply the logic

src/asset-swapper/utils/market_operation_utils/index.ts
```diff
@@ -138,6 +138,33 @@ export class MarketOperationUtils {
...
 
+    /**
+     * Get the token amount per wei for native token. Note this function does not perform a full search (only perform
+     * a direct price check on `token` and native token) and calls `sampler.getBestNativeTokenSellRate` internally.
+     *
+     * @param token The token to check price for.
+     * @param opts GetMarketOrdersOpts object.
+     * @returns Token amount per wei.
+     */
+    public async getTokenAmountPerWei(token: string, opts?: Partial<GetMarketOrdersOpts>): Promise<BigNumber> {
```
phil-ociraptor, line 149: Two things:
- Functional: can we short circuit on Native Token and Wrapped Native token and just return 1? 
- Comment: so we sometimes call the output of this function `{input|output}AmountPerEth`, but here it's explicitly `PerWei` - can you leave a comment about the units?

src/services/swap_service.ts
```diff
@@ -315,6 +326,69 @@ export class SwapService implements ISwapService {
...
+
+        // Prepare Sell Token Fees
+        let sellTokenFees: Fees | undefined = undefined; // fees object to return to the caller
+        let sellTokenFeeAmount = ZERO; // total sell token fee to charge
+        let sellTokenFeeAmounts: AffiliateFeeAmount[] = []; // sell token fee amounts used by affiliate fee transformer
+        let sellTokenFeeOnChainTransferGas = ZERO; // the gas cost for transferring sell token fee on-chain
+        if (metaTransactionVersion !== undefined && marketSide === MarketOperation.Sell) {
+            // Use sell token as fee
+
+            // The check is only intended for compile time type checking
```
phil-ociraptor, line 340: We sometimes just say:
```suggestion
            // Narrow the type
```

src/services/swap_service.ts
```diff
@@ -315,6 +326,69 @@ export class SwapService implements ISwapService {
...
+                    {
+                        sellTokenFees,
+                        sellTokenFeeAmount,
+                        sellTokenFeeAmounts,
+                        sellTokenFeeOnChainTransferGas,
+                        sellAmount,
+                    },
+                    'sellAmount <= sell token fee amount',
+                );
+                throw new InsufficientFundsError(`sellAmount too small`);
```
phil-ociraptor, line 382: nit:
```suggestion
                throw new InsufficientFundsError(`sellAmount insufficient to cover fees`);
```

src/services/swap_service.ts
```diff
@@ -315,6 +326,69 @@ export class SwapService implements ISwapService {
...
+            }
+        }
+
+        const amount =
+            marketSide === MarketOperation.Sell
+                ? sellAmount?.minus(sellTokenFeeAmount)
+                : // eslint-disable-next-line @typescript-eslint/no-non-null-assertion -- TODO: fix me!
+                  buyAmount!.times(getBuyTokenPercentageFeeOrZero(affiliateFee) + 1).integerValue(BigNumber.ROUND_DOWN);
```
phil-ociraptor, line 390: this line seems fairly cryptic - I don't understand it without a comment... I guess it's not something you actually changed, but you will now get the git blame 😆

src/services/swap_service.ts
```diff
@@ -800,4 +895,63 @@ export class SwapService implements ISwapService {
...
         return (await this._altRfqMarketsCache.getResultAsync()).result;
     }
+
+    /**
+     * Get sell token fee and corresponding on-chain transfers.
+     *
+     * @param feeConfigs Fee configs object.
+     * @param sellToken Sell token address.
+     * @param sellAmount Sell amount.
+     * @param sellTokenAmountPerWei Sell token token amount per 1 wei native token.
```
phil-ociraptor, line 905: nit: can you explain what it means if this is undefined?

src/services/swap_service.ts
```diff
@@ -800,4 +895,63 @@ export class SwapService implements ISwapService {
 
         return (await this._altRfqMarketsCache.getResultAsync()).result;
     }
+
+    /**
+     * Get sell token fee and corresponding on-chain transfers.
+     *
+     * @param feeConfigs Fee configs object.
```
phil-ociraptor, line 902: nit: can you explain what it means if this is undefined?

src/utils/fee_calculator.ts
```diff
@@ -11,33 +12,48 @@ import {
...
- * - Increase balance of the spender (SLOAD + SSTORE): 24,000
- * - Update allowance of the spender (SLOAD + SSTORE): 24,000
- */
-const TRANSFER_FROM_GAS = new BigNumber(72e3);
+interface OnChainTransfer {
+    feeToken: string;
+    feeAmount: BigNumber;
+    feeRecipient: string;
+    affilateFeeType: AffiliateFeeType;
+}
```
phil-ociraptor, line 20: Hmmm... is this the same as the `AffiliateFeeAmount` type? I recommend against reusing it (especially the concept of `AffiliateFeeType`). Instead, I would just have the following fields and handle conversion outside of this file


```suggestion
interface OnChainTransfer {
    feeToken: string;
    feeAmount: BigNumber;
    feeRecipient: string;
}
```

phil-ociraptor (commented):
second round of comments - we're almost there!

src/services/swap_service.ts
Haozhuo: Yeah let me move it back to where it was and added comment. I only changed the line on `sellAmount` and did not touch the line related to `buyAmount`

src/asset-swapper/swap_quoter.ts
```diff
@@ -158,6 +158,13 @@ export class SwapQuoter {
         return new BigNumber(gasPrices.fast);
     }
 
+    /**
+     * Returns the token amount per wei by checking token <-> native token in sampler
+     */
+    public async getTokenAmountPerWei(token: string, options: Partial<SwapQuoteRequestOpts>): Promise<BigNumber> {
+        return await this._marketOperationUtils.getTokenAmountPerWei(token, options);
```
kyu-c, line 165: nit: del `await`

src/services/swap_service.ts
```diff
@@ -315,6 +326,63 @@ export class SwapService implements ISwapService {
...
+                sellToken,
+                sellAmount,
+                sellTokenAmountPerWei,
+                providedGasPrice,
+                AVG_MULTIPLEX_TRANFORM_ERC_20_GAS, // use historic average gas cost for multiplex & transformERC20
+                metaTransactionVersion === 'v1' ? TRANSFER_GAS : TRANSFER_FROM_GAS, // meta-transaction v1 uses affiliate fee transformer which calls `transformerTransfer` for on-chain fee transfer.
+                // meta-transaction v2 uses `FixinTokenSpender._transferERC20TokensFrom`.
+            ));
+
+            if (sellAmount.lte(sellTokenFeeAmount)) {
```
kyu-c, line 371: under what condition can this happen? (is this something that should never happen?)

Haozhuo: This can happen if the fees taken are greater than the `sellAmount` specified by the user. The most likely scenario would be if the trade size (say $1) is really small on mainnet, then gas fee cost (part of `sellTokenFeeAmount`) is very likely to be greater than `sellAmount`. I will add a comment on this condition

src/asset-swapper/utils/market_operation_utils/index.ts
Haozhuo: - should be able to short circuit on native / wrapped native token. There might an edge case where wrapped native token <-> native token is not close to 1 (however, i think extremely unlikely though)
- Yep I intentionally named the function & variables `perWei`. The naming for `{input|output}AmountPerEth` is confusing and it actually intends to mean `perWei` (this has caused some confusions when people read the code). I did not change the existing `AmountPerEth` name in this PR as I want to limit this PR to only metaTxn sell token fees. However, I think I will come up later and update the existing names to be `perWei`

src/asset-swapper/quote_consumers/feature_rules/transform_erc20_rule.ts
```diff
@@ -97,6 +99,17 @@ export class TransformERC20Rule extends AbstractFeatureRule {
                         })),
                 }),
             });
+
+            // Adjust the sell amount by the fee for meta-transaction v1. We don't need to adjust the amount for v2
+            // since fee transfer won't happen in `transformERC20`
+            if (metaTransactionVersion === 'v1') {
+                const totalSellTokenFeeAmount = sellTokenAffiliateFees.reduce(
```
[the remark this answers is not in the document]
Haozhuo: Thanks Kyu. For this, I will keep the original approach as `BigNumber.sum(...sellTokenAffiliateFees.map((fee) => fee.sellTokenFeeAmount))` would return a `BigNumber` whose value is `NaN` if `sellTokenAffiliateFees` is `[]`. However, in this case I'd like the value to be 0.

phil-ociraptor (approved):

src/services/meta_transaction_service.ts
```diff
@@ -216,7 +217,7 @@ export class MetaTransactionService implements IMetaTransactionService {
             endpoint,
             isUnwrap: false,
             isWrap: false,
-            metaTransactionVersion: 'v1',
+            metaTransactionVersion: params.metaTransactionVersion,
```
kyu-c, line 220: any updated needed in `meta_transaction_test.ts`?

kyu-c (approved):

src/services/meta_transaction_service.ts
Haozhuo: Nothing related to this for now 👍 `params.metaTransactionVerion` can only be `v1` for now since meta-txn v2 is not live yet. Will need to add more tests when we add support for meta-txn v2 but good for now 🚀 I also added some tests in `meta_transaction_test.ts` but not related to this

src/services/swap_service.ts
```diff
@@ -315,6 +326,65 @@ export class SwapService implements ISwapService {
...
 
+        const marketSide = sellAmount !== undefined ? MarketOperation.Sell : MarketOperation.Buy;
+        let feeToken = buyToken; // Default fee token is buy token
+
+        // Prepare Sell Token Fees
+        let sellTokenFees: Fees | undefined = undefined; // fees object to return to the caller
+        let sellTokenFeeAmount = ZERO; // total sell token fee to charge
+        let sellTokenFeeAmounts: AffiliateFeeAmount[] = []; // sell token fee amounts used by affiliate fee transformer
+        let sellTokenFeeOnChainTransferGas = ZERO; // the gas cost for transferring sell token fee on-chain
+        if (metaTransactionVersion !== undefined && marketSide === MarketOperation.Sell) {
```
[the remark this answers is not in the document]
Haozhuo: just did some refactoring. Now, most of lines here should be a function call (formatted by linter to span on multiple lines)

PR #1278, opened 2023-02-16
chore: Delete rfqm dead code

kyu-c:
# Description

Delete rfqm dead code

# Testing & Simbot Run

* n/a

# Checklist


src/entities/RfqmJobEntity.ts
```diff
@@ -59,7 +59,7 @@ enum RfqmJobStatus {
  * @deprecated
  * Kept here to ensure existing migrations type check
  */
-export enum RfqmOrderTypes {
+enum RfqmOrderTypes {
```
rhinodavid, line 62: I think you can delete `migrations` and this too

rhinodavid (approved):
hell yes burn it down

src/entities/RfqmJobEntity.ts
kyu-c: looks like there is a bunch of related files. I will delete them in another PR.

PR #1289, opened 2023-02-21
fix: UniswapV3 filtering out bad path encodings [LIT-865]

SavDont:
# Description

# Testing & Simbot Run

# Checklist


kliao (approved):
nice find. 

any way we can unit test this and/or make sure the next univ3 clone we implement doesn't run into the same issue?

SavDont:
We could probably unit test this but I think the underlying issue is that router has some precision errors causing inexact input/output amounts that do not match sampler amounts. I would prefer to fix the router precision issue and add a unit test case for router.

